PR #1, opened 2017-11-22
agent: Initial implementation

sboeuf:
This PR creates the first implementation of the kata-containers agent relying on both Yamux and gRPC protocol.


sboeuf:
@sameo @gnawux @laijs @bergwolf I have updated this PR but I am still not happy about the way the different mounts are handled through what we call `Storage` or `Mount`.
I'd really like to understand the detail of what you plan to do (some very concrete examples) with a list of `Storage` from `StartSandbox`. Same thing for the list of mounts passed to the container, and the meaning of `vm:`, `tmp:` and `/dev/hostfs`. BTW I think having a boolean `Rootfs` would be easier to detect if a mount is a rootfs.
I am asking for some concrete examples of use cases because it would help me to handle properly every cases, because I am sure you have some use cases we didn't think about.

Please take a look at the PR during those next days and tell me if something is wrong with this.

laijs:
> Same thing for the list of mounts passed to the container, and the meaning of vm:, tmp: and /dev/hostfs.
These are removed without any components replaced. But they will be replaced by Storage if my suggestion is accepted.

laijs:
The runtime passes this storage config to the agent when starting sandbox:
```json
{
  "driver":"",
  "source":"sharedfs_tag",
  "fstype":"9p",
  "mount_point": "/storage/sharedfs"
}
```
Then agent mount it on "/storage/sharedfs".
(The runtime also mount the containers rootfs and other things to the shared fs parallel)
Then runtime pass the container config to agent with Spec.Root.Path = "/storage/sharedfs/container_name/rootfs".

laijs:
When start a container, Storages could be
```
[
{
  "source":"/dev/sda",
  "fstype":"xfs",
  "mount_point": "/storage/abc/"
},
{
  "source":"/dev/sdb",
  "fstype":"ext4",
  "mount_point": "/storage/xyz/"
}
]
```
And "/storage/abc/rootfs" and "/storage/xyz/_data" (notice the docker style "rootfs" and "_data") can be used in the oci.Specs for container root and volume.

Without new components, it is hard to do such things with oci.Specs merely.

agent.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	return nil
+}
+
+func (p *pod) startGRPC() (err error) {
+	var l net.Listener
+
+	if p.isVSock() {
+		l, err = vsock.Listen(1024)
```
bergwolf: nit: constify the port number and make it part of agent API.

channel.go
```diff
@@ -0,0 +1,139 @@
...
+	"strings"
+
+	"github.com/hashicorp/yamux"
+	"golang.org/x/sys/unix"
+)
+
+type channelType string
+
+const (
+	serialChannelName = "agent.channel.0"
```
bergwolf: move this to somewhere like api.go?

config.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+const (
+	optionPrefix      = "agent."
+	logLevelFlag      = optionPrefix + "log"
+	kernelCmdlineFile = "/proc/cmdline"
+)
+
+type agentConfig struct {
+	logLevel logrus.Level
```
bergwolf: I assume we want to use some kind of wrapper for logrus/glog in the agent as well?

agent.go
```diff
@@ -0,0 +1,461 @@
...
+
+	if _, err := file.Read(buf); err != nil {
+		return nil, err
+	}
+
+	return buf, nil
+}
+
+func (p *pod) initLogger() error {
+	agentLog.Logger.Formatter = &logrus.JSONFormatter{TimestampFormat: time.RFC3339Nano}
```
bergwolf: Can we just use text format for agent? logs are most likely to be read by eyes and JSON is not very readable for human eyes IMO.

agent.go
```diff
@@ -0,0 +1,461 @@
...
+		stream, err := session.Accept()
+		if err != nil {
+			agentLog.Error(err)
+			break
+		}
+		defer stream.Close()
+
+		go func() {
+			io.Copy(grpcConn, stream)
+		}()
```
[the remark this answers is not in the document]
laijs: I think some code will be added around io.Copy() in the function. So it's Ok for me now...

agent.go
```diff
@@ -0,0 +1,461 @@
...
+	agentName       = "kata-agent"
+	exitSuccess     = 0
+	exitFailure     = 1
+	fileMode0750    = 0750
+	defaultLogLevel = logrus.InfoLevel
+)
+
+const (
+	gRPCSockPath   = "/var/run/kata-containers/grpc.sock"
+	gRPCSockScheme = "unix"
```
laijs: I don't think these are needed. unix sock scheme proxy is only needed in the host side since proxy and shim/runtime are different processes. The origin design doesn't include it.

```
 |----runv------|----proxy-------|-------hyperstart
 container 1 ---|--------\             /---
 container 2 ---|--------->------|----<----  hyperstart actions
 exec 1---------|--------/             \---
  grpc      unixsock  yamux   serial   yamux     grpc
```
(@gnawux had drawn a picture from this ascii)

yamux.Server()'s return value implements net.Listener. So it can be directly used with grpc(grpcServer.Serve()).

However, I haven't tried. Cloud you have a try please?

grpc.go
```diff
@@ -0,0 +1,698 @@
...
+	if err := mountShareDir(req.Storage.Source, req.Storage.MountPoint, req.Storage.Fstype); err != nil {
+		return emptyResp, err
+	}
+
+	a.pod.shareDir = req.Storage.MountPoint
+
+	if err := setupDNS(a.pod.network.dns); err != nil {
+		return emptyResp, err
+	}
+
```
laijs: A pod pid namespace might be needed to be created. hyperstart creates one.
the pod's pid namespace's init process acts as infra container.

Or let it become an option in the StartSandboxRequest config?
@sameo @gnawux 

One more question, when infra container was requested from the k8s, will runtime ask the agent create it?

grpc.go
```diff
@@ -0,0 +1,698 @@
...
+		return emptyResp, fmt.Errorf("Terminal is not set, impossible to resize it")
+	}
+
+	fd := int(proc.termMaster.Fd())
+
+	// Get current terminal size.
+	winSize, err := unix.IoctlGetWinsize(fd, unix.TIOCGWINSZ)
+	if err != nil {
+		return emptyResp, err
+	}
```
laijs: I don't think we need to get the winsz at first, for we are going to modify all the values.
```
       TIOCGWINSZ     struct winsize *argp
              Get window size.

       TIOCSWINSZ     const struct winsize *argp
              Set window size.

       The struct used by these ioctls is defined as

           struct winsize {
               unsigned short ws_row;
               unsigned short ws_col;
               unsigned short ws_xpixel;   /* unused */
               unsigned short ws_ypixel;   /* unused */
           };
```

channel.go
```diff
@@ -0,0 +1,139 @@
...
+	virtIOPath        = "/sys/class/virtio-ports"
+	devRootPath       = "/dev"
+	vSockDevPath      = "/dev/vsock"
+)
+
+type channel interface {
+	getType() channelType
+}
```
sameo: So I think we should have a different channel interface:

```Go
type channel interface {
        open() error
        close() error
        listen() (*net.Listener, error)
}
```

With that interface:

* I would implement a `newChannel() (*channel, error)` function that would implement the entire logic of checking if we are going to do vsock or serial, and respectively return the vsock or serial implementation.
* The `func (p *pod) initChannel() error` would become something like:
```Go
func (p *pod) initChannel() error {
    c, err := newChannel()
    if err != nil {
         return err
    }

    p.channel = c
    return p.channel.open()

}
``` 
* The serial `open()` implementation would create the yamux server on its serial path and store the yamux session.
* The serial `listen()` implementation would simply return the yamux session
* `func (p *pod) startGRPC() (err error)` would be simplified to:
```Go
func (p *pod) startGRPC() error {
        l, err := p.channel.listen()
       if err != nil {
          return err
       }

	grpcImpl := &agentGRPC{
		pod: p,
	}

	grpcServer := grpc.NewServer()
	pb.RegisterHyperstartServiceServer(grpcServer, grpcImpl)

	p.wg.Add(1)
	go func() {
		defer p.wg.Done()

		grpcServer.Serve(l)
	}()

	return nil
}
```

And so on...
I think the code would be simplified and more readable overall.

mount.go
```diff
@@ -0,0 +1,261 @@
...
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+
+	pb "github.com/hyperhq/runv/hyperstart/api/grpc"
+	goudev "github.com/jochenvg/go-udev"
```
sameo: That one is a cgo wrapper around libudev. We can live with that for now, but we have to keep in mind that this makes our binary no longer static and brings glibc, lpthread and libudev dependencies.
And what we need from it is basically a netlink socket and monitor from it, so we probably should consider adding UEVENT support to github.com/vishvananda/netlink/. The genetlink support is already there, so it **should** not be a huge task (famous last words..)

agent.go
sboeuf: Yes I agree.

agent.go
sboeuf: Ok

agent.go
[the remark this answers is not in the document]
sboeuf: Yep

agent.go
```diff
@@ -0,0 +1,461 @@
...
+		return nil
+	}
+
+	serialChan, ok := p.channel.(*serialChannel)
+	if !ok {
+		return fmt.Errorf("Could not cast into serialChannel type")
+	}
+
+	// Connect gRPC socket.
+	conn, err := net.Dial(gRPCSockScheme, gRPCSockPath)
```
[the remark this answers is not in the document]
sboeuf: Here we dial into the gRPC socket. We know this socket is being listened since `startGRPC()` first listen onto this socket before to give the listener to `server.Serve()` inside the go routine. There is no reason the Dial would fail in that case.

channel.go
sboeuf: Yes I can do that.

channel.go
```diff
@@ -0,0 +1,139 @@
...
+	grpcConn     net.Conn
+	yamuxSession *yamux.Session
+}
+
+func (c *serialChannel) getType() channelType {
+	return serialChannelType
+}
+
+func isAFVSockSupported() (bool, error) {
+	fd, err := unix.Socket(unix.AF_VSOCK, unix.SOCK_STREAM, 0)
```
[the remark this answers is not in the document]
sboeuf: Why do you want to modprobe vsock ? Don't we want VSOCK to be in-kernel (guest kernel) if we support it ?

agent.go
sboeuf: Oh yes good point, I haven't realized that. It's going to simplify the code.

grpc.go
sboeuf: It is already there: https://github.com/kata-containers/agent/pull/1/files#diff-727ccd4217b7f335e689be03bfb69e9bR322

grpc.go
sboeuf: Well yes, if x and y position is unused I agree there is no need to get the window size before to set it.
@sameo any thoughts about this ?

agent.go
```diff
@@ -0,0 +1,461 @@
...
+	// Start gRPC server.
+	if err = p.startGRPC(); err != nil {
+		return
+	}
+
+	// Start yamux server.
+	if err = p.startYamux(); err != nil {
+		return
+	}
+
```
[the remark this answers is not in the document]
sboeuf: @bergwolf here is the cgroups definition for the current container: https://github.com/kata-containers/agent/pull/1/files#diff-727ccd4217b7f335e689be03bfb69e9bR324
We need to provide any cgroup we want through the OCI Spec we're passing to the agent.

grpc.go
```diff
@@ -0,0 +1,698 @@
...
+func (a *agentGRPC) DestroySandbox(ctx context.Context, req *pb.DestroySandboxRequest) (*gpb.Empty, error) {
+	if a.pod.running == false {
+		agentLog.WithField("pod", a.pod.id).Info("Pod not started, this is a no-op")
+		return emptyResp, nil
+	}
+
+	a.pod.Lock()
+	for key, c := range a.pod.containers {
+		if err := c.removeContainer(); err != nil {
+			return emptyResp, err
```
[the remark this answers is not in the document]
sboeuf: I want to make sure we can properly tear down a pod and its setup. This will avoid any issues related to unmounted directories, network configuration or other.
Also, I think we might want to support (for the future) a `RestartSandbox()`. In that case, we would save the launching time of the VM and it would result into a faster accessible pod.

agent.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		defer stream.Close()
+
+		go func() {
+			io.Copy(grpcConn, stream)
+		}()
+
+		go func() {
+			io.Copy(stream, grpcConn)
+		}()
```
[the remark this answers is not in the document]
sboeuf: Yes agreed

mount.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func waitForBlockDevice(devicePath string) error {
+	deviceName := strings.TrimPrefix(devicePath, devPrefix)
+
+	if _, err := os.Stat(devicePath); err == nil {
+		return nil
+	}
+
+	u := goudev.Udev{}
```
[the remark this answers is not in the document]
sboeuf: Ok I'll try to make this better.

mount.go
sboeuf: I would love that. This would be so much cleaner. I can take a look at this in parallel, so that we can improve pretty quickly.

agent.go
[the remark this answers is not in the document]
bergwolf: yup, you are right.

channel.go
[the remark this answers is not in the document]
bergwolf: hmm, fine with me if vsock is built-in.

agent.go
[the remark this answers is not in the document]
bergwolf: Let me rephrase my question - does libcontainer mount cgroup mountpoints if any of the cgroup configs are present in the OCI spec? If not, the agent should mount them before calling into libcontainer.

grpc.go
bergwolf: @sboeuf That one is one pid namespace per container, which is the reason docker has the pid1 problem. We need to introduce a pod level pid namespace, or at least make it configurable in `StartSandbox()`.

@laijs we have a running agent in each pod, I don't think the k8s pause container is necessary currently. But I'm not sure if k8s community will use it to do something other than zombie process recycling in future.

grpc.go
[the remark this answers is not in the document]
bergwolf: Fair enough. Just want to remind that if there are any issues with unmounted directories, network configuration or other, `DestroySandbox()` returns an error, and the only option runtime has is to shutdown the vm anyway.

BTW, it seem `c.removeContainer()` does not kill/stop the container but just wait endlessly. Then `DestroySandbox()` possibly becomes an API running forever?

grpc.go
laijs: https://github.com/kata-containers/agent/pull/12

grpc.go
```diff
@@ -0,0 +1,698 @@
...
+
+	// re-scan PCI bus
+	// looking for hidden devices
+	if err := ioutil.WriteFile(pciBusRescanFile, []byte("1"), pciBusMode); err != nil {
+		agentLog.WithError(err).Warnf("Could not rescan PCI bus")
+	}
+
+	defaultMountFlags := syscall.MS_NOEXEC | syscall.MS_NOSUID | syscall.MS_NODEV
+
+	config := configs.Config{
```
laijs: all the configs here should be replaced by the config from the request, or merged from the request.

laijs:
I prefer the name Sandbox rather than Pod.
We support various kinds of configs. For example, we allow container to have its own pidns, the containers will not compose as a k8s pod.
Doing s/Pod/Sandbox/g in the code will solve the problem.

agent.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	return buf, nil
+}
+
+func (p *pod) initLogger() error {
+	agentLog.Logger.Formatter = &logrus.JSONFormatter{TimestampFormat: time.RFC3339Nano}
+
+	config := newConfig(defaultLogLevel)
+	if err := config.getConfig(kernelCmdlineFile); err != nil {
```
gnawux: looks quite strange to get log config from `kernelCmdlineFile` though looks it works correctly.

sboeuf:
PR updated

grpc.go
sameo: Agreed with @laijs here. You probably want to convert the `req.OCI` structure into an actual `oci.OCI` and use as much as possible from it.

grpc.go
sameo: Agreed, you don't need to get winsize.

grpc.go
sameo: @bergwolf We got rid of the pause container in CC recently, trying to simplify things.

@sboeuf, please create a PID namespace when `sandbox_pidns` is set.

sameo:
Besides the fact that it needs to be rebased on top of the latest gRPC changes, this looks pretty good. The code is much cleaner and easier to read now.

grpc.go
sboeuf: Yeah I forgot to remove that, it's going to be fixed in next PR version.

grpc.go
sboeuf: Yes I will do that !

grpc.go
sboeuf: @sameo @lajis, if I understand correctly, you mean that we should expect everything to be provide by `req.OCI` ? Are you sure we don't need some default values in the agent ?

grpc.go
sboeuf: @laijs @sameo I have been checking libcontainer but I am not sure it allows for this case. Just to be clear and to make sure we're all on the same page, what do we expect through this ?
My understanding is that we want our agent to create a new PID namespace and provide this namespace to every container started from this agent. How can we achieve this ?

sboeuf:
I have rebased on latest gRPC. I have also fixed the way I handle the storage mounts.

grpc.go
laijs: IMO, agent can have its default values for the container, but the default value can be overwritten by the requested values. And runtime is also encouraged to pass more config even the agent has default value. For example runtime is encouraged to pass the procfs setting of the container into the agent.

sameo:
@sboeuf Please add the network API implementation now that the gRPC parts have been merged and then this will LGTM.

@bergwolf @laijs Could you please have another look?

grpc.go
bergwolf: @sboeuf it looks like if `configs.NEWPID` is not set in the spec, libcontainer creates the new container in caller's pid namespace. Then we can create a pid namespace at `CreateSandbox()`, and enter the pid namespace before calling into libcontainer. I haven't tested if this can work though.

grpc.go
sboeuf: @laijs Makes sense, I'll do that.

grpc.go
sboeuf: @gnawux @laijs @bergwolf @sameo @mcastelino libcontainer allows us to specify a specific namespace, which is good, but we need this namespace to exist already. Creating a persistent namespace could be the solution here, but there is still a problem with that. The first process joining the namespace is going to be considered as `init` inside this namespace, meaning that if it exits, all others processes into this namespaces are going to be killed. So basically starting a container into this shared pid ns would mean that when this container terminates, all other containers are going to be terminated too, which is not the behavior we want.
The solution is to have a persistent process as the "init" process in this ns. Either we go with the agent, but this means that any other process started by the agent could be seen by containers, or we start a `pause` container (everybody loves `pause` container ;)) when we create the sandbox. This way it will join this namespace first, and we are sure it would not die.
In case of `!sharedPidNs`, then we don't do anything. This means in docker case we would have the same output from a `ps -A` while in k8s case, we would see all containers processes + this pause container process as pid 1.
WDYT ?

mount.go
```diff
@@ -0,0 +1,222 @@
...
+		type9pFs,
+		syscall.MS_MGC_VAL|syscall.MS_NODEV,
+		strings.Join(mount.Options, ","))
+}
+
+func unmountShareDir(mount string) error {
+	return syscall.Unmount(mount, 0)
+}
+
+func waitForBlockDevice(devicePath string) error {
```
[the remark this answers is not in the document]
sboeuf: Yes I'll do that !

mount.go
```diff
@@ -0,0 +1,222 @@
...
+
+	fieldLogger := agentLog.WithField("device", deviceName)
+
+	// Check if the device already exists.
+	if _, err := os.Stat(devicePath); err == nil {
+		fieldLogger.Info("Device already hotplugged, quit listening")
+		return nil
+	}
+
+	fieldLogger.Info("Started listening for uevents for block device hotplug")
```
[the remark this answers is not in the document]
sboeuf: Yes I'll do that !

grpc.go
bergwolf: Or we let the agent spawn an `init` process that handles SIGCHLD for all containers like hyperstart currently does. But I can live with a `pause` container :)

bergwolf: Either way, I don't think it's a show stopper for the PR. We can add things incrementally.

bergwolf:
Except for the minor nits above, the PR looks in good shape and I'm giving my LGTM.

grpc.go
sboeuf: @bergwolf I said `pause` container but I was actually thinking of a simple pause binary we would start inside this specific ns. And we would kill it from `DestroySandbox()`.

sboeuf: @bergwolf it does not sound too complicated to implement, and if we agree on it, I can do this directly on this PR. But if we don't agree yet, then we can postpone this to another PR.

sboeuf:
@bergwolf thx for the LGTM, I will work today (and maybe tomorrow) on fixing last part we have been talking about, and I expect something to be ready by end of week!

grpc.go
sameo: @bergwolf @laijs I think ideally we'd implement a separate subreaper as part of this repo, that would does only one thing: reaping all processes inside a pidns. And the agent would launch it for every pidns it has to create. Is that what the hyperstart process does ?

sameo:
LGTM

sameo:
We can handle the shared pidns case in a separate PR.

sameo:
I think the only thing missing from this PR is the s/pod/sandbox/ as requested by @laijs

grpc.go
laijs: @sameo Not really, the agent (not systemd nor sandbox_pidns nor any additional subreaper like containerd-shim) should be the subreaper of all the containers&processes and possible helper processes. If the agent runs as the vm's init process, like hyperstart, "agent as subreaper" was already done, in the other side when agent is not the vm's init process, some code is needed to be added.

laijs: @sboeuf the implementation for sandbox_pidns can be postponed. Any other feature/behavior can be also be delayed to next PRs when it is complicated or under discussion. A issue might need to  be created to track the unimplemented feature/behavior.

sboeuf:
I have updated the PR today. The things missing are:
- Handling of shared PID ns
- Use of `libcontainer/specconv` to convert the OCI spec into a libcontainer config
I will take care of those items in follow up PRs, but I think we should go ahead and merge this one as it is because it looks enough to me as an initial commit.

laijs:
Since there are two lgtm's, I'm going to merge it and let other contributions in.

grpc.go
sboeuf: I have opened #20 and #21 to track that ;)

PR #2, opened 2017-11-23
[RFC] Initial gRPC protocol for agent communication

gnawux:
based on hyperstart gRPC protocol and the discussion in hyperhq/runv#628 and
kata-containers/runtimes#2


sameo:
My only concern is the `StringUser` part. I believe it's not needed, so I'll open an issue for it.
For now, this LGTM.

sameo:
@laijs PTAL and if this looks good to you we'll merge.

gnawux:
OK. Good start point.

I add `Hyper Inc` copyright for the files from runV originally, and add ` Intel Corporation` copyright for those comes from  hyperhq/runv#628.

However, I am thinking if we should put some header for kata org itself.

sameo:
@gnawux Thanks for putting the right copyright, I appreciate.
I don't think we want to give code copyrights to the kata organization, but I am not a lawyer :)

gnawux:
@sameo afaik we may have an additional header, such as

```
// Copyright 2017 HyperHQ Inc.
// Copyright 2017-2018 Kata...
```

But I don't know how it works in detail, I am not a lawyer as well.

laijs:
LGTM

Issue #3, opened 2017-11-24
Missing APIs


laijs:
hyperd depends on such hyperstart APIs:

- portmapping
  - Configured in CreateSandbox to setup portmapping
- ReadFile/WriteFiles
     Read/Write from a container, it is unused now.
   - ~update /etc/hosts~
- hyperstart-exec
    execute a process in the vm rather than in container. currently used for:
  - update portmapping
  - ipvs operations
  - `hyperctl exec --vm` peek/do something inside the vm. debug only.

In one words, more network APIs needs to be added and cover the above requirements.
(except the case for `hyperctl exec --vm`, but do we need to add some debug facilities for debug?)

gnawux:
I am thinking about an API or a pair of APIs for socket passing, i.e. implement mapping a unix socket to guest, such as docker.sock . Could this be done with ReadFile/WriteFile?

laijs:
I don't think ReadFile/WriteFile still needed. It is only used for updating /etc/hosts.
In docker case, /etc/hosts is always keeping updated via 9pfs+bindmount.
In other cases, we should introduce proper API for it.

laijs:
I'm sorry, /etc/hosts was also switched to 9pfs+bindmount in hyperd/hyper.sh.

9pfs-less feature is also in demand. We should provide some APIs to enable updating /etc/hosts, /etc/resolv.conf in such cases.

egernst:
Are these addressed with https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md?

If you agree, let's close this.

jodh-intel:
Re-ping.

PR #4, opened 2017-11-24
grpc: Extend the network API

sameo:
We add the following APIs:

- AddInterface, RemoveInterface
- RemoveRoute, UpdateRoutes


sameo:
@laijs @gnawux @bergwolf PTAL

sameo:
@sboeuf This is a rework of your runv PR.

gnawux:
In the `IpAddress`, Add an family field for IPv6 ? Got quite a few requests for v6 support recently.

sameo:
@gnawux 
> In the IpAddress, Add an family field for IPv6 ? Got quite a few requests for v6 support recently.

Done.

protocols/grpc/hyperstart.proto
```diff
@@ -108,20 +111,31 @@ message CreateSandboxRequest {
...
 message IpAddress {
-	string address = 1;
-	string mask = 2;
+	IpFamily family = 1;
+	string address = 2;
+	string mask = 3;
 }
 
-message UpdateInterfaceRequest {
-	uint64 type	=1;
```
gnawux: I checked the previous runV code for this field. And found it defined the type operations instead of interface. The operations include:

```
+type InfUpdateType uint64
 +
 +const (
 +	AddInf InfUpdateType = 1 << iota
 +	DelInf
 +	AddIP
 +	DelIP
 +	SetName
 +	SetMtu
 +)
```

This API is contributed by Wei from huawei in this commit: hyperhq/runv@84e1f86f40d5dd2d3ced128d92008bf06f1ca913  of PR hyperhq/runv#597 .

In hyperhq/runv#597 , Wei contributed a `interface` command, with which runV could work with external CNI script directly.

bergwolf: So we still need `AddIP`, `DelIP`, `SetName` and `SetMtu`? If so, we'd better define them clearly.

gnawux: What I mean here is -- the new API proposed in this API is in total different flavor:

- the new API is declarative, which include all the interface description in the request;
- the original one is an incremental API, which may only ships the changed value. For example, if we do `SetMtu`, we may know nothing about the IP addresses.

sameo: @gnawux Good point, thanks. Does the `UpdateInterfacesRequest` look better ?

sameo:
cc @mcastelino Do you think we need a `UpdateRoutes` API as well?

sboeuf:
@gnawux @laijs @bergwolf What do you think about this PR ? Do you think we could go ahead with this ?

sboeuf:
LGTM, just want to hear from @mcastelino about `UpdateRoutes()`.

gnawux:
Added @WeiZhang555 from Huawei to kata org, and Wei please help review if this patch has any potential issues on your cases.

protocols/grpc/hyperstart.proto
```diff
@@ -110,20 +113,46 @@ message CreateSandboxRequest {
 message DestroySandboxRequest {
 }
 
+enum IpFamily {
```
WeiZhang555: Is that possible to rename this to "IPFamily" without violation to ".proto" format?
I think "IP" is better than "Ip". The same for "IpAddress".
What do you think?

protocols/grpc/hyperstart.proto
```diff
@@ -110,20 +113,46 @@ message CreateSandboxRequest {
...
+enum UpdateType {
+	AddIP = 0;
+	DelIP = 1;
+	SetName = 2;
+	SetMtu = 3;
+}
+
+message UpdateInterface {
+	UpdateType type = 1;
+	Interface interface = 2;
```
WeiZhang555: Do we need the whole Interface  struct as identifier? 
Maybe a name could be enough？

protocols/grpc/hyperstart.proto
```diff
@@ -110,20 +113,46 @@ message CreateSandboxRequest {
...
 }
 
-message UpdateInterfaceRequest {
-	uint64 type	=1;
-	string device = 2;
-	string newName = 3;
-	repeated IpAddress ipAddresses = 4;
-	uint64 mtu = 5;
+message InterfacesRequest {
+	repeated Interface interfaces = 1;
```
WeiZhang555: I think it’s not good to handle a bunch of interfaces or routes in single request, it will makes things complicated when part of this request failed, because we need to decides if we should roll back the successful operations or continue with next one. Neither option is good enough

laijs:
Will some other network APIs such as portmapping as https://github.com/kata-containers/agent/issues/3 said will be added in this PR.

protocols/grpc/hyperstart.proto
sameo: @WeiZhang555 You need the whole interface to pass the information you want to see updated.
So e.g. if type is `SetMTU`, we will get the MTU from the `Interface` and discard all other fields.

protocols/grpc/hyperstart.proto
sameo: @WeiZhang555 Makes sense to me. I'll update the PR.

protocols/grpc/hyperstart.proto
sameo: @WeiZhang555 Sure, I'll fix that.

protocols/grpc/hyperstart.proto
WeiZhang555: Oh, yes, you’re right. I made a big mistake :smile:

mcastelino:
@laijs when do you anticipate the need to perform port mapping inside the virtual machine? All port mappings today are performed at the host level.

mcastelino:
@sameo we should add UpdateRoutes. IMO we should always use the update method whenever possible vs add/delete methods. Using update the agent can ensure that the setup is always reflective of expected state. Add/Delete makes things more stateful than we desire

protocols/grpc/hyperstart.proto
```diff
@@ -110,20 +116,55 @@ message CreateSandboxRequest {
...
+	Interface interface = 2;
+}
+
+message RouteRequest {
+	Route route = 1;
 }
 
-message AddRouteRequest {
+message UpdateRoutesRequest {
 	repeated Route routes = 1;
```
WeiZhang555: Can we also update single route rule in one request? As for UpdateInterface

WeiZhang555:
@sameo @laijs @gnawux 

Do we have any other channels except github? Such as IRC/ slack/ mailing list? (Maybe this is not the best place to ask, but you're all here :wink: )

protocols/grpc/hyperstart.proto
sameo: @WeiZhang555 The idea with routes updating is that we always give it the full set of routes that we want to see applied (see @mcastelino [comment](https://github.com/kata-containers/agent/pull/4#issuecomment-347587537)), or else we'd have to follow the same logic as with the interface updating mecahnism (i.e. add an update type). I believe that routes are less independent than interfaces and should be updated in bulk.

gnawux: normally routing table are modified rule by rule, i.e. it should be ok if we provide two interface for adding and removing rule(s)

protocols/grpc/hyperstart.proto
```diff
@@ -110,20 +116,55 @@ message CreateSandboxRequest {
...
+}
+
+message Route {
+	string dest = 1;
+	string gateway = 2;
+	string device = 3;
+}
+
+enum UpdateType {
+	IP = 0;
```
WeiZhang555: one `IP` field can't distinguish "add ip" or "remove" ip, for interfaces could have many ip addresses.

protocols/grpc/hyperstart.proto
WeiZhang555: Yes, I checked codes from hyperstart: https://github.com/hyperhq/hyperstart/blob/master/src/net.c#L294-L382 , route rules are handled one by one, so we don't need to pass all the route rules everytime, passing whole set of route rules is only behavior of runv in my thought.

and as @gnawux  said, "add" route and "remove" route should be enough for route operations, not sure if we need a "update" method since route is relatively simply comparing to network interface operations.

WeiZhang555: Yes, I checked codes from hyperstart: https://github.com/hyperhq/hyperstart/blob/master/src/net.c#L294-L382 , route rules are handled one by one, so we don't need to pass all the route rules everytime, passing whole set of route rules is only behavior of runv in my thought.

and as @gnawux said, "add" route and "remove" route should be enough for route operations, not sure if we need a "update" method since route is relatively simply comparing to network interface operations.

sameo: I'll let @mcastelino chime in. I'm fine with or without `UpdateRoutes`, and we can add it later on if we see a need for it.

protocols/grpc/hyperstart.proto
sameo: @WeiZhang555 Fixed.

protocols/grpc/hyperstart.proto
mcastelino: @WeiZhang555 Add and Delete assume that we are always stateful in both the runtime and the agent. The reason to have an update mechanism is to allow for a purely stateless setup of routes in this case, but for that matter anything else.  This will allow for stateless implementation and easier recovery.  An update in this case is an update of the full state, not necessarily of a single route (which does not make sense in any case). However we can add it later if  you feel that it is not necessary.

sameo:
I removed `UpdateRoutes` from this PR, in order to move forward with the API extensions. I think @mcastelino is quite right about being as stateless as possible and I hope we can add this API later on.

gnawux:
@WeiZhang555 does this look good to you? We'd like to merge it if everything goes well.

WeiZhang555:
LGTM :+1:

laijs:
@sameo @gnawux @sboeuf @mcastelino @WeiZhang555 
I'm trying to update the network api in #56 again.

PR #5, opened 2017-11-24
grpc: Add commands to stop and remove a container

sameo:
This commit introduces extra commands related to a container, and
specifically to the tear down of a container.

StopContainer() will be used in order to force a container to be
stopped without having to bother finding back the container process.
The StopContainerRequest timeout is used for waiting before forcing
a container termination.

RemoveContainer() will remove a container from the VM, by tearing down
everything related to this container, so that a new container with the
same ID could be started again later.


sameo:
@laijs @gnawux @bergwolf PTAL
@sboeuf This is a replacement for your pending runv PR.

protocols/grpc/hyperstart.proto
```diff
@@ -42,10 +44,17 @@ message CreateContainerRequest {
 	Spec OCI = 4;
 }
 
-message StartContainerRequest {
+message ContainerRequest {
 	string container_id = 1;
 }
 
+message RemoveContainerRequest {
```
bergwolf: Did you intend to type StopContainerRequest, according to the commit log?

bergwolf:
We also need to be clear that a running container cannot be removed by RemoveContainer() API.

gnawux:
Could we split the change request to two -- one for the stop, and the  other for the remove. We have almost no comment on the `RemoveContainer()`.

For the stop, there are two ways:

- signal + wait -- async request and wait exit event outside the vm, and this way is covered by the current APIs;
- independent stop -- sync request, and ship a graceful timeout seconds (might a signal?), which need add the api and a correspond code in agent.

I prefer to keep the agent simple. It is always possible to add the API whenever we found it is essential.

sameo:
@gnawux PR split into https://github.com/kata-containers/agent/pull/8 and https://github.com/kata-containers/agent/pull/7

protocols/grpc/hyperstart.proto
sameo: @bergwolf Yes, sorry.

PR #6, opened 2017-11-24
grpc: Use gogoproto for being closer to the the OCI spec

sameo:
With gogoproto, we get a few interesting protobuf extensions for
golang, like the ability to make structure fields non optional and unit
tests generation for the generated code.

This will make the conversion routines between OCI and gRPC faster.


sboeuf:
LGTM

hack/update-generated-agent-proto.sh
```diff
@@ -3,4 +3,5 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-protoc --proto_path=protocols/grpc --go_out=plugins=grpc:protocols/grpc protocols/grpc/hyperstart.proto protocols/grpc/oci.proto
+go get -u github.com/gogo/protobuf/gogoproto/
```
bergwolf: Please do not try to update the binary, which can be time consuming for us.

bergwolf:
Except for the extra step of updating gogoproto, the PR looks good to me.

hack/update-generated-agent-proto.sh
sameo: @bergwolf Done.

bergwolf:
@sameo this now conflicts with master

sameo:
@bergwolf Rebased now.

PR #7, opened 2017-11-26
grpc: Add RemoveContainer command

sameo:
RemoveContainer() will remove a container from the VM, by tearing down
everything related to this container, so that a new container with the
same ID could be started again later.


laijs:
I like this API. But I want to know..
What's the reason why it is needed? Since we have KillProcess().
What's the semantic meaning of the API?
Will it be always forcedly removed regardless the state of the container?
Is it better if this API attaches a return code or a reason ?

sameo:
@laijs `RemoveContainer` would have the same semantics as the hyperstart `RemoveContainer` json based command. Does that clarify things a bit?

sboeuf:
@laijs `RemoveContainer()` is totally different from `KillProcess()` in a context of Pod. Even if you `kill/stop` a container, you might still want to be able to remove all the associated resources from the Pod.
I'd expect `RemoveContainer()` to return an error in case the container is still running.

bergwolf:
@sboeuf In hyperstart, when a container is stopped, all of its resources are teared down. What resources are you referring to by `be able to remove all the associated resources`?

sameo:
> In hyperstart, when a container is stopped, all of its resources are teared down.

@bergwolf With the currrent gRPC API, we can't stop a container. So we can rename that command to `StopContainer` but the idea is to be able to stop and (implicitly) remove all associated resources to this container.

bergwolf:
@sameo What API sequence are you suggesting to gracefully stop/remove a container? I thought that with `SignalProcess()`, we are able to kill a container init process (and all the other exec processes get killed along with it) and thus all the container's resources get teared down.

But it looks like it's not the case and you want `SignalProcess()` + `RemoveContainer()` together to do the job?

sameo:
@bergwolf `SignalContainer()` would only send a signal (any signal) to the container processes. It could implicitly mean that *some* container related resources would also be torn down if e.g. the init process is SIGKILL'ed or SIGTERM'ed. `RemoveContainer()` will force kill all containers processes if there are any left and tear down all resources.
So yes, a clean sequence would be `SignalContainer(SIGTERM)` and then `RemoveContainer()`.

bergwolf:
@sameo please explain a bit more about what to expect when calling `RemoveContainer()`.
 - does it send any signals to any process?
 - does it wait process termination?

sameo:
@bergwolf `RemoveContainer` will kill (SIGKILL) the init process (and thus all remaining container processes), it does not wait for process termination.
I think we can decide if we want RemoveContainer() to return an error if all processes have not been stopped (SIGTERM'ed) or not.

WeiZhang555:
I agree with @laijs , I'm also confused about the meaning of this API.

If we kill the "init" process of a container, the container should be gone and all resources it takes are returned, so in which scenario should we call "RemoveContainer" explicitly?

Do you want to keep container resources after the container is killed, until we call "RemoveContainer" explicitly?

sameo:
@WeiZhang555 If you're trying to gracefully kill init, you're going to `SignalContainer(SIGTERM)` but you have no guarantee that:
1. The init process is effectively stopped
2. All processes from the containers are stopped

The `Signal*` API only guarantees that the signal has been delivered to one or more processes inside a container. If you do a `SignalContainer(SIGKILL)` you know that the processes will be gone, but you don't know if all internal data and resources associated with a container are gone. The following sequence:

1. `SignalContainer(SIGTERM)`
2. `SignalContainer(SIGKILL)`

is not clean imho, because a) You should only `SIGKILL` as a last resort for lingering processes and b) it takes an implicit assumption that sending a signal will clean all internal structures and resources associated with a container.

A graceful and clean container teardown would look like:

1. `SignalContainer(SIGTERM)` for trying a graceful termination
2. `RemoveContainer()` for `SIGKILL`'ing only the remaining processes and explictly asking for a complete container tear down.

sboeuf:
@bergwolf @laijs @gnawux @sameo I think `RemoveContainer()` makes things easier and cleaner when we are tearing down a container from the runtime. It helps removing the internal structures that we have been creating with `CreateContainer()`. I agree, after the container process has returned (naturally or because it has been signaled), we could cleanup, but this is all implicit and it would be easier to have a dedicated function for that. This changes nothing in the container flow. Basically we would have the container process terminating, and then the shim would receive the exit code (because it would have called into `WaitProcess()` before), and then Docker will detect the end of the shim and send `cc-runtime delete` or `runv delete`. Our `delete` implementation will call into this `RemoveContainer()` so that it really tells the agent to remove everything related to the container (even if everything means only internal structures). This way we're having a clear way of the cleanup being done.
To summarize, I don't think it would change anything to your flow either, since we're both implementing OCI specification, and it would bring more clarity.

laijs:
LGTM

sboeuf:
@gnawux @bergwolf Please merge it looks good to you too.

bergwolf:
@sameo @sboeuf I agree we need `RemoveContainer()`. I still think we need to clearly define what to expect with it. Namingly, as @sameo said above, agent will `SIGKILL` all remaining processes in the container but does not wait for process termination, then as `RemoveContainer()` returns, the process might still be running and the runtime does not know if the internal structures are cleaned up or not.

IMO `RemoveContainer()` should first `SIGKILL` all remaining processes in a container, and then make sure all processes have terminated and clean all related resources before returning (or return an error if any process cannot be killed). If so, we need an error code for the `RemoveContainer()` API.

WeiZhang555:
>Our delete implementation will call into this RemoveContainer() so that it really tells the agent to remove everything related to the container (even if everything means only internal structures).

Makes sense.  Adding `RemoveContainer` looks good.

Only one remaining question: If container still has alive process, should we force all processes die and then remove container,  or report an error and let upper level try to kill all processes(with SIGKILL) explicitly again?

gnawux:
>IMO `RemoveContainer()` should first `SIGKILL` all remaining processes in a container, and then make sure all processes have terminated and clean all related resources before returning (or return an error if any process cannot be killed). If so, we need an error code for the `RemoveContainer()` API.

Agree, and If we got a success return from `RemoveContainer()`, we could unplug any devices from the VM and related cleanup.

sameo:
@bergwolf @laijs @WeiZhang555 PR refreshed with documentation.

WeiZhang555:
LGTM

PR #8, opened 2017-11-26
grpc: Add SignalContainer command

sameo:
SignalContainer() sends a signal to the init process or all
processes inside a given container.


laijs:
What's semantic meaning of StopContainer()? Which state of the container will be left in sider the container ?
Namespace (mount)? processes? configure and name ?

sameo:
@laijs @bergwolf I modified this PR to actually send a signal to a container.
The semantic are simple: Sending this command will send a signal to either the container init process (`all = false`) or all processes inside the container (`all = true`).

protocols/grpc/hyperstart.proto
```diff
@@ -42,10 +43,18 @@ message CreateContainerRequest {
 	Spec OCI = 4;
 }
 
-message StartContainerRequest {
+message ContainerRequest {
```
bergwolf: no need to rename StartContainerRequest any more.

sameo: Done.

protocols/grpc/hyperstart.proto
```diff
@@ -46,6 +47,14 @@ message StartContainerRequest {
 	string container_id = 1;
 }
 
+message SignalContainerRequest {
+	string container_id = 1;
+	uint32 signal = 2;
+	// When all is set to true, the signal will be
+	// sent to all processes inside the container.
+	bool all = 3;
```
laijs: We don't need this flag(bool all) at all. If the only the init process is signaled, we can just use the SignalProcess().

SignalProcess() only kill the root processes of the init/exec process. So I admit that SignalContainer() might be needed if SignalContainer() kills not only the root processes but all the natural spawned linux process,  only when libcontainer support this operation.

laijs:
Where are requirement from for this feature?

Here are the help message for `runc kill`
`kill        kill sends the specified signal (default: SIGTERM) to the container's init process`

sameo:
@laijs 

```
runc kill --help
NAME:
   runc kill - kill sends the specified signal (default: SIGTERM) to the container's init process

USAGE:
   runc kill [command options] <container-id> [signal]

Where "<container-id>" is the name for the instance of the container and
"[signal]" is the signal to be sent to the init process.

EXAMPLE:
For example, if the container id is "ubuntu01" the following will send a "KILL"
signal to the init process of the "ubuntu01" container:
   
       # runc kill ubuntu01 KILL

OPTIONS:
   --all, -a  send the specified signal to all processes inside the container

```

See the `--all` option.
So this is where it's coming from and also libcontainer already supports it.

laijs:
API processes: the root/entry linux process of the init/exec processes
linux processes: just linux processes. including the root/entry linux process of the init/exec processes and any other linux processes spawned from the root/entry linux processes.

The signal request can be split to:
1) signal to non-init API processes
2) signal to init API process
3) signal to all the linux processes in the container.
We(hyper) want the 1)&2) use the same API and 3) uses another API(a).
You(intel) want the 2)&3) use the same API(b) and 1) uses another API.

API(a): SignalContainer() as this pr but be used for 3) only without the "bool all" argument.
API(b): This pr.

egernst:
Refreshing this stale PR.  Should we open an issue for what this PR is fixing, assuming we still want it?  Perhaps this'll help clarify the use case since there was contention on the implementation/usage?

laijs:
it is done by 'SignalProcess("container", "")'

PR #9, opened 2017-11-27
grpc: add missing process id

bergwolf:
So that we can find the process later on...


sameo:
@bergwolf Would it make sense to have both `CreateContainer` and `ExecProcess` return a process ID ?

laijs:
I prefer the runtime deciding the process id.

sameo:
@bergwolf @laijs Are you expecting to use that process ID for signaling the init process inside a specific container, through `SignalProcess` ?

laijs:
Sure, I don't see anything wrong.

protocols/grpc/hyperstart.proto
```diff
@@ -37,9 +37,10 @@ service HyperstartService {
 
 message CreateContainerRequest {
 	string container_id = 1;
-	StringUser string_user = 2;
-	repeated Storage storages =3;
-	Spec OCI = 4;
+	string process_id = 2;
```
sameo: Would this `process_id` be needed for signaling the container init process later on?

bergwolf:
@sameo process id in createcontainer removed. PTAL.

sameo:
LGTM

sameo:
@bergwolf Could you please rebase so that we can merge that one?

bergwolf:
@sameo rebased.

PR #11, opened 2017-11-27
grpc: Replace WaitProcess with GetEvents

sameo:
WaitProcess is meant to be called for monitoring a process and receiving
its exit code. For flexibility and generalization purposes, we build a
more generic, gRPC stream based API for receiving events. The first 2
events that we may stream are container (init process) and process
terminations.


sameo:
@bergwolf @laijs @gnawux @sboeuf Combining this PR with pr #8 would allow us to not pass a process ID as part of the container creation request (see #9 ).

protocols/grpc/hyperstart.proto
```diff
@@ -18,7 +19,6 @@ service HyperstartService {
 	rpc StartContainer(StartContainerRequest) returns (google.protobuf.Empty);
 	rpc ExecProcess(ExecProcessRequest) returns (google.protobuf.Empty);
 	rpc SignalProcess(SignalProcessRequest) returns (google.protobuf.Empty);
-	rpc WaitProcess(WaitProcessRequest) returns (WaitProcessResponse); // wait & reap like waitpid(2)
```
jodh-intel: `waitProcessRequest` is now unused so the message needs to be removed.

protocols/grpc/agent.proto
```diff
@@ -159,3 +162,26 @@ message Route {
 	string gateway = 2;
 	string device = 3;
 }
+
+message EventsRequest {
+	string container_id = 1;
+	uint32 PID = 2;
+	google.protobuf.Timestamp timestamp = 3;
```
WeiZhang555: Does the timestamp field means "since" ?

protocols/grpc/agent.proto
```diff
@@ -159,3 +162,26 @@ message Route {
...
+}
+
+message ProcessTermination {
+	string process_id = 1;
+	int32 exit_code = 2;
+}
+
+message Event {
+	oneof event {
+		ContainerTermination container_termination = 1;
```
WeiZhang555: I guess `ContainerTermination`  will be sent when all processes of container are terminated?

Is this equal to "receive ProcessTermination event of all processes"? It looks like it can be composed of multiple "ProcessTermination" event

sboeuf:
It's not relevant for now, and not even in the scope for 1.0. I think we can close it and have this discussion later if someone feels we should reopen the topic.

PR #12, opened 2017-11-28
grpc: specify if shared pidns is needed in CreateSandbox

bergwolf:
Use it to tell agent if a pod level pid namespace is required. We may want to just always use a pod level pidns but this option keeps in sync with k8s per-pod PID Namespace Sharing (https://github.com/kubernetes/features/issues/495) and crio `--disable-shared-pid-namespace` flag (https://github.com/kubernetes-incubator/cri-o/pull/1149).


bergwolf:
@sameo @sboeuf @laijs PTAL.

protocols/grpc/hyperstart.proto
```diff
@@ -103,6 +103,7 @@ message CreateSandboxRequest {
 	string hostname = 1;
 	repeated string dns = 2;
 	repeated Storage storages = 3;
+	bool shared_pidns = 4;
```
laijs: It seams that all the containers are forced to be **shared** the sandbox pidns.

I prefer create_sandbox_pidns or sandbox_pidns or infra_pidns.

bergwolf: Done.

laijs:
LGTM, @sameo @sboeuf @jodh-intel WDYT.
related discussion: https://github.com/kata-containers/agent/pull/1#discussion_r152782367

sameo:
LGTM

laijs:
hello, I'm trying to remove it and make it mandatory. #54

PR #13, opened 2017-11-29
grpc: Return a PID when starting a container or a process

sameo:
Instead of generating a process ID string from the runtime,
we can simply use the actual PID (container init or exec'ed
process) from the VM. A process is then uniquely identified
by a (container ID, PID) couple, which we can use for any
process specific operation.

With this patch we do not need to maintain 2 separate process
identification schemes, and can use a single one (actual Linux
PID) across all components.


sameo:
@laijs @bergwolf @sboeuf @mcastelino This is a PR intended to open a discussion as to why do we need to have the runtime generate process ID strings while we already have a unique identifier for each process generated by the kernel inside the container.
It seems more natural to me to use those PIDs across all components rather than maintain this list from the runtime and have the agent build an additional strings to PIDs map, so I wonder if I'm missing something that would explain the need for this added complexity.

protocols/grpc/hyperstart.proto
```diff
@@ -70,7 +73,7 @@ message WaitProcessResponse {
 
 message WriteStreamRequest {
 	string container_id = 1;
-	string process_id = 2;
+	string pid = 2;
```
WeiZhang555: uint32 ??

WeiZhang555:
I like this proposal, +1 on the design. But still we can wait for @laijs to check is there's something we're missing

protocols/grpc/hyperstart.proto
```diff
@@ -46,22 +46,25 @@ message StartContainerRequest {
...
 	string container_id = 1;
-	string process_id = 2;
 	StringUser string_user = 3;
 	Process process = 4;
 }
 
 message SignalProcessRequest {
 	string container_id = 1;
-	string process_id = 2;
+	uint32 pid = 2;
```
laijs: Is it allowed when runtime makes these requests by passing a pid which is not returned from StartContainer() nor ExecProcess(), but the passing pid is actually existing the the container?

laijs:
We are updating the API from the scratch, so I think it would be better to complement more documents in this period.

protocols/grpc/hyperstart.proto
WeiZhang555: I prefer keeping "container_id"  here. We can implement a mechanism to signal to "ALL" processes in a container with a SignalProcessRequest{container_id=xxx, process_id=0} 

Something more we can do is "sending a signal 0 means detect whether the process in a container is alive or not". As "man 2 kill" semantics

WeiZhang555: > Is it allowed when runtime makes these requests by passing a pid which is not returned from StartContainer() nor ExecProcess(), but the passing pid is actually existing the the container?

Any scenario?

laijs: @WeiZhang555 runtime might wrongly pass the incorrect number of pid, or pass an outdated pid when the pid is reused in the guest. The behavior should be documented.

sameo: @laijs I don't think we should allow for requests with a pid that has not been created by the agent and returned through gRPC. I'll document that.

sameo:
Documented the PID field.

jodh-intel:
lgtm

WeiZhang555:
LGTM :smile:

laijs:
LGTM

laijs:
@sameo CreateContainer() should return NewProcessResponse
The pid for the init-process should be known when `runtime create` return.
See `runc create --pid-file`

In our case the pid is the shim-pid, and shim's arguments needs the process's pid inside the vm.
The way to spawn the shim can be changed to fix it. But the better solution is fixing the API.

PR #15, opened 2017-11-29
protocol: grpc: Add Flags parameter to Storage structure

sboeuf:
protocol: grpc: Add Flags parameter to Storage structure
    
    Basically, the storage structure is being used to describe a mount
    that should be performed prior to the container launching. And if
    we take a look at the parameters needed to perform a mount, flags
    is one of them.
    
    The idea is that our agent will expose a map of string in order to
    describe the supported flags. This way, it will translate those
    strings into actual flags understood by the guest OS.
    

laijs:
I once thought we can use "ro" in the options.
Splitting "ro" from the options is good to me.
But there are some other flags might be in the options, such as MS_NOATIME, MS_NOSUID, MS_NODEV. I think we can also introduce a "flags" for these flags.

protocols/grpc/agent.proto
```diff
@@ -144,8 +144,9 @@ message Storage {
 	string driver = 1;           // empty in most cases. it will support "drbd", "bcache" ...
 	string source = 2;           // "/dev/sdb", "/dev/disk/by-scsi/xxxx", "none",...
 	string fstype = 3;           // "xfs", "ext4" etc. for block dev, or "9p" for shared fs, or  "tmpfs" for shared /dev/shm for all containers ...
-	repeated string options = 4; // options for the storage device, might required by some special devices.
-	string mount_point = 5;      // mount_point is only visible by VM, not by containers. This mount point can be used on oci.Mount.Source as "/Stroage/mount/point/{rootfs|_data}".
+	bool read_only = 4;          // specify if this should be mounted with the read only flag.
+	repeated string options = 5; // options for the storage device, might required by some special devices.
```
gnawux: a minor thing, how about move the `options` to the last

sboeuf: @gnawux yes sure I can do that.

sboeuf:
@laijs I agree it might be better to introduce a `Flags` parameter. But we need to make sure we are clear about what is considered as a flag and what is considered as an option. And which format we expect so that we can parse them properly.

sboeuf:
@laijs @sameo @bergwolf @gnawux please take a look, I have changed this PR by adding a generic list of flag instead of a single ReadOnly parameter.

laijs:
If flags is split out from the option, it should be `u32`, otherwise we can leave them in the `options` and let the agent parse the options and filter MS_flags into integer flags.

laijs:
Some requirement such as docker-init/volume-init as https://github.com/kata-containers/runtimes/issues/4 said or hyper.sh style file volume can be implemented via driver.
But a `driver-options` is needed for these purpose, could you also add the `driver-options`?

sboeuf:
@laijs by having flags as a `u32`, I am afraid the agent could have a different interpretation of what `syscall.MS_...` could be, don't you think ?
And yes I see your point, we can use the list of options to do that, but the problem is that you won't make the difference between a flag not supported (not known), and an option.

sboeuf:
Yes I can add a `driver-options`, no problem!

sboeuf:
Closing this PR since we have discussed that options will hold both the flags and the options: https://github.com/kata-containers/agent/issues/24

PR #16, opened 2017-11-30
protocol: grpc: Allow single storage entry for CreateSandbox()

sboeuf:
We don't need a storage list in case of CreateSandbox(). This function
is a very particular case where we are passing a shared directory via
9pfs. This shared directory is the way to share some specific files
and directories between host and guest. This is different from
CreateContainer() function where we expect several devices and
volumes to be passed to the VM.


sboeuf:
cc @laijs @bergwolf @gnawux @sameo

laijs:
We have serval possible sandbox level storages
1) 9pfs for all containers.
2) shared /dev/shm (tmpfs) for all containers.
The agent can make it to be default when sandbox_pidns is requested. But is better that the runtime explicitly makes the config.
3) multiple container images for a pod are allowed to be put into a block device.

so it is better that the storage config for the sandbox is also an array.

sboeuf:
@laijs fair enough, I'll change slightly the agent code to match with this behavior.

PR #25, opened 2017-12-01
protocol: grpc: Allow several updates of the same interface

sboeuf:
protocol: grpc: Allow several updates of the same interface
    
    When using UpdateInterface(), the way the UpdateType is defined only
    allows for one operation at a time. This commit modifies the enum
    values of the UpdateType so that they can be used as flags. This way,
    we can combine several updates (IPs, Name, MTU) of the same interface
    in one call.
    

sboeuf:
cc @sameo @laijs according to what we have discussed.

gnawux:
Could we add and remove IP Addresses in a single call?

sboeuf:
Well yes basically, that's why we need the agent implementation to forbid such behavior. In case the flag combination is `AddIP | RemoveIP`, we'll throw an error.

gnawux:
LGTM

sameo:
LGTM

laijs:
hello #56 is a proper way to allow several updates of the same interface IMO.
And it also support removing&adding at the same time.

Issue #19, opened 2017-11-30
Change route gRPC API from imperative to declarative


mcastelino:
Today we add/delete routes. However this will not work for plugins such as PTP which manipulate routes in non standard manner. So routes API should be performed one shot using update/set Routes.

For more details

https://github.com/containers/virtcontainers/issues/500

egernst:
Initial implementation works on a route at a time.  Will address this next by updating routes in a single shot.  To avoid conflict, the first step of this update will be to clear out all pre-existing routes.  It is assumed that this UpdateRoutes call will be made after the interfaces (and their addresses) are setup.

egernst:
Closed by https://github.com/kata-containers/agent/pull/148

PR #27, opened 2017-12-03
 protocols: add `ListInterface` and `ListRoute`

WeiZhang555:
Add two more grpc interfaces:

*`ListInterface`: list all the interfaces inside a VM.
*`ListRoute`: list all route rules inside VM.

Generally, the runtime is reaponsible to manage networks and routes via
sending command to agent inside VM, without these two functions, there's
no way to know accurate interface/route lists, which makes it hard to
sync up network information between runtime and agent.

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>


WeiZhang555:
Issues and links: https://github.com/kata-containers/agent/issues/19 and https://github.com/kata-containers/agent/pull/4
ping @mcastelino @sameo , this is still under discussion, we can find the best way doing network settings

WeiZhang555:
The `hack/update-generated-agent-proto.sh` doesn't work on my machine, so I added `--proto_path=protocols/grpc:/usr/local/include`,  "/usr/local/include" is where my protobuf included files sit. 
I'm not sure if this is caused by wrong configuration on my machine, I'll remove this change if I'm doing it wrong,

mcastelino:
@WeiZhang555 this is one way to solve the problem. The other method would be to make it the responsibility of the agent within the VM to ensure that desired state is achieved. This will allow for a simpler agent. 

If we go declarative would make for a simpler runtime but a slightly more complex agent.

WeiZhang555:
@mcastelino Yes, I agree. We can wait for other's opinion, and find a best way for this.

WeiZhang555:
@mcastelino Another reason for adding these two interfaces are that I'm thinking about the "runtime interface ls" cmdline, which can list all the interfaces inside VM. So without these two grpc interfaces,  we have to keep two sets of interface list in both `runtime` and `agent`, when `runtime` send an "AddInterface" command to `agent`, `agent` will save the list and `runtime` will also save the list.

One problem caused by this model is that if there's some failure or changes of interfaces inside VM, there's no way to let `runtime` know, and `runtime interface ls` will always show outdated data.

laijs:
This will complicate the agent API. The interface should be stable after successfully updated/added.
All we needed is to maintain some meta data outside the vm and protect the runtime from the possible hacked or misbehavior vm/agent.

sboeuf:
@WeiZhang555 I don't think a `runtime interface ls` would show outdated data since every time it will call into `Add/Update/Remove`, we will get an error if something got wrong, and we know this means the interface has not been added/updated/removed.

WeiZhang555:
@sboeuf One example (though I'm not sure if we should take care): what if user modify the interface configuration by hand? should we reflect this changes through `runtime interface ls` API?

laijs:
@WeiZhang555  It is not the case in container world.

WeiZhang555:
@laijs I agree that in most cases such as public cloud or a K8S cluster, we always use a CNI plugin to manage networks and user in container won't have enough privilege to modify the network configuration. But actually in container world, there're still some "quite weird" use cases, for example, we have a special container type named "system container", which will run a container with "systemd" running and user will feel like they are running inside a VM instead of a container(redhat has same technology as I know). In such scenario, user is allowed to modify the networks by hand. But maybe what I talked about is too weird, I can close this if you and @sboeuf still consider this change irrational

sboeuf:
@WeiZhang555 I don't think this change is irrational, but let's postpone this and reopen this PR only in case we will end up having a real need for that. Do you think that's fair ?

WeiZhang555:
@sboeuf Yep, fair enough. I'm closing this now and let's focus on other part currently.

PR #28, opened 2017-12-04
agent: Implement subreaper support

### LINKED ISSUE #23: make agent itself as subreaper
- [ ] A loop that calls syscall.Wait() to reap processes
     And to collect processes' exitcode
- [ ] Make itself as PR_SET_CHILD_SUBREAPER
     It is essential when the agent is not launched as the vm's init process.

sboeuf:
agent: Implement subreaper support
    
This commit turns the agent into a subreaper, in order to reap all
    the processes started from any container. This is the only way to
    make sure we won't end up with some zombies processes in case we
    have some child processes terminating before their own children.
    
Fixes #23
    
### DIFF
--- a/agent.go
+++ b/agent.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"os/signal"
 	"runtime"
 	"sync"
 	"syscall"
@@ -21,6 +22,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
 )
 
@@ -57,6 +59,7 @@ type sandbox struct {
 	grpcListener net.Listener
 	sharedPidNs  namespace
 	mounts       []string
+	subreaper    *reaper
 }
 
 type namespace struct {
@@ -247,7 +250,7 @@ func (s *sandbox) setupSharedPidNs() error {
 		Cloneflags: syscall.CLONE_NEWPID,
 	}
 
-	if err := cmd.Start(); err != nil {
+	if err := s.subreaper.start(cmd); err != nil {
 		return err
 	}
 
@@ -272,12 +275,42 @@ func (s *sandbox) teardownSharedPidNs() error {
 		return err
 	}
 
-	if _, err := s.sharedPidNs.init.Wait(); err != nil {
+	// Using helper function wait() to deal with the subreaper.
+	osProcess := (*reaperOSProcess)(s.sharedPidNs.init)
+	if _, err := s.subreaper.wait(osProcess.Pid, osProcess); err != nil {
 		return err
 	}
 
 	// Empty the sandbox structure.
 	s.sharedPidNs = namespace{}
+
+	return nil
+}
+
+// This loop is meant to be run inside a separate Go routine.
+func (s *sandbox) reaperLoop(sigCh chan os.Signal) {
+	for sig := range sigCh {
+		switch sig {
+		case unix.SIGCHLD:
+			if err := s.subreaper.reap(); err != nil {
+				agentLog.Error(err)
+				return
+			}
+		default:
+			agentLog.Infof("Unexpected signal %s, nothing to do...", sig.String())
+		}
+	}
+}
+
+func (s *sandbox) setSubreaper() error {
+	if err := unix.Prctl(unix.PR_SET_CHILD_SUBREAPER, uintptr(1), 0, 0, 0); err != nil {
+		return err
+	}
+
+	sigCh := make(chan os.Signal, 512)
+	signal.Notify(sigCh, unix.SIGCHLD)
+
+	go s.reaperLoop(sigCh)
 
 	return nil
 }
@@ -380,9 +413,17 @@ func main() {
 	s := &sandbox{
 		containers: make(map[string]*container),
 		running:    false,
+		subreaper: &reaper{
+			exitCodeChans: make(map[int]chan int),
+		},
 	}
 
 	if err = s.initLogger(); err != nil {
+		return
+	}
+
+	// Set agent as subreaper.
+	if err = s.setSubreaper(); err != nil {
 		return
 	}
 
--- a/reaper.go
+++ b/reaper.go
@@ -1,0 +1,185 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"sync"
+
+	"github.com/opencontainers/runc/libcontainer"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+type reaper struct {
+	sync.RWMutex
+
+	chansLock     sync.RWMutex
+	exitCodeChans map[int]chan int
+}
+
+func exitStatus(status unix.WaitStatus) int {
+	if status.Signaled() {
+		return exitSignalOffset + int(status.Signal())
+	}
+
+	return status.ExitStatus()
+}
+
+func (r *reaper) getExitCodeCh(pid int) (chan int, error) {
+	r.chansLock.RLock()
+	defer r.chansLock.RUnlock()
+
+	exitCodeCh, exist := r.exitCodeChans[pid]
+	if !exist {
+		return nil, fmt.Errorf("Process %d not found", pid)
+	}
+
+	return exitCodeCh, nil
+}
+
+func (r *reaper) setExitCodeCh(pid int, exitCodeCh chan int) {
+	r.chansLock.Lock()
+	defer r.chansLock.Unlock()
+
+	r.exitCodeChans[pid] = exitCodeCh
+}
+
+func (r *reaper) deleteExitCodeCh(pid int) {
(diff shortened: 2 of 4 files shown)


sboeuf:
Fixes #23 
cc @bergwolf @laijs @gnawux @sameo

bergwolf:
I think we should combine shared pidns support with zombie children reaper so that either
  - the agent is pid1 of the shared pidns and handles SIGCHLD, or
  - the `pause` process lives is the pid1 of the shared pidns and handles SIGCHLD

agent.go
```diff
@@ -54,6 +56,9 @@ type sandbox struct {
 	grpcListener net.Listener
 	sharedPidNs  bool
 	mounts       []string
+
+	procChLock  sync.RWMutex
+	processesCh map[int]chan int
```
laijs: the exitcode should be saved on type process.
the type process for the exec-processes will be removed when a successful WaiProcess()
the type process for the init-process will be removed when RemoveContainer()

laijs:
A new mechanism interacting with the loop is needed to be implemented for replacing the origin exec.Cmd.Wait(), os.Process.Wait(), in case when the agent need to exec some helper processes and wait them.

We can move above sentence to a new issue and implement it later.

agent.go
sboeuf: Not sure I follow here. Is there something wrong with the way I have implemented the channel message delivery between the subreaping loop and the WaitProcess() ?

laijs: There is nothing incorrect with the processesCh.
I hope the variables for the process are put into the type process.

sboeuf: Ok I get it, but the issue by trying to find the right process entry is that we're going to loop over every container map of processes and check for each one if there is a corresponding entry. Having this very simple map at the pod level simplifies things and we don't need all the informations contained into a `process` structure.

laijs: How about adding `map[int] process` in the `sandbox`?

laijs: type ProcessID int

sboeuf: Oh you mean somehting like this ?
```go
type ProcessID int

type sandbox struct {
        ...
        processesCh map[int]chan ProcessID
```

I need this map to be a map of channels since I expect WaitProcess() to listen on this channel until it gets the exit code.

laijs: add `map[ProcessID] process` to sandbox
add `chan exitcode` to type process

sboeuf: Ok this would work but this means we would have a map of process structures and we would only use the exit code. That's a bit too much, don't you think ?

I could add map[ProcessID] ExitCode with both ProcessID and ExitCode being int types, this would be simpler.

laijs: I don't want the fields of a process to be scattered.
If it is complicated due to the locking, the locking can be simplified by only using the lock in the sandbox and removing the lock in container.

sboeuf:
@laijs @bergwolf FWIW, I have applied the same kind of changes to our Clear Containers agent here: https://github.com/clearcontainers/agent/pull/178 and I did some testing about the case where a child returns before a grandchild. I confirm I got the SIGCHLD from the agent running on the root PID and both child and grandchild were running into a container in a different PID namespace. So I confirm this works as expected.

For testing, I have started a container with `docker run -it busybox`. Then, I retrieved the container ID (`2217c770e3cc`) and I have started `docker exec -it 2217c770e3cc sh`. From the exec process shell, I have started `# nohup sleep 30 &` in order to spawn a process and get the terminal back. Right after, I have exited the terminal, leading to the termination of the exec process (the child in our case), and after about 30 seconds, I have seen some agent logs saying that it received a SIGCHLD signal and it had to wait for one process, the one corresponding to the `sleep 30`.

I just wanted to let you know that we're good here since it got properly tested :)

agent.go
sboeuf: Well this field is not scattered since it's not even needed into the `process` structure.

sameo: @laijs It's more about duplication than locking complexity. Each `container` already has a `processes   map[int]*process` map and we'd be duplicating this information in the sandbox now. And we'd have to keep those maps coherent, which adds complexity as well.
If we want to keep all processes information in a single structure (Which I agree would make sense), I'd prefer to keep one PID <-> process map per container and no more. We could maintain a PID <-> container id map at the sandbox level to avoid going through all containers to find a process when being given a PID

sboeuf: @laijs any comment ?

laijs: The death messages of the other helper processes created directly by agent are routed to this loop. We need a mechanism to help programing and to replace exec.Cmd.Wait() or os.Process.Wait(). We can reuse the mechanism in containerd or implement a new one.

sboeuf: @laijs I am sorry I really don't follow on this one. Can you try to explain what you need me to implement and why ? Also, I don't get how the last comment is related to the entire conversation.

laijs: @sboeuf when agent is subreaper, exec.Cmd.Wait() and os.Process.Wait() for helper processes are not workable, we should provide a replacement.

laijs: @sboeuf it is quite simple in containerd.
https://github.com/containerd/containerd/blob/master/reaper/reaper.go

sboeuf: @laijs What do you mean by "helper processes" ?
Also, do you expect that we provide a different implementation of `Wait()` so that we can override the standard behavior ? How can we do that exactly ?
I have looked at the containerd code and I think it partially does what we do, or am I missing something that you expect me to add to our code ?

sboeuf: @laijs actually I get it now :)
You expect that we define a new way to spawn processes in case we want to use something simple like `exec.Cmd.Wait()`. I am sorry it took me so long but I thought you were trying to solve an issue in the current implementation. Instead you're anticipating that we won't be able to rely on the stardard `exec` and `os` functions since we would end up in a racy case where we would not know who will get to reap the child process first (between this function call and our subreaper).
I'll implement this and update the PR.

sboeuf:
@laijs I have updated the PR by including a new mechanism allowing proper Wait() for `*exec.Cmd` and `*os.Process` types. @bergwolf I have updated the shared PID ns code by using those new helpers.
Let me know how it looks now!

agent.go
```diff
@@ -57,6 +59,9 @@ type sandbox struct {
 	grpcListener net.Listener
 	sharedPidNs  namespace
 	mounts       []string
+
+	procChLock  sync.RWMutex
+	processesCh map[int]chan int
```
laijs: The code is not clean for me.

I hope the new exec_process.go is renamed to subreaper.go or reaper.go
And `type Subreaper struct {xx}` in the file.
Move the above two fields to the struct. sync.RWMutex can be anonymous for some convenience.
Move the related functions to be the functions of `type Subreaper`.
Other new functions in this PR may need to be moved to the file too.

exec_process.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return nil
+}
+
+// waitProcess blocks until the expected process has been reaped. After the
+// reaping from the subreaper, the exit code is sent through the provided
+// channel.
+// This function is a helper for exec.Cmd.Wait() and os.Process.Wait() since
+// both cannot be used directly, because of the subreaper.
+func (s *sandbox) waitProcess(obj interface{}) (int, error) {
```
laijs: please define an interface with the Wait() as its function.

agent.go
sboeuf: done

exec_process.go
sboeuf: done

sboeuf:
@laijs I have updated the PR according to your comments !

sameo:
LGTM

reaper.go
```diff
@@ -0,0 +1,184 @@
...
+		ws  unix.WaitStatus
+		rus unix.Rusage
+	)
+
+	// When running new processes, libcontainer expects to wait
+	// for the first process actually spawning the container.
+	// This lock allows any code starting a new process to take
+	// the lock prior to the start of this new process, preventing
+	// the subreaper from reaping unwanted processes.
+	r.Wait()
```
laijs: A new request might just arrive after r.Wait() returns.
In this case the new call into the libcontainer and the next unix.Wait4() happen parallel.
One of the methods to solve this problem is changing the sync.WaitGroup to sync.RWMutex too.

sboeuf: Yes I agree. But to make this safe, I will have to lock the entire `for` loop, right ? This is the only way to make sure the new request will wait until the reaper exits the loop.

sboeuf: PR updated!

reaper.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/opencontainers/runc/libcontainer"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+type reaper struct {
+	sync.Mutex
+
+	chansLock     sync.RWMutex
+	exitCodeChans map[int]chan int
```
laijs: In general, the invoking count of write lock is the same as the invoking count of reader lock.
they are both about the number of the registered pid multiply by 2.

the major lock of the reaper (reaper.Mutex) makes the creations of processes sequenced. it can be changed to RWMutex to allow them parallel.

Anyway, both are very trivial problems.

sboeuf: @laijs I have updated the PR by using a read lock for new processes, so that they can be started in parallel. The reaper is a RW lock since we do want it to block until all read locks are done.
About the 2 different mutexes, this is the way to synchronize the reaper and processes in one hand, while in the other hand, we can ensure a map is not accessed at the same time from two different processes.

laijs:
LGTM

PR #29, opened 2017-12-04
agent: Add shared PID namespace support

### LINKED ISSUE #20: Implement sandbox shared PID namespace support
We need a way to launch all containers in the same PID namespace in case we get the flag `SandboxPidns` when `CreateSandbox()` is called.

sboeuf:
agent: Add shared PID namespace support
    
    We want to be able to share the same PID namespace across all
    containers of a sandbox. This commit allows for such a thing by
    starting a new process from a pause binary, into a new PID ns.
    
    This way, the process is considered as init process inside this
    namespace and the namespace will exist as long as this process
    runs.
    
    Fixes #20
    
### DIFF
--- a/agent.go
+++ b/agent.go
@@ -11,8 +11,10 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"runtime"
 	"sync"
+	"syscall"
 	"time"
 
 	pb "github.com/kata-containers/agent/protocols/grpc"
@@ -53,8 +55,13 @@ type sandbox struct {
 	network      network
 	wg           sync.WaitGroup
 	grpcListener net.Listener
-	sharedPidNs  bool
+	sharedPidNs  namespace
 	mounts       []string
+}
+
+type namespace struct {
+	path string
+	init *os.Process
 }
 
 var agentLog = logrus.WithFields(logrus.Fields{
@@ -222,6 +229,57 @@ func (s *sandbox) readStdio(cid string, pid int, length int, stdout bool) ([]byt
 	}
 
 	return buf, nil
+}
+
+// setupSharedPidNs will reexec this binary in order to execute the C routine
+// defined into pause.go file. The pauseBinArg is very important since that is
+// the flag allowing the C function to determine it should run the "pause".
+// This pause binary will ensure that we always have the init process of the
+// new PID namespace running into the namespace, preventing the namespace to
+// be destroyed if other processes are terminated.
+func (s *sandbox) setupSharedPidNs() error {
+	cmd := &exec.Cmd{
+		Path: selfBinPath,
+		Args: []string{os.Args[0], pauseBinArg},
+	}
+
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Cloneflags: syscall.CLONE_NEWPID,
+	}
+
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	// Save info about this namespace inside sandbox structure.
+	s.sharedPidNs = namespace{
+		path: fmt.Sprintf("/proc/%d/ns/pid", cmd.Process.Pid),
+		init: cmd.Process,
+	}
+
+	return nil
+}
+
+func (s *sandbox) teardownSharedPidNs() error {
+	if s.sharedPidNs.path == "" {
+		// Nothing needs to be done because we are not in a case
+		// where a PID namespace is shared across containers.
+		return nil
+	}
+
+	// Terminates the "init" process of the PID namespace.
+	if err := s.sharedPidNs.init.Kill(); err != nil {
+		return err
+	}
+
+	if _, err := s.sharedPidNs.init.Wait(); err != nil {
+		return err
+	}
+
+	// Empty the sandbox structure.
+	s.sharedPidNs = namespace{}
+
+	return nil
 }
 
 func (s *sandbox) initLogger() error {
--- a/grpc.go
+++ b/grpc.go
@@ -336,6 +336,11 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 			},
 			{
 				Type: configs.NEWPID,
+				// In case the path is empty because we
+				// don't expect the containers to share
+				// the same PID namespace, a new PID ns
+				// is going to be created.
+				Path: a.sandbox.sharedPidNs.path,
 			},
 		}),
 		Cgroups: &configs.Cgroup{
@@ -673,7 +678,12 @@ func (a *agentGRPC) CreateSandbox(ctx context.Context, req *pb.CreateSandboxRequ
 	a.sandbox.id = req.Hostname
 	a.sandbox.network.dns = req.Dns
 	a.sandbox.running = true
-	a.sandbox.sharedPidNs = req.SandboxPidns
+
+	if req.SandboxPidns {
+		if err := a.sandbox.setupSharedPidNs(); err != nil {
+			return emptyResp, err
+		}
+	}
 
 	mountList, err := addMounts(req.Storages)
 	if err != nil {
@@ -713,12 +723,15 @@ func (a *agentGRPC) DestroySandbox(ctx context.Context, req *pb.DestroySandboxRe
 		return emptyResp, err
 	}
 
+	if err := a.sandbox.teardownSharedPidNs(); err != nil {
+		return emptyResp, err
+	}
+
 	a.sandbox.id = ""
 	a.sandbox.containers = make(map[string]*container)
 	a.sandbox.running = false
 	a.sandbox.network = network{}
 	a.sandbox.mounts = []string{}
-	a.sandbox.sharedPidNs = false
 
 	// Synchronize the caches on the system. This is needed to ensure
 	// there is no pending transactions left before the VM is shut down.
--- a/pause.go
+++ b/pause.go
@@ -1,0 +1,46 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+/*
+#cgo CFLAGS: -Wall
+#define _GNU_SOURCE
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#define PAUSE_BIN "pause-bin"
+
+static void sigdown(int signo) {
+	psignal(signo, "shutting down, got signal");
+	exit(0);
+}
+
+void __attribute__((constructor)) sandbox_pause(int argc, const char **argv) {
+	if (argc != 2 || strcmp(argv[1], PAUSE_BIN)) {
+		return;
+	}
+
+	if (signal(SIGINT, sigdown) == SIG_ERR)
+		exit(1);
+
+	if (signal(SIGTERM, sigdown) == SIG_ERR)
+		exit(2);
+
+	for (;;) pause();
+
+	fprintf(stderr, "error: infinite loop terminated\n");
+	exit(42);
+}
+*/
+import "C"
+
+const (
+	pauseBinArg = string(C.PAUSE_BIN)
+)
--- a/api.go
+++ b/api.go
@@ -30,4 +30,5 @@ const (
 	exitFailure     = 1
 	fileMode0750    = 0750
 	defaultLogLevel = logrus.InfoLevel
+	selfBinPath     = "/proc/self/exe"
 )


sboeuf:
Fixes #20 
cc @sameo @bergwolf @gnawux @laijs

laijs:
It is better not to introduce a secondary binary, It would be a hell for maintainance.

It would be better to use "/proc/self/exe" as the binary, and use "kata-agent", "sandbox-pidns" as the arguments to launch the process.

The C code can be put in the `void __attribute__((constructor)) sandbox_pidns(int argc, const char **argv)`

laijs:
And this problem should be fixed too:
https://github.com/hyperhq/hyperstart/issues/326

We can use a new mountns for sandbox_pidns and make the
sandbox_pidns process's rootfs as an empty dir.

sboeuf:
@laijs you introduce C code in Go, it's not very neat... I can write the same short program in Go, but in both cases, a `reexec` is going to cost more memory than a simple C binary exec, don't you think ?

laijs:
libcontainer has already had C code: ./libcontainer/nsenter/nsexec.c which we also use it in a indirectly way.

sboeuf:
@laijs yeah fair enough for the inclusion of C code. What about the overhead of reexec a Go binary instead of exec a C binary directly. Will this be the same eventually, or should we expect a memory footprint difference ?

laijs:
I think `__attribute__((constructor))` is better than rexec in memory consumption. No go code involved before `__attribute__((constructor))`. Correct me if I was wrong...

WeiZhang555:
I don't have a preference between "reexec golang kata-agent" or "a second pause C binary", but the memory footprint is important to me. Maybe we should test the memory footprint difference first, and see how large is that difference. We should try to save more memory for user if possible.

sboeuf:
@laijs there is already some Go code in the agent taking care of the reexec case through the unit() function. Don't you think this is going to conflict if we define this another init function in the C way ?

laijs:
The argv in `void __attribute__((constructor)) sandbox_pidns(int argc, const char **argv)` can be checked carefully to avoid conflict. I don't mind you implement it via go + reexec.
If anyone concern about it later or any problem happened, then we switch to constructor scheme.

sboeuf:
@laijs of course libcontainer is doing a `reexec`. Everything is prepared with `libcontainer,New()`, then `factory.Create()`, and eventually `container.Run()`. If you look into details, the `InitPath` of the factory becomes the `initPath` of the container, and then this path is used to spawn the container. So yes, this is a reexec which will go through the init() function and be `execve`.

Now, about the reexec in Go, this is really bad about memory consumption... The agent binary is pretty big and that's why I had written a tiny one in C. I think the options are:
- Either we keep an external C binary
- Or we can try your proposal about `__attribute__((contructor))` if we can avoid conflict with the Go `init()` function. But I think this is gonna have a large memory footprint since this is the entire executable in Go which is "reexecuted", no matter how tiny our C code will be.
@sameo @mcastelino I'd like some feedback on this, I am not sure about what to expect.

laijs:
__attribute__((contructor)) is earlier than any go code, don't need to care about the memory consumption.
The entire executable in Go are shared mapped with the agent, it don't consume memory.

laijs:
https://gist.github.com/laijs/10a9ec476cf5875d2ba75c3071769128

 The whole memory consumption (RSS & Dirty) in the `constructor` stage of the golang program ./contructormaps is almost the same as the pure c lang program ./c-contructormaps

```
# ./c-contructormaps
14744:   ./c-contructormaps
Address           Kbytes     RSS   Dirty Mode  Mapping
0000000000400000       4       4       0 r-x-- c-contructormaps
0000000000600000       4       4       4 r---- c-contructormaps
0000000000601000       4       4       4 rw--- c-contructormaps
0000000000623000     132       4       4 rw---   [ anon ]
00007fd4fa3ba000    1760     292       0 r-x-- libc-2.17.so
00007fd4fa572000    2048       0       0 ----- libc-2.17.so
00007fd4fa772000      16      16      16 r---- libc-2.17.so
00007fd4fa776000       8       8       8 rw--- libc-2.17.so
00007fd4fa778000      20      12      12 rw---   [ anon ]
00007fd4fa77d000     132     112       0 r-x-- ld-2.17.so
00007fd4fa98b000      12      12      12 rw---   [ anon ]
00007fd4fa99d000       4       4       4 rw---   [ anon ]
00007fd4fa99e000       4       4       4 r---- ld-2.17.so
00007fd4fa99f000       4       4       4 rw--- ld-2.17.so
00007fd4fa9a0000       4       4       4 rw---   [ anon ]
00007ffd9b80a000     132      16      16 rw---   [ stack ]
00007ffd9b84f000       8       4       0 r-x--   [ anon ]
ffffffffff600000       4       0       0 r-x--   [ anon ]
---------------- ------- ------- -------
total kB            4300     504      92
```

```
# ./contructormaps
14913:   ./contructormaps
Address           Kbytes     RSS   Dirty Mode  Mapping
0000000000400000    1204      16       0 r-x-- contructormaps
000000000072c000       4       4       4 r---- contructormaps
000000000072d000      76       8       8 rw--- contructormaps
0000000000740000     124       0       0 rw---   [ anon ]
0000000001adc000     132       4       4 rw---   [ anon ]
00007f6da1fad000    1760     292       0 r-x-- libc-2.17.so
00007f6da2165000    2048       0       0 ----- libc-2.17.so
00007f6da2365000      16      16      16 r---- libc-2.17.so
00007f6da2369000       8       8       8 rw--- libc-2.17.so
00007f6da236b000      20      12      12 rw---   [ anon ]
00007f6da2370000      92      48       0 r-x-- libpthread-2.17.so
00007f6da2387000    2044       0       0 ----- libpthread-2.17.so
00007f6da2586000       4       4       4 r---- libpthread-2.17.so
00007f6da2587000       4       4       4 rw--- libpthread-2.17.so
00007f6da2588000      16       4       4 rw---   [ anon ]
00007f6da258c000     132     112       0 r-x-- ld-2.17.so
00007f6da279a000      12      12      12 rw---   [ anon ]
00007f6da27ac000       4       4       4 rw---   [ anon ]
00007f6da27ad000       4       4       4 r---- ld-2.17.so
00007f6da27ae000       4       4       4 rw--- ld-2.17.so
00007f6da27af000       4       4       4 rw---   [ anon ]
00007ffd244f4000     132      16      16 rw---   [ stack ]
00007ffd245ba000       8       4       0 r-x--   [ anon ]
ffffffffff600000       4       0       0 r-x--   [ anon ]
---------------- ------- ------- -------
total kB            7856     580     108
14913:   ./contructormaps
Address           Kbytes     RSS   Dirty Mode  Mapping
0000000000400000    1204     676       0 r-x-- contructormaps
000000000072c000       4       4       4 r---- contructormaps
000000000072d000      76      40      32 rw--- contructormaps
0000000000740000     124      40      40 rw---   [ anon ]
0000000001adc000     132       4       4 rw---   [ anon ]
000000c000000000       4       4       4 rw---   [ anon ]
000000c41fff8000    1056     232     232 rw---   [ anon ]
00007f6d9b7ff000       4       0       0 -----   [ anon ]
00007f6d9b800000    8192    2048    2048 rw---   [ anon ]
00007f6d9c000000     132       4       4 rw---   [ anon ]
00007f6d9c021000   65404       0       0 -----   [ anon ]
00007f6da07aa000       4       0       0 -----   [ anon ]
00007f6da07ab000    8192       8       8 rw---   [ anon ]
00007f6da0fab000       4       0       0 -----   [ anon ]
00007f6da0fac000    8192       8       8 rw---   [ anon ]
00007f6da17ac000       4       0       0 -----   [ anon ]
00007f6da17ad000    8192       8       8 rw---   [ anon ]
00007f6da1fad000    1760     300       0 r-x-- libc-2.17.so
00007f6da2165000    2048       0       0 ----- libc-2.17.so
00007f6da2365000      16      16      16 r---- libc-2.17.so
00007f6da2369000       8       8       8 rw--- libc-2.17.so
00007f6da236b000      20      12      12 rw---   [ anon ]
00007f6da2370000      92      64       0 r-x-- libpthread-2.17.so
00007f6da2387000    2044       0       0 ----- libpthread-2.17.so
00007f6da2586000       4       4       4 r---- libpthread-2.17.so
00007f6da2587000       4       4       4 rw--- libpthread-2.17.so
00007f6da2588000      16       4       4 rw---   [ anon ]
00007f6da258c000     132     112       0 r-x-- ld-2.17.so
00007f6da26fa000     652      52      52 rw---   [ anon ]
00007f6da27ac000       4       4       4 rw---   [ anon ]
00007f6da27ad000       4       4       4 r---- ld-2.17.so
00007f6da27ae000       4       4       4 rw--- ld-2.17.so
00007f6da27af000       4       4       4 rw---   [ anon ]
00007ffd244f4000     132      20      20 rw---   [ stack ]
00007ffd245ba000       8       4       0 r-x--   [ anon ]
ffffffffff600000       4       0       0 r-x--   [ anon ]
---------------- ------- ------- -------
total kB          107876    3692    2528
```

laijs:
All you need is to put the whole origin c pause logic into constructor, and keep the program staying in the constructor.

laijs:
I think a simple clone(CLONE_NEWPID) without any execute() will work too, because the c code will not going back the the go code.

sboeuf:
@laijs I have run the tests and even if my RSS results are higher, the logic is the same. I mean that I get almost the same RSS between the C program and the Cgo from `_attribute_(constructor)`. I have been also trying the reexec from Go to Go, and I can tell you it's awful. Everything gets loaded and duplicated, which leads to almost the same RSS between the original process and the reexec one. Therefore I think we really want to remove this solution from the possibilities.

About the CGo, the gap I have is 1200Kbytes (c-process) vs 1600Kbytes (cgo-process). The diff is completely acceptable since it would be added only once per VM/agent since we won't create several shared PID namespaces.

So I agree the Cgo would be the best way to go to get rid of an external binary, but I am wondering if it is worth it since the code is going to get ugly (I really don't like introducing this C code inside our Go code...).

@sameo I would like your input on this, and I'll apply the changes if you're okay with the solution proposed by @laijs

pause/pause.c
```diff
@@ -0,0 +1,28 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+static void sigdown(int signo) {
+	psignal(signo, "shutting down, got signal");
+	exit(0);
+}
+
+int main() {
```
jodh-intel: A C pause binary is going to be smaller than a golang one, but this might (again) lead to problems supporting CentOS and RHEL...?:

- https://github.com/clearcontainers/runtime/pull/434

WeiZhang555: Will `cgo` pause binary have same problem?

bergwolf: @jodh-intel That is because we want to support multiple distro rootfs. IMO the guest rootfs should be minimal and contains as less packages as possible. There is little value in support different distros. In hyperstart, we simple throw in a few binaries (busybox, iptables, ipvsadm, mount.nfs etc.) for different architectures. And there is no distro support at all.

For kata, I understand the difficulty to add systemd support on our own and we can rely on a certain LTS distro. But I still don't see the value of support different distros in guest rootfs.

sameo: @bergwolf The value is for users/customers that want to control what's in the guest and basically let their OSV support the guest image.

sameo:
@sboeuf If we'd be introducing `cgo` with @laijs proposal I'd be against it. But since unfortunately libcontainer already relies on it (I did not know) and our `kata-agent` binary is already dynamically linked, I actually think @laijs approach makes sense.
As for ugly code, we can entirely put the cgo code into its own go file. But we really need to document this from `agent.go` `main()` routine.

sboeuf:
@sameo @laijs ok then, I'll do that !

pause/pause.c
WeiZhang555: I vote to supporting different guest images, this could be useful feature for us.

sboeuf:
@laijs I have updated the code, please let me know if you like it better this way !

amshinde:
@sameo @sboeuf The cgo code is required in libcontainer to do a setns to a mount namespace. See this: https://github.com/golang/go/issues/8676
As mentioned in the issue, a setns system call for a mount namespace does not work in Go, something I have run into in the past.

pause.go
```diff
@@ -0,0 +1,46 @@
...
+#include <unistd.h>
+
+#define PAUSE_BIN "pause-bin"
+
+static void sigdown(int signo) {
+	psignal(signo, "shutting down, got signal");
+	exit(0);
+}
+
+void __attribute__((constructor)) constructor_map(int argc, const char **argv) {
```
laijs: The example code use "constructor_map()" to show the memory map on the stage of "__attribute__((constructor))".

And the proper name here might be "sandbox_init_process()", or "sandbox_pause()", or something else...

agent.go
```diff
@@ -224,6 +231,47 @@ func (s *sandbox) readStdio(cid string, pid int, length int, stdout bool) ([]byt
 	return buf, nil
 }
 
+func (s *sandbox) setupSharedPidNs() error {
+	cmd := exec.Command(selfBinPath, pauseBinArg)
```
laijs: please keep the arg[0] unchanged via using
```go
&exec.Cmd{
     Path: selfBinPath, 
     Args: []string{os.Args[0], pauseBinArg},
}

**Comment by laijs** [comment]
//off-topic

@amshinde Except the reason that a setns system call for a mount namespace does not work in Go, are there any other reasons why nsexec() in libcontainer has to be implemented in C?

//I didn't keep watching the disccusions/changes of nsexec.c.

**Comment by WeiZhang555** [comment]
@laijs check this blog: https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix

It perfectly illustrates how `setns` in Go codes mess everything up :smile:  and why we should implement `nsexec()` in C:

> This finding raised a question whether we can safely change a namespace in Go. Unfortunately, the answer is no, as we do not have almost any control on scheduling goroutines.
>
>One could argue that locking a goroutine with runtime.LockOSThread could help, but a) the goroutine might spawn a new goroutine which would run in a wrong namespace b) locking does not prevent the runtime from creating a new OS thread for scheduling.

File: agent.go (hunk 3 above)
Comment (sameo) [reply agent.go]: And please document that.

File: pause.go (hunk 2 above)
Comment (sboeuf) [reply pause.go]: Yes, sorry I forgot to change the name...

**Comment by sboeuf** [comment]
@WeiZhang555 yes namespaces are pretty tricky in Go. But I still think that using runtime.LockOSThread actually works if you know exactly what you're doing in your code. But this means we have to be careful how to handle such a thing. About the agent itself, I don't think we'll have to deal with this since everything related to namespaces is going to be handled by libcontainer, but we might have this same discussion later about the runtime.

File: pause.go (hunk 2 above)
Comment (sboeuf) [reply pause.go]: done

**Comment by sameo** [comment]
LGTM

File: agent.go [hunk 4]
```diff
@@ -224,6 +231,57 @@ func (s *sandbox) readStdio(cid string, pid int, length int, stdout bool) ([]byt
...
+		// where a PID namespace is shared across containers.
+		return nil
+	}
+
+	// Terminates the "init" process of the PID namespace.
+	if err := s.sharedPidNs.init.Kill(); err != nil {
+		return err
+	}
+
+	if _, err := s.sharedPidNs.init.Wait(); err != nil {
```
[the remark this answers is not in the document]
laijs: That pr is not merged yet, it is that pr's responsibility to handle this situation. You can add a new issue in case we forgot it.

pause.go
```diff
@@ -0,0 +1,46 @@
...
+
+void __attribute__((constructor)) sandbox_pause(int argc, const char **argv) {
+	if (argc != 2 || strcmp(argv[1], PAUSE_BIN)) {
+		return;
+	}
+
+	if (signal(SIGINT, sigdown) == SIG_ERR)
+		exit(1);
+
+	if (signal(SIGTERM, sigdown) == SIG_ERR)
```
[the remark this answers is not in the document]
laijs: Oh, I didn't notice it. could you sent a pr for it?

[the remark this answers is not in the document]
bergwolf: #60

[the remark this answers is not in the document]
sboeuf: @bergwolf nice catch!

grpc.go
```diff
@@ -336,6 +336,11 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 			},
 			{
 				Type: configs.NEWPID,
+				// In case the path is empty because we
+				// don't expect the containers to share
+				// the same PID namespace, a new PID ns
+				// is going to be created.
+				Path: a.sandbox.sharedPidNs.path,
```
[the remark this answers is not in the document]
sboeuf: That's the point, in that case, we expect the container to be run inside the given PID ns. Otherwise setting NEWPID type with an empty Path will start the container in a new PID ns.

PR #35, opened 2017-12-04
mount: Handle storage options as flags and options combination

### LINKED ISSUE #26: Handle Storage options as a combination of flags and options
The `Options` field inside `Storage` structure is going to be used both for passing some `flags` and `data` for the `mount()` system call. The semantic behind this will be to check if an option from the list matches with a flag from a list of flags defined in the agent API. In case it does not match with any flag, we will consider the option as a _real_ option that should be used as `data`.

sboeuf:
mount: Handle storage options as flags and options combination
    
    The Options field inside Storage structure is used to pass both
    flags and data to the mount() system call. The semantic behind
    this actually checks if an option from the list matches with a
    flag from a specific list of flags. In case it does not match
    any flag, the option is considered as a real option that should
    be used as data parameter.
    
    Fixes #26
    
### DIFF
--- a/mount.go
+++ b/mount.go
@@ -17,6 +17,7 @@ import (
 	"github.com/kata-containers/agent/pkg/uevent"
 	pb "github.com/kata-containers/agent/protocols/grpc"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -25,6 +26,36 @@ const (
 	timeoutHotplug = 3
 	mountPerm      = os.FileMode(0755)
 )
+
+var flagList = map[string]int{
+	"acl":         unix.MS_POSIXACL,
+	"bind":        unix.MS_BIND,
+	"defaults":    0,
+	"dirsync":     unix.MS_DIRSYNC,
+	"iversion":    unix.MS_I_VERSION,
+	"lazytime":    unix.MS_LAZYTIME,
+	"mand":        unix.MS_MANDLOCK,
+	"noatime":     unix.MS_NOATIME,
+	"nodev":       unix.MS_NODEV,
+	"nodiratime":  unix.MS_NODIRATIME,
+	"noexec":      unix.MS_NOEXEC,
+	"nosuid":      unix.MS_NOSUID,
+	"rbind":       unix.MS_BIND | unix.MS_REC,
+	"relatime":    unix.MS_RELATIME,
+	"remount":     unix.MS_REMOUNT,
+	"ro":          unix.MS_RDONLY,
+	"silent":      unix.MS_SILENT,
+	"strictatime": unix.MS_STRICTATIME,
+	"sync":        unix.MS_SYNCHRONOUS,
+	"private":     unix.MS_PRIVATE,
+	"shared":      unix.MS_SHARED,
+	"slave":       unix.MS_SLAVE,
+	"unbindable":  unix.MS_UNBINDABLE,
+	"rprivate":    unix.MS_PRIVATE | unix.MS_REC,
+	"rshared":     unix.MS_SHARED | unix.MS_REC,
+	"rslave":      unix.MS_SLAVE | unix.MS_REC,
+	"runbindable": unix.MS_UNBINDABLE | unix.MS_REC,
+}
 
 // mount mounts a source in to a destination. This will do some bookkeeping:
 // * evaluate all symlinks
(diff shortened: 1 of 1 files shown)


sboeuf:
Fixes #26 
cc @sameo @bergwolf @gnawux @laijs @WeiZhang555

mount.go
```diff
@@ -26,6 +27,50 @@ const (
...
+	"mand":          unix.MS_MANDLOCK,
+	"noacl":         unix.MS_POSIXACL,
+	"noatime":       unix.MS_NOATIME,
+	"nodev":         unix.MS_NODEV,
+	"nodiratime":    unix.MS_NODIRATIME,
+	"noexec":        unix.MS_NOEXEC,
+	"noiversion":    unix.MS_I_VERSION,
+	"nolazytime":    unix.MS_LAZYTIME,
+	"nomand":        unix.MS_MANDLOCK,
+	"norelatime":    unix.MS_RELATIME,
```
laijs: The handling for "noiversion", "nolazytime", "nomand", "norelatime" ... "suid" are incorrect.
The corresponding flags should be cleared rather than be added.

jodh-intel: This looks to be a re-implementation of code already in libcontainer as used by the agent. Unfortunately, the function is private :sad::

- https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L646..L702

sboeuf: @laijs oh yeah you're right and actually we don't need those `no...` flags.

sboeuf:
PR updated.

mount.go
laijs: "noexec", "nodev", "nosuid" are removed in your updated pr.
I think it's better to copy the exact code.

sboeuf: I have updated the PR by making sure I was adding only what matches a flag. I don't think we need to copy the entire code from https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L648-L732 since I don't expect a user of the gRPC protocol using `dev` for example in case he wants `nodev` to be disabled. The caller should only ask for flags to be enabled since it controls the entire list of flags.

laijs:
LGTM now

sboeuf:
@sameo @bergwolf @WeiZhang555 please take a look and merge if you're fine with this.

mount.go
```diff
@@ -26,6 +27,39 @@ const (
 	mountPerm      = os.FileMode(0755)
 )
 
+var flagList = map[string]int{
+	"acl":         unix.MS_POSIXACL,
+	"async":       unix.MS_SYNCHRONOUS,
```
laijs: This line is wrong

mount.go
```diff
@@ -26,6 +27,39 @@ const (
...
+	"bind":        unix.MS_BIND,
+	"defaults":    0,
+	"dirsync":     unix.MS_DIRSYNC,
+	"iversion":    unix.MS_I_VERSION,
+	"lazytime":    unix.MS_LAZYTIME,
+	"loud":        unix.MS_SILENT,
```
laijs: This line is wrong

mount.go
```diff
@@ -26,6 +27,39 @@ const (
...
+	"noatime":     unix.MS_NOATIME,
+	"nodev":       unix.MS_NODEV,
+	"nodiratime":  unix.MS_NODIRATIME,
+	"noexec":      unix.MS_NOEXEC,
+	"nosuid":      unix.MS_NOSUID,
+	"rbind":       unix.MS_BIND | unix.MS_REC,
+	"relatime":    unix.MS_RELATIME,
+	"remount":     unix.MS_REMOUNT,
+	"ro":          unix.MS_RDONLY,
+	"rw":          unix.MS_RDONLY,
```
laijs: this line is wrong

mount.go
sboeuf: Fixed

mount.go
sboeuf: Fixed

mount.go
sboeuf: Fixed

sboeuf:
@laijs PR updated.

laijs:
LGTM

jodh-intel:
lgtm

[Approved with PullApprove](https://pullapprove.com/kata-containers/agent/pull-request/35/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=kata-containers/agent)

PR #36, opened 2017-12-05
fix client unix dialer and mock server

bergwolf:
1. dial with blocking
2. test addr scheme correctly
3. fix mock server startcontainer check


sameo:
LGTM

PR #41, opened 2017-12-05
makefile: Add install target

jcvenegas:
Install target will install systemd units if INIT makefile variable is "no".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 kata-agent
+kata-agent.service
+agent
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,44 @@
 TARGET = kata-agent
 
-$(TARGET):
+DESTDIR :=
+PREFIX := /usr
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1 +1,3 @@
 kata-agent
+kata-agent.service
+agent
```
jodh-intel: This line is redundant I think?

jcvenegas: Because it is  `kata-agent`  to be ignored right? I added because go build also generates that binary (if makefile is not used). So I like to see the binary in git status.

jodh-intel: Ah - fair point. On the topic of `go build`, once we start getting elaborate, we might want to consider something like:

- https://github.com/clearcontainers/runtime/pull/272#issuecomment-316339910

jodh-intel:
lgtm

WeiZhang555:
LGTM

PR #43, opened 2017-12-05
network: Allow several updates on the same interface

### LINKED ISSUE #42: network: allow several updates on the same interface in one API call
Need to implement the support for update actions as flags.

sboeuf:
network: Allow several updates on the same interface
    
    With the recent protocol change for the list of possible actions
    updating a network interface, it is now possible to combine several
    actions in only one call to UpdateInterface(), by adding those actions
    as flags.
    
    This commit implements the support for this feature. It returns an
    error in case none of the known actions are parsed from the action
    type. Indeed, we should not call into UpdateInterface() if there is
    nothing to update.
    
    Fixes #42
    
### DIFF
--- a/network.go
+++ b/network.go
@@ -88,18 +88,38 @@ func updateInterfaceMTU(netHandle *netlink.Handle, link netlink.Link, mtu int) e
 }
 
 func updateLink(netHandle *netlink.Handle, link netlink.Link, iface *pb.Interface, actionType pb.UpdateType) error {
-	switch actionType {
-	case pb.UpdateType_AddIP:
+	knownActionType := false
+
+	// Add IPs
+	if (actionType & pb.UpdateType_AddIP) == pb.UpdateType_AddIP {
+		knownActionType = true
 		return updateInterfaceAddrs(netHandle, link, iface.IpAddresses, true)
-	case pb.UpdateType_RemoveIP:
+	}
+
+	// Remove IPs
+	if (actionType & pb.UpdateType_RemoveIP) == pb.UpdateType_RemoveIP {
+		knownActionType = true
 		return updateInterfaceAddrs(netHandle, link, iface.IpAddresses, false)
-	case pb.UpdateType_Name:
+	}
+
+	// Update Name
+	if (actionType & pb.UpdateType_Name) == pb.UpdateType_Name {
+		knownActionType = true
 		return updateInterfaceName(netHandle, link, iface.Name)
-	case pb.UpdateType_MTU:
+	}
+
+	// Update MTU
+	if (actionType & pb.UpdateType_MTU) == pb.UpdateType_MTU {
+		knownActionType = true
 		return updateInterfaceMTU(netHandle, link, int(iface.Mtu))
-	default:
+	}
+
+	// Error
+	if !knownActionType {
 		return fmt.Errorf("Unknown UpdateType %v", actionType)
 	}
+
+	return nil
 }
 
 func (s *sandbox) addInterface(netHandle *netlink.Handle, iface *pb.Interface) (err error) {


sboeuf:
Fixes #42 
cc @sameo @bergwolf @gnawux @laijs @WeiZhang555

sboeuf:
@gnawux heads up about this PR. I know we talked about making sure we don't allow both `AddIP` and `RemoveIP` at the same time, but I got a second thought on this and we might want that to be possible.
Let's say you have already some IPs on a specific interface, well at the same time, you might want to remove some of them, and add some new ones, I don't think  there is something wrong about that, WDYT ?

network.go
```diff
@@ -88,18 +88,38 @@ func updateInterfaceMTU(netHandle *netlink.Handle, link netlink.Link, mtu int) e
...
 
 func updateLink(netHandle *netlink.Handle, link netlink.Link, iface *pb.Interface, actionType pb.UpdateType) error {
-	switch actionType {
-	case pb.UpdateType_AddIP:
+	knownActionType := false
+
+	// Add IPs
+	if (actionType & pb.UpdateType_AddIP) == pb.UpdateType_AddIP {
+		knownActionType = true
 		return updateInterfaceAddrs(netHandle, link, iface.IpAddresses, true)
```
WeiZhang555: `return` will stop the process, to achieve this purpose, I think you should call

```
if err := updateInterfaceAddrs(netHandle, link, iface.IpAddresses, true); err != nil {
    // log or return error?
}
```

WeiZhang555:
@sboeuf Hi Sebastien, I just read the previous implementation of `updateInterfaceAddrs`:

```
 57 func updateInterfaceAddrs(netHandle *netlink.Handle, link netlink.Link, addrs []*pb.IPAddress, add bool) error {
 58     for _, addr := range addrs {
 59         netlinkAddrStr := fmt.Sprintf("%s/%s", addr.Address, addr.Mask)
 60 
 61         netlinkAddr, err := netlink.ParseAddr(netlinkAddrStr)
 62         if err != nil {
 63             return fmt.Errorf("Could not parse %q: %v", netlinkAddrStr, err)
 64         }
 65 
 66         if add {
 67             if err := netHandle.AddrAdd(link, netlinkAddr); err != nil {
 68                 return fmt.Errorf("Could not add %s to interface %v: %v",
 69                     netlinkAddrStr, link, err)
 70             }
 71         } else {
 72             if err := netHandle.AddrDel(link, netlinkAddr); err != nil {
 73                 return fmt.Errorf("Could not remove %s from interface %v: %v",
 74                     netlinkAddrStr, link, err)
 75             }
 76         }
 77     }
 78 
 79     return nil
 80 }
```

if the `netHandle.AddrAdd` failed, the loop is over, so the situation could be, we try to add many IPs (and remove IP simultaneously ), then one IP failed to be added, user get an error, but some IP is still added successfully?  How do user know which some failed and which some succeed? 

This is exactly what I keep worrying about many operations in single request, it's hard to guarantee the atomicity!

To solve this issue, the option we have is:
* support adding/removing many IPs in single request, return a detailed information to user, tell them IP a,b,c are added successfully, and IP d, e, f aren't added and why.
* support adding/removing many IPs in single request, keep the operation atomic, if one step failed, all failed, we rollback the added IP and keep the network device untouched as the request had never come before.  But if rollback also failed, we still need to report which IPs are added and which IPs aren't.
* don't support adding many IPs in single request, instead, we only allow add one IP/remove one IP/setMtu or so.

So maybe you havn't got time to do this, do you have a plan for solving this? What's you idea?

sboeuf:
This PR is blocked until we resolve the discussion on #56

sboeuf:
Replaced by a more global PR #126

PR #47, opened 2017-12-07
mockserver.go: replace fmt.Errorf and errors to status

teawater:
status.Errorf can set codes to error.  Then the function can run some special
error handler with the error.

Signed-off-by: Hui Zhu <teawater@hyper.sh>

### DIFF
--- a/protocols/mockserver/mockserver.go
+++ b/protocols/mockserver/mockserver.go
@@ -7,13 +7,13 @@
 package mockserver
 
 import (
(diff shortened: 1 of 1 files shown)


bergwolf:
Can you please fix this for other errors? If we want gRPC error codes, we should set them whenever possible.

teawater:
@bergwolf OK

bergwolf:
LGTM

PR #34, opened 2017-12-04
vendor: Initialize the vendoring with dep tool

sboeuf:
vendor: Initialize the vendoring with dep tool
    
    This is the first commit related to the vendoring of all dependencies.
    This will prevent from any breakage in case a dependency would break
    the backward compatibility.
    
    Fixes #32
    
    Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

PR #48, opened 2017-12-07
mockserver locking, vsock address handling and CI

bergwolf:
1. enable mock locking so that parallel client RPC does not cause data race.
2. fix vsock address handling error due to url.URL vs. gRPC address format difference.
3. enable travis ci for kata agent.
4. add vendor so that travis can run properly.


.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+#
+
+sudo: required
+dist: trusty
+
+language: go
+go_import_path: github.com/kata-containers/agent
+
+go:
+  - 1.8
```
jodh-intel: How about also adding the following?:

```
  - 1.9
  - tip
```

bergwolf: I'm OK with both. Just note that other components are also using v1.8. If we want to go with v1.9 and tip, we should change others' travis settings as well.

jodh-intel: Ah - sorry, I forgot to check the other components. I'm happy to raise PRs to add these 2 versions in for the others if we all agree to use `1.9` and `tip` though?

bergwolf: yeah, sounds good to me!

laijs:
LGTM

laijs:
@bergwolf Could you cherry-pick the #34 and update the commit. Please ensure the author name unchanged.

bergwolf:
cherry-picked #34 as the last commit and CI passed.

PR #50, opened 2017-12-11
Add README badges and report code coverage

sameo:
* Add codecov and Travis badges
* Report code coverage back to codecov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,6 @@ install:
 
 script:
   - cd ${TRAVIS_BUILD_DIR} && make test
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


jodh-intel:
lgtm

.travis.yml
```diff
@@ -22,3 +22,6 @@ install:
 
 script:
   - cd ${TRAVIS_BUILD_DIR} && make test
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
bergwolf: Do you need to ask travis to create the coverage profiles or does the bash script automatically generate them?

jodh-intel: This seems to be the canonical way to invoke codecov (and I cannot find a profile option).

bergwolf: The current cov report is unknown after merging this PR. I guess we do need go test to create the coverage profiles. cc @sameo

jodh-intel: Looks like it might be a permissions issue?

PR #55, opened 2017-12-12
[RFC] Move the storage operation out from CreateContainer()

laijs:
It makes CreateContainer() more like oci runtime create.

It also add a storage_id to indentify the storages.
The `driver` is renamed to `controller`.
The missing `controller_options` is also added.


sameo:
@laijs So the assumption here is that you can _always_ hotplug all volumes, rootfs and shared 9pfs dir?

laijs:
Not always. For example when 9pfs is configed,

1. the qemu is booted with 9pfs config
2. request start sandbox
3. request add storage for 9pfs
4. use it for containers, never remove it

sameo:
Ok, that makes sense to me. I like that change.

sameo:
LGTM

sboeuf:
@sameo @laijs I am fine with the renaming of the Storage fields, but I am not sure why we need to add some extra functions. `Add` is supposed to be done every time we create a new container and `Remove` is supposed to happen when we remove a container. What is the added value of those extra functions ?

sameo:
@sboeuf I think it's more flexible and e.g. allow us to create a sandbox w/o any storage at all and hotplug it asynchronously whenever it's ready.

sboeuf:
@sameo I haven't taken into account this case, and it actually makes sense. But I honestly still think we're not gonna save lots of time compared to the additional complexity. I have the feeling that it is better to sometimes let the agent handle things instead of doing everything from the consumer of the API.

laijs:
@sboeuf 
The importance for me: API simplicity > Agent simplicity >> runtime simplicity.
The problem of the origin API is that agent has to implement "garbage collection" for resource.

For example, container A,B,C are added sequently, B and C share a volume which is registered in B's CreateContainerRequest(). When B quits earlier than C. When should the agent unmount the mount point and hotplug-ly remove the device which was hotpluged in as a blocked?

Kata is not used on k8s only, containers may have different life cycle.

sboeuf:
@laijs well in your example, I would expect the same volume to be hotplugged twice through the VM, so that it will appear as two different volumes inside the VM and each container can use it.
I get your point about allowing different sequences though, and I usually like when we can find the more generic way to do something, but I am wondering if we're not trying to cover a case that will never need to be covered here.

laijs:
@sboeuf double mounts is safe, because the guest kernel knows it.

A block to be hotplugged twice will be considered as two different devices by the kernel in the guest. every thing will wrong.

laijs:
moved to  #64

sameo:
@laijs Closing this one as it's been superseeded by #64

PR #56, opened 2017-12-12
[RFC] make UpdateInterface() declarative

laijs:
make UpdateInterface() declarative

And add interface_id to identify interfaces.

UpdateInterface() is using bitwise operation which is
too lowel and it doesn't support remove&add IPs at the
same grpc request. So it is changed to declarative way.

On declarative way, the agent should roll back when failed.
Rolling back can also fail again, the runtime need to know
the final result of the interface in this case, so the API
also returns the interface regardless its success or not.


protocols/grpc/agent.proto
```diff
@@ -170,29 +174,29 @@ message Route {
...
-	string name = 1;
+	string interface_id = 1;
 }
 
+// interface_id MUST NOT empty
+// all the other fields should be updated when it is set
 message UpdateInterfaceRequest {
-	UpdateType type = 1;
-	Interface interface = 2;
+	string interface_id = 1;
```
sameo: Where would that `interface_id` come from? Is the runtime supposed to generate it?
Why do we need it when we have an interface name already?

sameo:
I'm already not a huge fan of the current interface update mechanism as I think it should simply pass an Interface and expect the agent to be smart enough to make it the interface end state. i.e. it should not expect the runtime to know which fields to change and which to not change.

This change makes it more complex imo, for the only benefit of being able to remove and add IPs in one shot. So for me, either we keep the current update interface API intact or we simplify it to have a stateless runtime.

protocols/grpc/agent.proto
```diff
@@ -170,29 +174,29 @@ message Route {
...
+	string name = 3;
+	repeated IPAddress remove_addresses = 4;
+	repeated IPAddress add_addresses = 5;
+	uint64 mtu = 6;
+	string hwAddr = 7;
 }
 
 message RouteRequest {
-	Route route = 1;
+	repeated Route routes = 1;
```
sameo: Why do we change that one as well?

protocols/grpc/agent.proto
laijs: yes the runtime is supposed to generate it.
I'm afraid interface name is not always reliable.
I hope every resource has its ID which is generated/controlled by runtime or even upper layer.

protocols/grpc/agent.proto
laijs: a list of route is more likely to be added rather than a single one.

sameo:
> I can change the pr to declarative with return return. WDYT @sameo ?

That would look much better imho.

sboeuf:
@sameo @laijs this sounds good to me too, but please let's make sure we all agree (cc @WeiZhang555  and @mcastelino) about this. Otherwise, we'll end up going back and forth on our agent implementation.

protocols/grpc/agent.proto
mcastelino: @laijs we used the macaddress as the identifier in clear containers as that is reliable.

protocols/grpc/agent.proto
```diff
@@ -170,29 +174,29 @@ message Route {
...
 message UpdateInterfaceRequest {
-	UpdateType type = 1;
-	Interface interface = 2;
+	string interface_id = 1;
+	string device = 2;
+	string name = 3;
+	repeated IPAddress remove_addresses = 4;
+	repeated IPAddress add_addresses = 5;
+	uint64 mtu = 6;
+	string hwAddr = 7;
```
mcastelino: The only issue with this is that we do not pack the routes into this update. Ideally we should pack the routes into this request too. There are cases where interface routes are not desired and can cause unexpected behavior if set transiently on interface creation.

protocols/grpc/agent.proto
```diff
@@ -37,9 +37,13 @@ service AgentService {
...
+	// Add interface, the interface should be removed when failed.
 	rpc AddInterface(AddInterfaceRequest) returns (google.protobuf.Empty);
 	rpc RemoveInterface(RemoveInterfaceRequest) returns (google.protobuf.Empty);
-	rpc UpdateInterface(UpdateInterfaceRequest) returns (google.protobuf.Empty);
+	// Update the interface, the agent should try to roll back the operation
+	// when failed. Return the final effective interface config regardless
+	// failed or not.
+	rpc UpdateInterface(UpdateInterfaceRequest) returns (Interface);
 	rpc AddRoute(RouteRequest) returns (google.protobuf.Empty);
 	rpc RemoveRoute(RouteRequest) returns (google.protobuf.Empty);
```
mcastelino: We are lacking a UpdateRoutes API.

protocols/grpc/agent.proto
sameo: @laijs I really hope the runtime does not need to create, store and manage IDs when there already are reliable ways to identify resources. In the network case, I think the MAC address is a reliable ID, as @mcastelino says.

protocols/grpc/agent.proto
laijs: Could you write a pr for the route?

laijs:
@sboeuf we need more people to know the changes and make a vote.

protocols/grpc/agent.proto
WeiZhang555: I have same doubt, we don't need an interface_id. MAC address should be good enough.

protocols/grpc/agent.proto
```diff
@@ -170,29 +174,29 @@ message Route {
...
+	string interface_id = 1;
 }
 
+// interface_id MUST NOT empty
+// all the other fields should be updated when it is set
 message UpdateInterfaceRequest {
-	UpdateType type = 1;
-	Interface interface = 2;
+	string interface_id = 1;
+	string device = 2;
```
WeiZhang555: What does this field mean? what device to update?

protocols/grpc/agent.proto
WeiZhang555: Actually I think a `name` is also unique and reliable ID, we can't have same `name` for two network interfaces.

protocols/grpc/agent.proto
WeiZhang555: Same discussion on `declarative API` vs `imperative API`

WeiZhang555:
First, I'm not against the `declarative` API (related: https://github.com/kata-containers/agent/issues/19), what I want is a stable and reliable API. That meas for a delarative `UpdateInterface` API, if you send  a command to agent like `Please update the interface A, add IP b and c, delete IP d, and update its name to E`, it will complete this command as a whole or failed as a whole. 

I hate so much about half successful APIs, such as you send the request to agent, and agent tell you `Oh, I have successfully added IP b and c for you, but I failed to delete IP d, so that's all, you got a new interface with additional IP b and c with its name unchanged`, or what's worse, the agent told you `Sorry, I failed to do you command, but please, guess what you got? I won't tell you, hahaha`

We do need delcarative APIs absolutely, but we should do better. Our current APIs apparently aren't good enough now. 
check my previous comments: https://github.com/kata-containers/agent/pull/43#issuecomment-349509414

protocols/grpc/agent.proto
WeiZhang555: Do we have some requirement for updating `hwAddr`？ Maybe we can add this if we meet a real need in future.

laijs:
Hello, I update the pr

make UpdateInterface() declarative

And add interface_id to identify interfaces.

UpdateInterface() is using bitwise operation which is
too lowel and it doesn't support remove&add IPs at the
same grpc request. So it is changed to declarative way.

On declarative way, the agent should roll back when failed.
Rolling back can also fail again, the runtime need to know
the final result of the interface in this case, so the API
also returns the interface regardless its success or not.

protocols/grpc/agent.proto
```diff
@@ -170,24 +174,17 @@ message Route {
...
-	None = 0;
-	AddIP = 1;
-	RemoveIP = 2;
-	Name = 4;
-	MTU = 8;
-}
-
 message AddInterfaceRequest {
-	Interface interface = 1;
+	string interface_id = 1;
```
sameo: We do not need to pass an interface ID for a network interface. As discussed, the MAC address should be sufficient to unambiguously identify a network interface.

protocols/grpc/agent.proto
```diff
@@ -170,24 +174,17 @@ message Route {
...
+	Interface interface = 2;
 }
 
 message RemoveInterfaceRequest {
-	string name = 1;
+	string interface_id = 1;
```
sameo: The MAC address should be used here.

WeiZhang555: Sorry, I need to ask,  why `name` can't identify an interface? I maybe missed something you have discussed

protocols/grpc/agent.proto
```diff
@@ -170,24 +174,17 @@ message Route {
 	string device = 3;
 }
 
-enum UpdateType {
-	None = 0;
-	AddIP = 1;
-	RemoveIP = 2;
-	Name = 4;
-	MTU = 8;
-}
```
WeiZhang555: If remove this, how to distinguish AddIP and RemoveIP ?

sameo: The idea is for the runtime to give a complete state of the interface and ask the agent to put the interface in that state. So there would not be any update operation anymore, it would only be: Update the interface to that state (including IP, mac, MTU, etc...). That really is a declarative interface: You only describe the state you want to reach, not the steps in between.

sameo:
@WeiZhang555 
>  why name can't identify an interface

It can, but from the runtime it's hard to predict what is going to be the interface name inside the vm, while the MAC address is a unique identifier across the VM boundary.

protocols/grpc/agent.proto
laijs: MAC is also a field allowed to be changed.
for example, future container migration via vm migration. Although we can remove&add to achieve it.

laijs: I hope every obj have its name which is from and can be randomly generated by upper layers

protocols/grpc/agent.proto
WeiZhang555: I get the point, this sounds so good

WeiZhang555:
@sameo Do you mean that we can leave name field empty when AddInterface and let runtime generate a predefined name for it? 

I’m ok with using Mac as identifier, but in my thought, mac or name can meet same problem, they both can be given by user or generated randomly by runtime, and they both can be updated (mac maybe doesn’t need to be updated, not sure)

protocols/grpc/agent.proto
mcastelino: @laijs the reason to use macaddress as an identifier is that it maps to the same value both outside and inside the VM. Hence we do not need to know what the predicted interface name within the VM  needs to be. That allows the name within the VM to be changed to refect the desired name of the interface within the virtualmachine. For example the VM kernel/udev rules may generate enp0s1 and we may want to change it to eth0. So in that case using the name as an identifier will not work for the first time configuration. Hence using hardware based identifiers like mac or pci b/d/f which are fully predictable makes more sense.

protocols/grpc/agent.proto
```diff
@@ -37,9 +37,13 @@ service AgentService {
 	rpc TtyWinResize(TtyWinResizeRequest) returns (google.protobuf.Empty);
 
 	// networking
+	// Add interface, the interface should be removed when failed.
 	rpc AddInterface(AddInterfaceRequest) returns (google.protobuf.Empty);
```
WeiZhang555: I was thinking one more thing for this. Do you allow `AddInterfaceRequest` has some empty field?  I would suppose we do.

For example, if AddInterface doesn't specify a MTU value, agent should set a default MTU for it. 

Then `AddInterface` should also return an `Interface` as response to tell runtime the final state of interface, to let user know which value is chosen by agent.

protocols/grpc/agent.proto
WeiZhang555: @laijs MAC can be modified, but different VM can't own same MAC, right? Or the network and route wont work.

@mcastelino Yes, I agree that MAC is a better identifier than name, you're right.

protocols/grpc/agent.proto
WeiZhang555: Since we makes `UpdateInterface` declarative, `UpdateRoutes` looks reasonable now. So yes, I think we can add `UpdateRoutes`. 

It's also fine to put it into a new PR for this

WeiZhang555: According to our previous discussion, you hope  `UpdateRoutes` can updates a group of route rules in one request, right?

Then with this `UpdateRoutes` API, I don't see any need for `AddRoute` and `RemoveRoute`. I'm sorry that I'm still not very clear about your idea

sameo:
@mcastelino Once we change the `interface_id` with a MAC, would you be ok with that PR?
@laijs Do you mind changing `interface_id` to MAC address?

bergwolf:
@sameo @mcastelino  @WeiZhang555 the only constraint with MAC address is that it MUST be unique inside a level-2 network. What if there are two NICs and each belongs to a different level-2 network. Their MAC address and other NIC properties are assigned by higher level network components. There is no guarantee that their MAC addresses must be different. If that is possible, we cannot use MAC address to identify a NIC.

WeiZhang555:
@bergwolf I think we only need to make sure MAC is unique inside same VM,  and that’s easy to do from runtime level.

bergwolf:
@WeiZhang555 It is possible that MAC is assigned by high level network components. How can a runtime make sure MAC is unique inside the same VM?

protocols/grpc/agent.proto
mcastelino: Yes. Once we have UpdateRoute we will not need Add and Remove I think. Once we add UpdateRoute this PR LGTM.

WeiZhang555:
@bergwolf by reporting error when this happens, runtime should deny the operation when MAC is conflict. This can be a limitation rule

bergwolf:
@WeiZhang555 Yes it can. But then we assume the upper layer network component should handle the error case of duplicating MAC addresses and retry with a different MAC address, which is confusing because MAC is simply an attribute of a NIC and (to the network component's knowledge) the MAC is not duplicated inside the level-2 network at all.

WeiZhang555:
@sameo @mcastelino  As I'm also not sure if we will have a scenario that allows two interface having same MAC in single VM, maybe we should keep something like a link ID as identifier?

For example 

```
$ ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
7938: veth760c9c2@if7937: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue master docker0 state LOWERLAYERDOWN mode DEFAULT group default 
    link/ether 6e:30:a8:f5:77:4b brd ff:ff:ff:ff:ff:ff link-netnsid 0
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 00:0a:f7:92:80:17 brd ff:ff:ff:ff:ff:ff
7939: tap0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master docker0 state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether 7a:62:bb:79:5a:b2 brd ff:ff:ff:ff:ff:ff
```

We keeps the link ID such as `7938` or `2` as interface identifier, so in this case we don't have the unique MAC limitation any more. 

What's your idea?  @sameo @mcastelino

sameo:
@WeiZhang555 @bergwolf 
> As I'm also not sure if we will have a scenario that allows two interface having same MAC in single VM

AFAIK several interfaces with the same MAC can only be supported if they're in different broadcast domains. I can't see how we would ever want to support that in the same VM, in a container/Kubernetes context. I feel we're trying to find a solution for a non existing problem here.

sboeuf:
Hi folks,
So what's the status on this PR ? My feeling is that we all agreed with the declarative way, which is nice, this will make things more simple both for runtime and agent.
There is the UpdateRoute() request from @mcastelino and I also have the feeling that everybody was okay about that. @laijs could you add this to your PR ?

Now the last point is about the roll over handled by the agent and the status returned at the end. I am wondering if it worth to bother doing this ? What I mean is that with the declarative way, you want to get a result (success/failure) and in case it fails, I am not sure the runtime will do something different based on the interface returned. Furthermore, knowing that we got a failure is enough to trigger a new try.

laijs:
@WeiZhang555 What do you think about removing the "rollback" in @sboeuf 's comments?

WeiZhang555:
> Furthermore, knowing that we got a failure is enough to trigger a new try.

Fair enough. As it seems impossible to guarantee atomicity of each operation even with rollback capabilities, it's OK to only return final state of interface, then user could compare the one he got and the one he desires and makes a new try when error happens.

@laijs @sboeuf

laijs:
Ok, please take it, Big thanks!

sboeuf:
No problem ;)

sboeuf:
Actually, I have been talking a lot with @egernst about this PR, and there are still a few things we'd like to clarify before we implement it completely. He's gonna submit another PR in order to propose two different options here. I think we have captured most of what we agreed on here.
I am closing this PR and it'll be good to continue this discussion on the new PR @egernst will open.

PR #60, opened 2017-12-13
pause-bin: remove signal handlers

bergwolf:
Remove signal handlers from pause-bin to prevent it from being
killed by other members of the pid namespace, causing all processes
in the same pid namespace being killed.

The agent can always kill the pause-bin with SIGKILL from outside.

### DIFF
--- a/pause.go
+++ b/pause.go
@@ -17,21 +17,10 @@ package main
 
 #define PAUSE_BIN "pause-bin"
 
-static void sigdown(int signo) {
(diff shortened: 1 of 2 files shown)


laijs:
LGTM

sboeuf:
LGTM

laijs:
How to make `codecov/project — 27.77% (-0.25%)` passed?

bergwolf:
Need unit tests...

bergwolf:
added unit test.

PR #63, opened 2017-12-13
add health check and version API

### LINKED ISSUE #58: add health check in the api
The guest/agent may stall on some cases or panic. but the vm look normal out side.
health check can be used to check this situation. In hyperstart runtime call the Version()
api periodically. In kata, we also need such mechanism.

[grpc health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) might help.

### LINKED ISSUE #61: add grpc version API
We will need a version mechanism for agent gRPC. The most likely case we will want to handle is runtime upgrading and it reconnects to an old agent. While we can keep gRPC API backward compatibility, runtime needs to know this situation and take care not to send new API requests to old agents.

We can start by having a version check API first. And it can live with the health check service.

bergwolf:
fixes #61  
fixes #58


protocols/grpc/health.proto
```diff
@@ -21,6 +21,12 @@ message HealthCheckResponse {
 	ServingStatus status = 1;
 }
 
+message VersionCheckResponse {
+	string grpc_version = 1;
+	string impl_version = 2;
```
sameo: Could we please rename that one to `agent_version` ?

bergwolf: done.

sameo:
LGTM

agent.go
```diff
@@ -317,10 +317,12 @@ func (s *sandbox) startGRPC() error {
 
 	grpcImpl := &agentGRPC{
 		sandbox: s,
+		version: version,
 	}
 
 	grpcServer := grpc.NewServer()
 	pb.RegisterAgentServiceServer(grpcServer, grpcImpl)
+	pb.RegisterHealthServer(grpcServer, grpcImpl)
```
sboeuf: Why is this call needed ? I thought this would be part of the same API.

WeiZhang555:
code LGTM, but hmmm,  codecov reports error again

sameo:
@WeiZhang555 Because the added code is not unit tested. I'll add some gogo generated unit tests, this would look better.
For now, I'm merging that one.

PR #65, opened 2017-12-13
mount: Do not resolve storage source for 9pfs

sameo:
The source is the fs tag, it's not an actual path.


sboeuf:
LGTM

laijs:
I can't see the value of EvalSymlinks(). The kernel can handle it.

sameo:
@laijs 
> I can't see the value of EvalSymlinks(). The kernel can handle it.

That's a different issue then. Please open a PR to fix that if you think the original code can be optimized.

bergwolf:
LGTM

Issue #59, opened 2017-12-13
[RFC] Always create sandbox init process and knobs for joining


laijs:
In hyperstart, the pod_init process has its own net/pid/uts/ipc ns.

And in kata, these namespaces are also essential.

Net ns for the sandbox should be always created since the container's
network should be isolated from the vm's initial network to
avoid messing.

Pid ns for the sandbox should be always created since the containers
always join this ns. We can use new pid ns for each container,
but in most cases, they are shared the same sandbox init pid ns.

Uts ns for the sandbox should be always created since the hostname
is set in it without mess it up with the vm's initial UTS.

Ipc ns for the sandbox should be always created since the containers
always join this ns. We can use new ipc ns for each container,
but in most cases, they are shared the same sandbox init ipc ns.

So sandbox init process should be always created
in its own net/pid/uts/ipc ns and let the later containers
join in. __A pr is required for it__

__In above info, we know that container should always join the net/uts ns.
And container should NOT use the guest's initial pid/ipc ns.
But container can use the shared or private pid/ipc ns.
We have knob for pid ns. Should we add a knob for the ipc ns?
Or we can make it always to be one of shared or private.__

bergwolf:
I agree that new namespaces should always be created and we let runtime specify what is shared/private via the OCI spec. The agent should check to make sure no containers join the agent's namespace (except for the user ns?).

PR #71, opened 2017-12-15
oci: Convert the OCI spec into libcontainer config

### LINKED ISSUE #21: Convert the OCI specification into libcontainer configuration
Instead of hardcoding some values of the libcontainer configuration, this should be obtained by a conversion from the OCI Spec received from `CreateContainer()`. If some values are not provided by the spec, but needed to get the agent working properly, they should be set from a set of default values.

sboeuf:
Since the agent is receiving a full OCI specification structure,
this commit simplifies the conversion to libcontainer configuration
by relying on the "specconv" package provided by libcontainer.

Fixes #21


sboeuf:
I have simplified the OCI conversion as discussed in #21 but I think that we still have some things to discuss. Particularly about the capabilities and the mounts. Right now, we have some default capabilities and some default mounts applied by the agent. But using the spec converter removes those defaults. I was wondering if we expect those data to be provided through the OCI spec, meaning the runtime is responsible to properly fill all of this, or if we expect those defaults to be applied all the time (and only if the same caps or mount is not already defined) ?

cc @sameo @WeiZhang555 @laijs @bergwolf @gnawux @mcastelino @jodh-intel

grpc.go
```diff
@@ -310,6 +356,24 @@ func (a *agentGRPC) runProcess(cid string, agentProcess *pb.Process) (pid int, e
...
 }
 
+func (a *agentGRPC) updateContainerConfig(spec *specs.Spec, config *configs.Config) error {
+	// Update shared PID namespace.
+	for idx, ns := range config.Namespaces {
+		if ns.Type == configs.NEWPID {
+			// In case the path is empty because we don't expect
+			// the containers to share the same PID namespace, a
+			// new PID ns is going to be created.
+			config.Namespaces[idx].Path = a.sandbox.sharedPidNs.path
```
sameo: I feel the comment is a little weird: If the Path is empty and `Type == NEWPID`, then this means the container wants to leave on its own PID namespace. But the code sets it to `a.sandbox.sharedPidNs.path`. AFAIU there are a few assumptions here:

1) If `ns.Path` is empty and the pod does not want shared PID ns, `a.sandbox.sharedPidNs.path` is going to be empty, and this is a NO-OP.
2) If `ns.Path` is NOT empty, then we're overwriting is with the shared PID ns, i.e. shared PID ns takes precedence over a specific PID ns the container may want to join. AFAIK this is not documented anywhere.
3) If the shared PID ns does not exist at that point, libcontainer will create it.

Could you please add more comments to make those assumptions more explicit ?

grpc.go
```diff
@@ -310,6 +356,24 @@ func (a *agentGRPC) runProcess(cid string, agentProcess *pb.Process) (pid int, e
...
+			break
+		}
+	}
+
+	// Add the value for NoNewPrivileges option.
+	config.NoNewPrivileges = spec.Process.NoNewPrivileges
+
```
sameo: I'd prefere to see 2 different routines, one for updating the PID ns path, and another one for the privileges.

sboeuf: Done

grpc.go
sboeuf: I have added more comments. Also, a follow up PR should actually take care of the case where the spec gives us some paths for the namespace types defined. Right now, we ignore any ns paths since we don't know how to handle it.
We can make something simple but the question is how k8s would handle a shared ns case through the spec ? I mean that we need to decide from where the information should come in case of shared ns.
For instance, if k8s uses properly the spec to share the same ns between containers, then we only need to rely on the spec, instead of passing a specific parameter when creating the sandbox (as @laijs proposed last week IIRC).

grpc.go
```diff
@@ -310,6 +356,51 @@ func (a *agentGRPC) runProcess(cid string, agentProcess *pb.Process) (pid int, e
...
+// sandbox information. When the sandbox is created, it can be setup in a way
+// that all containers will share some specific namespaces. This is the agent
+// responsibility to create those namespaces so that they can be shared across
+// several containers.
+// If the sandbox has not been setup to share namespaces, then we assume all
+// containers will be started in a new namespace.
+// The value of a.sandbox.sharedPidNs.path will always override the namespace
+// path set by the spec.
+// TODO: This needs to be handled in a better way, so that we don't ignore
+// namespace paths from the spec.
```
bergwolf: I'm afraid we **SHOULD** ignore them. Those fields are used to share namespaces on the same host for docker containers. With kata agent we better fit in the kubernetes pod model that shared/private namespaces are both at a pod level. IOW, we cannot share a namespace from a container that lives in a different sandbox. If some containers shares a namespace, they should share the pod level one.

Then we can implement as @laijs suggested in #59 that the agent always creates pod level namespaces. And if a container wants to share some namespace, it joins the pod level one.

fyi, the k8s pod spec sets namespace sharing at the pod level. See https://v1-8.docs.kubernetes.io/docs/api-reference/v1.8/#podspec-v1-core.

sboeuf: @bergwolf yes you're right, I was not thinking about those drawbacks and limitaitons. Then let me remove the `TODO` from this comment, and I think the PR is ready then.

grpc.go
bergwolf: @sboeuf what about the case that no `ns.Type == configs.NEWPID` set in spce? Which pidns does the container live in when libcontainer creates it?

sboeuf: Well in that case, the container will share the same PID ns than the agent. I don't think we want this case to happen, and that's why I was asking if we should consider some default values that we would apply from the agent every time.

bergwolf: Yeah, we don't want it to happen. pidns should either be the pause process one, or containers live on their own. I think the default value is controlled by `SandboxPidns` field in CreateSandbox() request.

sameo: @sboeuf Agree with @laijs  and you here: We don't want the container to live in the agent pidns. So depending on the `SandboxPidns` value, the default should be the pause process one or we create a new pidns for each container.

sboeuf: I would prefer that by default, if we have no `ns.Type == configs.NEWPID` from the list of namespaces provided by the spec, we actually add one from the agent. And then, we apply the `sandboxPidns`. This means that we make sure the container will end up in its own PID ns, or in the pause container one if `sandboxPidns` is not empty. Here, we would be adding a security in case the runtime does not provide this `NEWPID` type.
Are you all okay with that ? @sameo @bergwolf @laijs 
The other solution would be to throw an error in case the `NEWPID` type is not provided by the spec.

bergwolf: Your proposal sounds good to me @sboeuf. I would prefer that agent handles it rather than throwing errors when `NEWPID` is not set.

sboeuf: @bergwolf ok great, I'll add this to the code then!

sboeuf:
PR updated !

bergwolf:
LGTM. @sboeuf you may need to add some unit tests to pass codecov.

sboeuf:
@bergwolf yep thanks for the comment ! I have updated the PR with some unit tests, now codecov is happy ;)
@sameo @laijs  could you take a look at this please.

Issue #38, opened 2017-12-05
api: CreateContainer() should return NewProcessResponse


laijs:
The pid for the init-process should be known when runtime create return.
See runc create --pid-file

In our case the pid is the shim-pid, and shim's arguments needs the process's pid inside the vm.
The way to spawn the shim can be changed to fix it. But the better solution is fixing the API.

laijs:
@sameo @sboeuf @bergwolf

sboeuf:
@laijs well you mentioned the reason why we don't have the pid. That's because we are providing the shim's pid, and this is the runtime responsibility. We're trying to implement properly what `runc` has to offer, but I don't see why we should try to make the gRPC API reflecting `runc` API.
The CreateContainer() does not start the container process, and for this reason, we cannot provide a PID in response of this. And there is no reason to try to provide the shim PID through this API, those are different things. The guest PIDs from inside the VM should not be mixed up with the host PID of the shim, this would bring more confusion IMO.

laijs:
but `runtime exec` is allowed before `runtime start`, how do you achieve this?

sboeuf:
@laijs is it ? This is very weird since you should not be able to `exec` a process if the container process itself is not running, right ?
Now, let's say it is acceptable (I really think it's not but anyway), we can still do it by calling into `exec` after `create`, we don't need the PID of the container for that, we only need the container ID.
@laijs am I missing something here ? I ask because I am trying to understand what you're trying to achieve with this.

laijs:
runc allows this behavior.
We'd better to follow the way that runc calls into libcontainer, unless there is a strong demand with convictive reasons.

laijs:
@sameo @sboeuf The shim should be created after `runtime create` returns before `runtime start`
In this case, the shim will be started without NewProcessResponse, so a special tunnel has to be set up to tell the shim about it. It complicates everything.

CreateContainer() returning NewProcessResponse or runtime generating process ID can solve this problem.

sboeuf:
@laijs I am sorry but I am gonna repeat myself on this one, why do you need the ID from the agent for this case ?
The shim will be started by `runtime create` and you will have the real PID in your hand so that you can fill the pid file properly. Why is that not enough ?

laijs:
> The shim will be started by `runtime create` and you will have the real PID in your hand so that you can fill the pid file properly. Why is that not enough ?

[...]

PR #72, opened 2017-12-19
[RFC] grpc: use runtime generated process ID

bergwolf:
Currently we use agent generated PID to identify a process inside
a container, which has following limitations:

1. A process pid can be reused. The PID runtime got might refer to
   a totally different process when referencing it in later RPC calls.

2. The current use of map[pid]process inside the agent is problematic.
   We have to maintain a process structure until it is wait()ed. If
   another process is created with the same pid (after the previous one
   dies), this model is failing to properly maintain process mappings.

3. We might add a timestamp field to solve issue 1 and 2. But then the
   the timestamp shall live through all stack and the kata runtime API
   also becomes complicated.

So how about just getting back to a runtime generated process ID that
uniquely identifies a process inside a container?

Then we use <containerid, processid> to identify a process rather than
using <contaienrid, pid, timestamp> inside a sandbox, removing the need
of timestamps all together.

This is WiP. Once we agree on the gRPC interfaces, I'll go ahead changing the agent implementation.

cc @sameo @sboeuf  @laijs @WeiZhang555


sameo:
@bergwolf A few comments here:

1. When talking about PID reuse, we're talking about something that typically happens after 32768 processes have bee started. That's quite unusual in a container context
2. Then about the `map[pid]process` table managed by the agent: As long as a process has not been reaped, it consumes a PID in the kernel process table. So even if it terminates, its PID won't be reused until the agent wait for it, so we will never get to create a new entry in that table for a PID that's already in use.
3. If the runtime addresses a PID in the VM, I think it's safe to assume that he's going to wait for it at some point and should be aware of the fact that it's terminated or not.

laijs:
@bergwolf I wrote a draft in https://github.com/laijs/kata-agent/commit/abaff59a3564339a30cc4b9faa884d194f8395ee which is using in the https://github.com/hyperhq/runv/pull/651
you can move some comment from it.

laijs:
@sameo The agent itself is a subreaper, the pid is almost immediately reaped.

bergwolf:
@sameo about your comments:
1.  It is a bit assumption that pid will not be reused in container. We cannot control what is running inside.
2.  I was talking about `processes   map[int]*process` in `container` struct. It is not deleted until the runtime wait() the process, which is different from agent waiting the real process and kernel reaping process resources.
3.  the agent does not know when a runtime will wait a process, so agent has to maintain the process in `map[int]*process` for a unknown period of time. That is the reason for `processes   map[int]*process` to be problematic.

bergwolf:
Thanks @laijs. I've included your comment in the proto file.

laijs:
@sameo The runtime should use and depend on the reliable identifications.
guest pid, mac address, etc... all are unreliable, they can be messed up by the hacked guest.

A UUID or any generated ID from the host-side for process/interface... can be considered as C-lang pointers, which makes life easy for low-level programming. We want to keep agent simple, do we?

sameo:
@bergwolf 
> It is a bit assumption that pid will not be reused in container. We cannot control what is running inside.

Sure, all I'm saying is that we're looking at adding runtime and agent complexity for a use case that's really rare.

> I was talking about processes map[int]*process in container struct. It is not deleted until the runtime wait() the process, which is different from agent waiting the real process and kernel reaping process resources.

Ok, sorry I misunderstood what you meant. The process map issue you're pointing at is valid and could be handled without having to push the complexity of generating PIDs down to the runtime (because the guest kernel already does that for us). We could for example have a zombie table in the agent for keeping the reaped but not waited for processes, and move processes there when a process is reaped by the agent but the runtime has not had time to wait for it. And yes, this would not entirely fix the overlapping issue, only postponing it to a 2nd overlap, i.e. typically after 64K processes.

sameo:
@laijs 
> guest pid, mac address, etc... all are unreliable, they can be messed up by the hacked guest.

If your guest is hacked in such a way that it can generate PIDs as it wants, it basically has full ring 1 control. In such case I'm pretty sure sending a host generated UUID won't fix anything. Also, the agent will always have to map those generated UUIDs to the actual guest resources so I'm not sure how this is simpler than using those resources directly.

bergwolf:
@sameo Your proposal can workaround the mapping table issue but it is more complex than introducing an ID to the agent protocol. Notice that you'll need extra locking to ensure the "moving" of a process is atomic.

In fact there is no extra complexity for the agent if we add ProcessID, -- it just uses the ProcessID rather than current pid to map the process, and that's all.

We are designing the protocol from ground. It is not good to have defects in the first place (be it 32K or 64K limitations). Plus we don't need the zombie process table workaround if we can fix the root cause. That's why I think ProcessId is a better solution.

laijs:
the guest can trigger some bugs of the runtime by deliberately fabricated pids.

sboeuf:
@laijs @bergwolf if the concern is about the reuse of __real__ PID number (I see the point of overlap that you've been describing), then why don't we just maintain a list of IDs from the agent ? Your concern comes from the fact that we don't control the assignment of real PIDs. But we can maintain our own list of PIDs, and I don't see the benefits of doing this from the runtime.

sboeuf:
Moreover, the runtime being a short time process, how are you planning to know the PID that you have been already using ? I guess we will have to store some information through the filesystem, which would end up decreasing the performances.
The agent being always up and running, it's easier to maintain list of objects, because there is no need to go through the filesystem for that.

sboeuf:
Sorry, I hit the wrong button.....

bergwolf:
@sboeuf If I understand your proposal correctly, I don't think ListProcess() can solve the problem. Like StartContainer(), the result returned by ListProcess() is still unreliable because a process can die after ListProcess() and kernel can reuse the same pid for another process.

It is not about controlling PID assignment, it is guest's process pid itself cannot reliably point to the same process with time passing by. 

> Moreover, the runtime being a short time process, how are you planning to know the PID that you have been already using ? I guess we will have to store some information through the filesystem, which would end up decreasing the performances.

The runtime has to store information on disk per OCI spec (e.g., the state.json file). And for ProcessID case, runtime can generate a random ProcessID string, and handle collision errors returned from the agent, which would be really rare because ProcessID collision only happens inside the same container.

sameo:
@bergwolf Sorry for the delayed reply...
> We are designing the protocol from ground. It is not good to have defects in the first place (be it 32K or 64K limitations). Plus we don't need the zombie process table workaround if we can fix the root cause. That's why I think ProcessId is a better solution.

I understand your point and I agree we should aim at getting things right from the beginning. But we also have to find the right balance between supporting and handling all possible use cases and keeping the API simple and intuitive. IMHO your proposal makes the API more complex and less natural: Why, as a caller, would I need to specify a process ID when each process already has an ID? As a user of this API, I would find that approach awkward and would wonder why I have to maintain and generate process IDs, pretty much like the kernel.
If I understand your concern properly, it's about keeping all process IDs unique as long as the VM lives. Again, I think this is a valid concern but I believe the component that manages all processes is the agent and as such it should generate a unique ID for each process it spawns. So what about changing `NewProcessResponse` to:

```
message NewProcessResponse {
	uint32 PID = 1;
        string  exec_id = 2;
}
```

and then changing the request messages like that e.g.:

```
message SignalProcessRequest {
	string container_id = 1;
	string exec_id = 2;
	uint32 signal = 3;
}
```

bergwolf:
>  Why, as a caller, would I need to specify a process ID when each process already has an ID? 

@sameo It's actually pretty common for caller of an API to specify an ID when creating an object. E.g, the container ID is specified by runtime when calling `CreateContainer()`. It's irrelevant to whether a process has a unique ID of its own or not. I'll open an issue to consolidate with the interface ID discussion. I believe we can enforce the same design pattern in all the creation cases.

sameo:
@bergwolf 
> It's actually pretty common for caller of an API to specify an ID when creating an object. E.g, the container ID is specified by runtime when calling CreateContainer()

It is for objects that are not already managed by the OS (host or guest) like e.g. processes, NICs, disks, etc...Those are already identified.

I think we should bring that discussion to the arch committee to get some additional input. 

> It's irrelevant to whether a process has a unique ID of its own or not.

IIUC you opened that PR for addressing the overlapping PID issue and how we consequently managed the process map inside the agent.

bergwolf:
@sameo yes I agree we should consult the arch committee. I raised the PID collision issue as an example of why internal structures are not generally suitable for unique object identification -- they are prone to collisions.

sameo:
@bergwolf I feel that not generating IDs from the runtime would make things simpler, but it's only a personal opinion. I'm happy to go with whatever the arch committee may decide upon.

bergwolf:
@sameo yup, I'm happy to take whatever the committee decide on as well. We've been going back and forth on this kind of topic for too long. We should just get a consensus and push the project forward.

sboeuf:
@bergwolf The proposal about `ListProcesses()` is completely unrelated to this topic, this would be used in case we want to implement the equivalent of `ps` as proposed by `runc`.
Now about this process ID, as I mentioned before,
> if the concern is about the reuse of real PID number (I see the point of overlap that you've been describing), then why don't we just maintain a list of IDs from the agent ? Your concern comes from the fact that we don't control the assignment of real PIDs. But we can maintain our own list of PIDs, and I don't see the benefits of doing this from the runtime.

What do you think about this ? Do you have other concerns ?

sboeuf:
According to the discussion on #38 and given the fact that we're seeking for simplification and not making things more complicated, I think this current proposal should be okay.
As @laijs and @bergwolf mentioned before, we could generate the process ID from the runtime, and it will simplify things for the agent. The good point solved by this approach is that we don't need #38 anymore since the runtime would be aware about the process ID for the container. This way, it will be able to provide the ID to the shim.
@sameo is that okay for you ?

sboeuf:
I can implement that if you want BTW

sboeuf:
Well it's actually more complicated, we also need to provide the process ID through `CreateContainer()`, otherwise, when the shim will be created (before the call into `StartContainer()`), it will try to do `WaitProcess()` and `ReadStdout()`, but those will fail since the agent won't have anything related to the specific process ID.

sameo:
So I've been implementing the kata shim support into virtcontainers, and I now understand better why it makes sense to use an ID for a container process, one that is different from the guest kernel PID.
If we want to be able to create a container and start the shim right away (Doing otherwise would be racy), we need to either pass a runtime generated ID or get an agent generated ID back in order to have the shim use that as it's monitored process identifier.
I'm ok with having the runtime generating that ID, because it's different from what we mean by process ID (PID) which is a kernel object. But:

* I'd like to use a different name. `process_id` is confusing, we're not talking about an actual PID. I'd prefer if we use something like `exec_id` or `task_id`
* For consistency sake, `CreateContainerRequest` should also take an ID, we should not rely on some implicit naming for the init process.

bergwolf:
@sboeuf @sameo thanks for the feedback! I am OK with `ProcessID` renamed to `exec_id` or `task_id`. 

@sboeuf in `runv shim`, it waits a signal before doing any actual shim work if it finds out itself is a shim for the container init process. I had `kata-shim` to do the same but then the `ProcessID` was removed from agent API so I had to drop it. I'll add that back once we have a name for the container init process.

@sameo in that case I think it makes sense to use either `container_id` or a constant name for the container init process, just so that it is easier for the shim to discover that itself is an init process shim. Another approach is to pass a boolean argument to let shim know that it is an init process shim. Which one do you like better?

bergwolf:
@sboeuf and I discussed this on irc and we think we can proceed with following approach:

1. the runtime creates process ID (named `exec_id`) and pass it through `ExecProcessRequest`
2. for container init process, use `ContainerId` in `CreateContainerRequest` as its process id
3. kata shim will wait for SIGUSR1 before proceeding if it finds out itself an init process shim (by comparing containerid and processid arguments)
4. the runtime should send SIGUSR1 to shim after calling `StartContainer`
5. add `exec_id` to `CreateContainerRequest` that must equal to containerID for kata runtime

A bit more explanation about adding `exec_id` to `CreateContainerRequest`. It is to make the agent neutral because `ContainerId == ExecId` for the init process is a requirement for kata runtime/shim. It opens the possibility for runtime/shim implementation to have a different approach to identify the container init process, such as using a `boolean` to tell if shim if it is for init process.

@laijs @sameo @WeiZhang555 @gnawux What do you think?

sboeuf:
#83 implements those changes for the agent.

sameo:
@bergwolf PR #83 looks good to me, I think it implements what you describe.

About sending SIGUSR1 to the shim to notify it that the container is started, I'm ok if we start with that approach but I think it is not necessary. It's a bit of an undocumented, OOB protocol.
IIUC, we need this because we don't want the shim to call `Wait` or `Read/Write` on stdin/stdout/stderr. Please correct me if I'm wrong. If not, I believe we could handle those operations from the agent:

- `Wait` is easy. We won't return anyway until the process exits, so nothing will happen before `StartContainer`.
- `WriteStdin` should have the agent return an error if the container is not started. Writing on a process stdin if it does not even exists is simply an error.
- `ReadStdout/ReadStderr` is probably the most complex case to handle from the agent, but I believe we could make it blocking until we do get an stdout handler for the corresponding process, from the agent itself.

Am I missing something?

bergwolf:
replaced by #83

bergwolf:
sorry just saw @sameo's comments. We start `shim` before calling `StartContainer()` to get shim pid. If we do not pause the shim process, there might not be any running container yet when `shim` calls any agent APIs.

bergwolf:
sorry just saw @sameo's comments. We start `shim` before calling `StartContainer()` to get shim pid. If we do not pause the shim process, there might not be any running container yet when `shim` calls any agent APIs.

bergwolf:
Close this in favor of #83. We can continue discussion though.

bergwolf:
@sameo Please note that shim waiting on signal is not related to agent gRPC protocols. It is about supporting `runc create --pid-file value` option that requires getting shim's pid before starting the container, as discussed by @laijs and @sboeuf in #38

PR #78, opened 2017-12-21
change most fmt.Errorf to "google.golang.org/grpc/status"

teawater:
grpcStatus implements errors returned by gRPC.  These errors are
serialized and transmitted on the wire between server and client, and allow
for additional data to be transmitted via the Details field in the status
proto.  gRPC service handlers should return an error created by this
package, and gRPC clients should expect a corresponding error to be
returned from the RPC call.

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -25,6 +25,9 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"golang.org/x/sys/unix"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	grpcStatus "google.golang.org/grpc/status"
 )
 
 type agentGRPC struct {
@@ -175,7 +178,7 @@ func (a *agentGRPC) Version(ctx context.Context, req *pb.CheckRequest) (*pb.Vers
 
 func (a *agentGRPC) getContainer(cid string) (*container, error) {
 	if a.sandbox.running == false {
-		return nil, fmt.Errorf("Sandbox not started")
+		return nil, grpcStatus.Error(codes.FailedPrecondition, "Sandbox not started")
 	}
 
 	ctr, err := a.sandbox.getContainer(cid)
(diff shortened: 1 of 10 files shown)


laijs:
LGTM

grpc.go
```diff
@@ -23,6 +23,9 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"golang.org/x/sys/unix"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
sboeuf: Please rename the import with something like:
```go
import (
        ...
        grpcStatus "google.golang.org/grpc/status"
)
```
This way, we won't have to rename the variables `status` which I like.
You could do the same with `codes` into `grpcCodes` so that it won't conflict in the future.

sameo:
@teawater It looks good to me, but could you please add a one line commit message explaining how this is better than `fmt` based errors?

grpc.go
```diff
@@ -274,7 +277,7 @@ func (a *agentGRPC) runProcess(cid string, agentProcess *pb.Process) (pid int, e
 	defer a.sandbox.subreaper.RUnlock()
 
 	if err := ctr.container.Run(&(proc.process)); err != nil {
-		return -1, fmt.Errorf("Could not run process: %v", err)
+		return -1, status.Errorf(grpc.Code(err), "Could not run process: %v", err)
```
laijs: http://127.0.0.1:6060/pkg/google.golang.org/grpc/#Code says grpc.Code() is deprecated.
But it still seems that grpc.Code() is fit here.

WeiZhang555:
And can you squash all the commits? We don’t need one commit per file :)

egernst:
@teawater  -- anything new on this PR?

jodh-intel:
Hi @teawater - can you rebase this PR please?

egernst:
@teawater - and updates here?  If we do want to make these changes, which isn't clear, then perhaps other folks can step in and finish this PR/issue?

teawater:
I am working on this issue and will update branch later.

jodh-intel:
lgtm.

wdyt @sboeuf?

sboeuf:
LGTM

PR #83, opened 2017-12-22
protocols: grpc: Rely on exec_id provided by the runtime

### LINKED ISSUE #82: Get exec_id from the runtime
Instead of generating exec IDs (or process IDs) from the agent, we will rely on the runtime to make things simpler.

sboeuf:
protocols: grpc: Rely on exec_id provided by the runtime
    
    This commit radically changes the way the process IDs are generated.
    Indeed, for simplicity, the process ID, called exec_id, will always
    be provided by the runtime.
    
    The shim needs to be started knowing which exec_id it actually
    represents. By having this information from the runtime, we make
    things simple since the runtime will provide this directly to the
    shim, with no need to interact with the agent.
    
    Also, this commit stores the exitCode channel from the caller, so
    that we don't need to keep it into a specific map for too long. This
    prevents from the issue of continuing to use a map with a specific
    PID entry, while this PID could have been re-assigned by the kernel.
    
    Fixes #82
    
### DIFF
--- a/agent.go
+++ b/agent.go
@@ -27,13 +27,14 @@ import (
 )
 
 type process struct {
-	id          int
+	id          string
 	process     libcontainer.Process
 	stdin       *os.File
 	stdout      *os.File
 	stderr      *os.File
 	consoleSock *os.File
 	termMaster  *os.File
+	exitCodeCh  chan int
 }
 
 type container struct {
@@ -43,7 +44,7 @@ type container struct {
 	initProcess *process
 	container   libcontainer.Container
 	config      configs.Config
-	processes   map[int]*process
+	processes   map[string]*process
 	mounts      []string
 }
 
@@ -63,8 +64,9 @@ type sandbox struct {
 }
 
 type namespace struct {
-	path string
-	init *os.Process
+	path       string
+	init       *os.Process
+	exitCodeCh <-chan int
 }
 
 var agentLog = logrus.WithFields(logrus.Fields{
@@ -122,15 +124,15 @@ func (p *process) closePostExitFDs() {
 	}
 }
 
-func (c *container) setProcess(pid int, process *process) {
+func (c *container) setProcess(execID string, process *process) {
 	c.Lock()
-	c.processes[pid] = process
+	c.processes[execID] = process
 	c.Unlock()
 }
 
-func (c *container) deleteProcess(pid int) {
+func (c *container) deleteProcess(execID string) {
 	c.Lock()
-	delete(c.processes, pid)
+	delete(c.processes, execID)
 	c.Unlock()
 }
 
@@ -145,13 +147,13 @@ func (c *container) removeContainer() error {
 	return removeMounts(c.mounts)
 }
 
-func (c *container) getProcess(pid int) (*process, error) {
+func (c *container) getProcess(execID string) (*process, error) {
 	c.RLock()
 	defer c.RUnlock()
 
-	proc, exist := c.processes[pid]
+	proc, exist := c.processes[execID]
 	if !exist {
-		return nil, fmt.Errorf("Process %d not found (container %s)", pid, c.id)
+		return nil, fmt.Errorf("Process %s not found (container %s)", execID, c.id)
 	}
 
 	return proc, nil
@@ -181,7 +183,7 @@ func (s *sandbox) deleteContainer(id string) {
 	s.Unlock()
 }
 
-func (s *sandbox) getRunningProcess(cid string, pid int) (*process, *container, error) {
+func (s *sandbox) getRunningProcess(cid, execID string) (*process, *container, error) {
 	if s.running == false {
 		return nil, nil, fmt.Errorf("Sandbox not started")
 	}
@@ -200,7 +202,7 @@ func (s *sandbox) getRunningProcess(cid string, pid int) (*process, *container, 
 		return nil, nil, fmt.Errorf("Container %s %s, should be %s", cid, status.String(), libcontainer.Running.String())
 	}
 
-	proc, err := ctr.getProcess(pid)
+	proc, err := ctr.getProcess(execID)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -208,8 +210,8 @@ func (s *sandbox) getRunningProcess(cid string, pid int) (*process, *container, 
 	return proc, ctr, nil
 }
 
-func (s *sandbox) readStdio(cid string, pid int, length int, stdout bool) ([]byte, error) {
-	proc, _, err := s.getRunningProcess(cid, pid)
+func (s *sandbox) readStdio(cid, execID string, length int, stdout bool) ([]byte, error) {
+	proc, _, err := s.getRunningProcess(cid, execID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


sboeuf:
cc @sameo @bergwolf @laijs @WeiZhang555 @jodh-intel 
Hey guys I have implemented the things discussed with @bergwolf on #72 
I am happy about the changes except for one part, that is the handling of the execution of new processes through `exec.Cmd.Start()` for instance. In that case, we don't have an `execID` provided by the runtime since this is internal to the agent... I have handled this by pulling a package that we have in `virtcontainers/pkg/uuid`, so that I could generate some unique IDs. We can move this package to one of the Kata repo, I am fine with that, but I wanted to solve the PR before to bother thinking about a place for this package. Anyway, this package is relying on crypto/rand which means there is no chance (very tiny maybe) we will end up with the same ID than one of the ID provided by the runtime.
Let me know if you think this is the right approach.
The other option would be to ignore any exit code returned by the execution of `exec.Cmd.Start()` and `exec.Cmd.Wait()`. This way, we would not need to handle this properly through the `reaper`, we would only ignore the err returned by `Wait()` function.

sameo:
@sboeuf This LGTM.
And generating a UUID for the internal agent spawned process is fine imo. You can copy the virtcontainers uuid package to avoid adding a virtcontainers dependency.

sboeuf:
@sameo if that's fine with @bergwolf @laijs and @WeiZhang555 , I'll copy it :)

bergwolf:
LGTM! Thanks @sboeuf! And I fine with copying the uuid package from virtcontainers.

sboeuf:
@bergwolf @sameo I have updated the PR with the fix for the mockserver. I have also got rid of virtcontainers dependency by copying this `uuid` package to `pkg/uuid`.

bergwolf:
One small nit, @sboeuf would you mind splitting the PR in smaller commits? e.g., gRPC change, agent change, mock server change and uuid package change can each live in separate commits.

sameo:
@bergwolf grpc and agent changes should live together as otherwise it will break bisection.

bergwolf:
@sameo fair enough. I can accept the PR as is.

reaper.go
```diff
@@ -21,7 +22,8 @@ type reaper struct {
 	sync.RWMutex
 
 	chansLock     sync.RWMutex
-	exitCodeChans map[int]chan int
+	exitCodeChans map[string]chan int
+	pidToExecID   map[int]string
```
laijs: I would like to keep using `exitCodeChans map[int]chan int` with an adapted reaper:

1) the reaper loop automatically delete the entry when the pid is reaped.
  (simplify codes and also fix the problem of reusing pid by the kernel).
2) the `chan int` is also saved on the process struct for WaitProcess(). 
   (don't need to search it from the reaper again)
3) reaper.start() returns `chan int`, the caller should wait for it before exec.Cmd.Wait() or other Wait().
4) remove reaper.wait() and corresponding interfaces.

sboeuf: @laijs thanks for your comments, I agree it makes things simpler. I have addressed 1, 2, 3, but not 4 because I have made the choice to pass the `exitCodeCh` as an argument of the reaper function `wait()`, so that we can keep the generic call here. No need to duplicate from the caller.

sboeuf:
@laijs @bergwolf codecov was not passing because of the updated code of `health.pb.go` but I don't get why since this file should not be tracked. I have added `codecov.yml` in order to ignore `protocols/grpc`, but I am not sure this is the right approach.

reaper.go
```diff
@@ -119,34 +116,31 @@ func (r *reaper) reap() error {
 // start starts the exec command and registers the process to the reaper.
 // This function is a helper for exec.Cmd.Start() since this needs to be
 // in sync with exec.Cmd.Wait().
-func (r *reaper) start(c *exec.Cmd) error {
+func (r *reaper) start(c *exec.Cmd) (chan int, error) {
```
laijs: returns the receiving channel `<-chan int`

sboeuf: PR updated.

reaper.go
```diff
@@ -119,34 +116,31 @@ func (r *reaper) reap() error {
...
 // from the subreaper, the exit code is sent through the provided channel.
 // This function is a helper for exec.Cmd.Wait() and os.Process.Wait() since
 // both cannot be used directly, because of the subreaper.
-func (r *reaper) wait(pid int, proc waitProcess) (int, error) {
-	exitCodeCh, err := r.getExitCodeCh(pid)
-	if err != nil {
-		return -1, err
-	}
-
+func (r *reaper) wait(exitCodeCh chan int, proc waitProcess) (int, error) {
```
laijs: here is also receiving chan ...

reaper.go
```diff
@@ -108,6 +101,10 @@ func (r *reaper) reap() error {
 			continue
 		}
 
+		// Let's delete the entry here since the channel has been
+		// stored by the caller, in order to wait for the exit code.
+		r.deleteExitCodeCh(pid)
+
```
laijs: close the chan after sent the exit code. (I can't comment on unmodified code, so I comment here)

reaper.go
laijs: and several other places...

sboeuf: Wait, I have a problem with that. From `wait()` I get the following error message:
```
./reaper.go:154: invalid operation: close(exitCodeCh) (cannot close receive-only channel)
```
Don't you think we can stay with those channels without detailing the direction ?

agent.go
```diff
@@ -122,15 +124,15 @@ func (p *process) closePostExitFDs() {
 	}
 }
 
-func (c *container) setProcess(pid int, process *process) {
+func (c *container) setProcess(execID string, process *process) {
 	c.Lock()
-	c.processes[pid] = process
+	c.processes[execID] = process
```
laijs: StartContainer() calls a.runProcess(req.ContainerId, "", nil).
Will the control flow go to here?

sboeuf: What do you mean by control flow ? `runProcess()` is a global function designed to handle both `StartContainer()` and `ExecProcess()`

laijs: the execID is "" in this case.
Should container init process save in the c.processes? I'm Ok with any answers.
Is c.processes[""] = process.  (I never try it before ... let me check it in the playground)

reaper.go
laijs: it is not a good idea to close it in reaper.wait().
if it needs to be closed, the closing operation should be in the reaper loop.

agent.go
laijs: Sorry, ignore this...

reaper.go
sboeuf: If the closing operation happens in the reaper loop, are you sure the exit code will be received even if the `Wait()` is called later ?

sboeuf: Nevermind, I have tested it and it's okay, we can actually close the channel after we sent something, the data will be read at some point by the Wait().

sboeuf:
@laijs PR updated. I have changed all the channels to appropriate direction channels.

sboeuf:
@laijs if you're okay with the PR, please merge it without taking codecov into account. We need another PR to introduce some ignored paths for codecov, otherwise a lot of PR will have some troubles passing.

laijs:
LGTM

PR #92, opened 2018-01-11
agent: allow exec before start

bergwolf:
1. create container init process in CreateContainer rpc
2. call container.Exec() to start container
3. call container.Run() to exec a new process in container
4. do not reject exec/stdio in a created state container

This follows the way runc uses libcontainer. And shim does
not need to hold itself waiting start signals.

Fixes: #38


agent.go
```diff
@@ -198,8 +198,8 @@ func (s *sandbox) getRunningProcess(cid, execID string) (*process, *container, e
 		return nil, nil, err
 	}
 
-	if status != libcontainer.Running {
-		return nil, nil, fmt.Errorf("Container %s %s, should be %s", cid, status.String(), libcontainer.Running.String())
+	if status == libcontainer.Stopped {
```
sameo: We may want to rename `getRunningProcess` to `getCreateProcess` or simply `getProcess` then.

bergwolf: Done.

bergwolf:
@sboeuf in libcontainer.Start(), the container process is created but put on hold waiting execfifo so there is nothing to be reaped. The container process only proceeds to execute when libcontainer.Exec() is called.

see https://github.com/opencontainers/runc/blob/master/libcontainer/standard_init_linux.go#L172
and https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L270

> Also, the process is going to be started during the CreateContainer(), I don't get what's the point of the StartContainer() now.

The process is only created `CreateContainer()` but not really started. It is still started (exec()ed) in `StartContainer()`. The idea is to follow runc's usage of libcontainer.

sboeuf:
@bergwolf oh I missed that part between `Start()` and `Exec()`. Based on your explanation of the behavior, I agree your change is fine then !

Issue #70, opened 2017-12-15
vendor: Update libcontainer vendoring version


sboeuf:
We should bump the vendoring version of libcontainer in order to get the fix https://github.com/opencontainers/runc/pull/1678 included in our code.
This will prevent libcontainer from waiting SIGKILL'ed processes.

vbmade2000:
Can I work on this ?

jodh-intel:
Hi @vbmade2000 - sure! You probably want to read the following if you haven't already done so:

- https://github.com/kata-containers/agent/blob/master/CODE_OF_CONDUCT.md

jodh-intel:
... also, although it's not actually part of the kata project, you might want to have a read of https://github.com/clearcontainers/runtime/blob/master/CONTRIBUTING.md as that's the process we're tending to follow (we need to add CONTRIBUTING.md docs to the kata repos "real soon now" (TM) :smile:)

vbmade2000:
@jodh-intel Sure. I'll take a look on both links. If the same https://github.com/clearcontainers/runtime/blob/master/CONTRIBUTING.md file you want to put then I can put it in this repo :)

jodh-intel:
Thanks @vbmade2000.

---

Regarding the contributing file, that would of course need to be a different PR. We'll need to get agreement from the team of course.

I was in fact imagining we would store a central `CONTRIBUTING.md` file in one of these two repos:

- https://github.com/kata-containers/documentation
- https://github.com/kata-containers/community

We would then then create a 1-line "stub" `CONTRIBUTING.md` file in each of the other kata repos saying something like:

> See https://github.com/kata-containers/.../blob/master/CONTRIBUTING.md.

That way, there is only 1 file to update and we don't end up with some repos out of sync with others if we ever need to update the file.

But what do you think @sameo, @gnawux, @grahamwhaley, @laijs ?

jodh-intel:
Hi @vbmade2000 - responding to your slack question here, yes, I think we need to update to the latest `libcontainer`. If we look at the top few commits of https://github.com/opencontainers/runc/ commits, we see:

```
$ git log -5 --oneline
96086e5 (HEAD -> master, origin/master, origin/HEAD) Merge pull request #1683 from vikaschoudhary16/runc-systemd-race
d5b4a3e Fix race against systemd
e6516b3 Merge pull request #1678 from sboeuf/sboeuf/subreaper
7f24b40 Merge pull request #1675 from tklauser/apparmor-no-cgo
[...]

PR #93, opened 2018-01-12
vendor: update libcontainer vendoring version

vbmade2000:
Bump the vendoring version of libcontainer in order
to get the fix opencontainers/runc#1678 included in
our code. This will prevent libcontainer from waiting
SIGKILL'ed processes.

Fixes: #70


jodh-intel:
Thanks very much for your contribution! There is just one last thing to do...

For re-vendor PRs like this, we list the range of commits that we're adding (in this case to `libcontainer`). That information is useful to see when looking back through the git history to help diagnose behaviour changes, etc.

For this PR, you'd do something like this:

```
$ cd $GOPATH/src/github.com/opencontainers/runc
$ old_commit=2e7cfe036e2c6dc51ccca6eb7fa3ee6b63976dcd
$ new_commit=e6516b3d5dc780cb57a976013c242a9a93052543
$ git log --no-merges --abbrev-commit --pretty=oneline "${old_commit}..${new_commit}")|sed 's/^/    /g'
```

You'd then paste the git log output into the commit body itself. The `sed` adds an indent which is useful as github then renders the commit list in a fixed-width font which looks great and makes the list easier to read. Here's an example from the Clear Containers runtime where we recently updated the virtcontainers vendoring:

- https://github.com/clearcontainers/runtime/pull/901

I've just realised that this isn't actually documented (:smile:), so I'm about to add these details to https://github.com/kata-containers/community/pull/9...

Thanks again.

jodh-intel:
I've now updated the contributing guide PR to outline this process:

- https://github.com/kata-containers/community/pull/9

vbmade2000:
@jodh-intel  Sure. Done.

jodh-intel:
Thanks! I've just noticed that there are 3 commits here. I think you only need a single one and I don't think you should need a merge commit? You should be able to:

- clone https://github.com/kata-containers/agent/
- `git checkout -b ISSUE-70`
- run `go dep`
- `git commit`
- `git push -f` to your https://github.com/vbmade2000/agent branch.

PR #99, opened 2018-01-16
mount: Fixes for both 9pfs and virtio-blk/scsi

sameo:
* Set `absSource` properly.
* Create 9pfs target directories in the guest.

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -64,7 +64,9 @@ func mount(source, destination, fsType string, flags int, options string) error 
 	var absSource string
 
 	if fsType != type9pFs {
-		absSource, err := filepath.EvalSymlinks(source)
+		var err error
+
+		absSource, err = filepath.EvalSymlinks(source)
 		if err != nil {
 			return fmt.Errorf("Could not resolve symlink for source %v", source)
 		}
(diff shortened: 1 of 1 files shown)


jodh-intel:
This doesn't compile for me?

sameo:
@jodh-intel Sorry I was rushing things out...Fixed now.

mount.go
```diff
@@ -57,14 +57,25 @@ var flagList = map[string]int{
 	"runbindable": unix.MS_UNBINDABLE | unix.MS_REC,
 }
 
+func createDestinationDir(dest string) error {
+	targetPath, _ := filepath.Split(dest)
+	if err := os.MkdirAll(targetPath, mountPerm); err != nil {
```
jodh-intel: You could simplify this to:

```
return os.MkdirAll(targetPath, mountPerm)
```

bergwolf:
LGTM

bergwolf:
travis service is broken. Merge regardlessly.

laijs:
It uses https://golang.org/pkg/path/filepath/#Split

if the destination is '/path/to/9pfs/dir', only the '/path/to/9pfs/' is created, not '/path/to/9pfs/dir'.
Is it deliberated?

@sameo

sameo:
@laijs No, it's not deliberated. It works for us because we're always passing a `/` terminated path to our gRPC `Storage` messages. But we should make that code more robust by calling `Split` only when the mount point is not a directory.

laijs:
for 9pfs case, I think it's better always to consider the destination is dir.

Issue #96, opened 2018-01-12
grpc: Return nil from SignalProcess() if container already stopped


sboeuf:
There is no need to try to signal a container already stopped. Only need to return nil in that case.

PR #100, opened 2018-01-16
grpc: Return nil from SignalProcess() if container already stopped

vbmade2000:
There is no need to try to signal a container already stopped.
Only need to return nil in that case.

Fixes: #96

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -550,6 +550,7 @@ func (a *agentGRPC) SignalProcess(ctx context.Context, req *pb.SignalProcessRequ
 			"sandbox":     a.sandbox.id,
 			"signal":      signal.String(),
 		}).Info("discarding signal as container stopped")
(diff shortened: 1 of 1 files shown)


grpc.go
```diff
@@ -550,6 +550,7 @@ func (a *agentGRPC) SignalProcess(ctx context.Context, req *pb.SignalProcessRequ
 			"sandbox":     a.sandbox.id,
 			"signal":      signal.String(),
 		}).Info("discarding signal as container stopped")
+		return nil, nil
```
jodh-intel: I think the first value returned here should be `emptyResp` for consistency.

vbmade2000: Correct.

sboeuf: Yes @vbmade2000 please do this change.

vbmade2000: Done

Issue #95, opened 2018-01-12
agent: Add missing nsenter import


sboeuf:
It is mandatory that we import `nsenter` package, otherwise the `init()` function won't be able to spawn a new container.
This line has to be added to `agent.go`:
```go
import (
        ...
        _ "github.com/opencontainers/runc/libcontainer/nsenter"
        ...
)
```

This requirement comes from the libcontainer documentation here: https://github.com/opencontainers/runc/blob/master/libcontainer/README.md
It explains that we need to add this import so that it will handle part of the container bootstrapping.
More details here: https://github.com/opencontainers/runc/blob/master/libcontainer/nsenter/README.md

vbmade2000:
Can I work on this simple issue ?

sameo:
@vbmade2000 Yes, please do.

vbmade2000:
@sboeuf  There does not exist github.com/opencontainers/runc/libcontainer/nsenter path. in vendors  It exists in runc repo in release v1.0.0-rc4.

jodh-intel:
@sboeuf - when you are back tomorrow, could you also add a bit more detail to this issue explaining why the container spawn will fail?

sboeuf:
@jodh-intel I have updated the issue description.

jodh-intel:
Thanks @sboeuf.

sboeuf:
@vbmade2000 I don't get what is the problem here ? The vendoring of `github.com/opencontainers/runc` refers to the version 1.0.0-rc4 as you mentioned. And the `nsenter` package does exist for this version.
Now if what you mean is that nsenter is not into the `/vendor/...` directory, this will be solved by first modifying the code so that you actually import `_ "github.com/opencontainers/runc/libcontainer/nsenter"`, and then by running the following command:
```bash
dep ensure && dep ensure -update github.com/opencontainers/runc && dep prune
```

sameo:
@vbmade2000 You need to add the import into the code and then `dep` will detect this new dependency and add `nsenter` to the vendoring path.

As @sboeuf said, you need to run:

```
dep ensure -v && dep ensure -update -v github.com/opencontainers/runc && dep prune -v
```

sboeuf:
@vbmade2000 are you still planning on implementing this ?
I am sorry to rush you here but we're trying to enable our runtime with the Kata agent, and we would need this to be merged asap.

vbmade2000:
@sboeuf I am done with it. Raising PR now.

[...]

PR #101, opened 2018-01-17
agent: Add missing nsenter import

vbmade2000:
We need to import `nsenter` package to make `init` function to be able to spawn a
new container.

Fixes: #95 

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -21,6 +21,7 @@ import (
 	pb "github.com/kata-containers/agent/protocols/grpc"
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	_ "github.com/opencontainers/runc/libcontainer/nsenter"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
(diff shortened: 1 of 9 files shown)


sameo:
LGTM

agent.go
```diff
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+    _ "github.com/opencontainers/runc/libcontainer/nsenter"
```
sboeuf: This is not indented the right way, and it's not in the right place in the code I think.
Could run some linters to fix that please:
```bash
go_packages=$(go list ./... | grep -v protocols | grep -v vendor)
go_files=`go list -f  '{{.Dir}}/*.go' $go_packages`
go fmt $go_packages
golint $go_packages
go vet $go_packages
gofmt -s -w $go_files
gocyclo -over 15 $go_files
ineffassign .
```

vbmade2000: Sure. Quickly.

vbmade2000:
@sboeuf If you still see some issue and in rush then I can send you patch.

sboeuf:
@vbmade2000 thx for the quick fix !

sameo:
@vbmade2000 Thanks for your contribution, merging now.

PR #103, opened 2018-01-17
Fix agent to make it work

### LINKED ISSUE #102: Fix the agent to make it functional
We need to fix a few things on the agent code so that we can run our first container using it. This issue covers several tiny fixes. More details on the commits of the PR covering the implementation.

sboeuf:
This PR submits several fixes in order to get the agent properly working.

Fixes #102


sboeuf:
cc @sameo @bergwolf @jodh-intel @laijs

grpc.go
```diff
@@ -364,8 +364,13 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		mounts:      mountList,
 	}
 
-	err = container.container.Start(&builtProcess.process)
-	if err != nil {
+	// This lock has to be taken to make sure we let the libcontainer
+	// spawning process being reaped internally by libcontainer. This
+	// prevents our subreaper from doing it.
```
bergwolf: IIUC the libcontainer spawning process is put on hold until libcontainer.Exec() is called. Which process does the subreaper lock protect here?

grpc.go
```diff
@@ -201,6 +201,7 @@ func (a *agentGRPC) execProcess(ctr *container, proc *process) (err error) {
 	defer a.sandbox.subreaper.RUnlock()
 
 	if proc == nil {
+		proc = ctr.initProcess
```
bergwolf: ah, good catch, thanks!

grpc.go
```diff
@@ -393,8 +393,10 @@ func (a *agentGRPC) StartContainer(ctx context.Context, req *pb.StartContainerRe
 		return nil, err
 	}
 
-	if status != libcontainer.Created {
-		return nil, fmt.Errorf("Container %s status %s, should be %s", req.ContainerId, status, libcontainer.Created)
+	if status != libcontainer.Created && status != libcontainer.Stopped {
```
bergwolf: So we allow `Stopped` -> `Running` state transition for agent containers. I'm not sure if this can work because the libcontainer container process does not exist and thus libcontainer.Exec() does nothing.

grpc.go
```diff
@@ -324,6 +324,8 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	ociSpec.Linux.IntelRdt = nil
```
bergwolf: Shouldn't this be controlled by runtime instead?

grpc.go
```diff
@@ -439,6 +342,18 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
+			config.Mounts[idx] = &configs.Mount{
+				Device:      "tmpfs",
+				Source:      "shm",
+				Destination: "/dev/shm",
+				Data:        "mode=1777,size=65536k",
+				Flags:       defaultMountFlags,
+			}
+		}
+	}
+
```
laijs: > The specific case of /dev/shm requires the agent to handle this mount
in a special way that cannot be properly handled by the runtime.

I don't get it. Why the runtime can't pass the config as the above code?

grpc.go
sameo: @sboeuf In which cases do we need to start a container that's in the `Stopped` state?

grpc.go
laijs: the target process may be failed and exit or be killed. the subreaper lock is needed in this case.

grpc.go
sameo: @sboeuf I agree with @laijs: This mount should be curated from the runtime.

grpc.go
sameo: @sboeuf With https://github.com/containers/virtcontainers/pull/568, this should not be needed.

grpc.go
sboeuf: Removed! I am gonna try to make this work another way.

grpc.go
sboeuf: @sameo Yes I know, but it doesn't, I have to investigate. I have removed the commit in the meantime.

grpc.go
sboeuf: My bad, with all those moving pieces, I have just realized that I don't need this anymore to make the code working.

sboeuf:
I have updated the PR, and it has been tested without the need for SIGUSR1 to be send to the shim. I have made some internal changed to virtcontainers to get this working.
Now, about this PR, I think we have something pretty clean and fitting the expected behavior of runc/libcontainer.

sboeuf:
I have added a patch to fix the spec converter. This actually fixes the issue related to `IntelRdt` being not nil as expected.

grpc.go
```diff
@@ -188,9 +188,17 @@ func (a *agentGRPC) getContainer(cid string) (*container, error) {
 	return ctr, nil
 }
 
-// Shared function between StartContainer and ExecProcess, because those expect
-// a process to be run.
-func (a *agentGRPC) execProcess(ctr *container, proc *process) (err error) {
+// Shared function between CreateContainer and ExecProcess, because those expect
+// a process to be run, and the console to be properly setup.
```
sameo: I see where you want to go with that one, but it feels like the `execProcess` function is doing exec and non exec operations now.
What you're trying to factorize here is the process post start phase from `execProcess()`. IMHO the following would make the code more readable:

1. Factorize the process post start code:

```Go
func (a *agentGRPC) postStartProcess(ctr *container, proc *process) (err error) {
	// Get process PID
	pid, err := proc.process.Pid()
	if err != nil {
		return err
	}

	proc.exitCodeCh = make(chan int, 1)

	// Create process channel to allow WaitProcess to wait on it.
	// This channel is buffered so that reaper.reap() will not
	// block until WaitProcess listen onto this channel.
	a.sandbox.subreaper.setExitCodeCh(pid, proc.exitCodeCh)

	// Setup terminal if enabled.
	if proc.consoleSock != nil {
		termMaster, err := utils.RecvFd(proc.consoleSock)
		if err != nil {
			return err
		}

		if err := setConsoleCarriageReturn(int(termMaster.Fd())); err != nil {
			return err
		}

		proc.termMaster = termMaster
	}

        ctr.setProcess(proc)

	return nil
}
```

2. Reduce `execProcess()` to
```Go
func (a *agentGRPC) execProcess(ctr *container, proc *process) (err error) {
	// This lock is very important to avoid any race with reaper.reap().
	// Indeed, if we don't lock this here, we could potentially get the
	// SIGCHLD signal before the channel has been created, meaning we will
	// miss the opportunity to get the exit code, leading WaitProcess() to
	// wait forever on the new channel.
	// This lock has to be taken before we run the new process.
	a.sandbox.subreaper.RLock()
	defer a.sandbox.subreaper.RUnlock()

	if proc == nil {
		err = ctr.container.Exec()
	} else {
		err = ctr.container.Run(&(proc.process))
	}

        return err
}
```

3. Call `postStartProcess()` after `execProcess()` in both the `start` and `run` cases.
4. Call `postStartProcess()` after  `container.container.Start()` for the `create` case.

grpc.go
```diff
@@ -748,3 +747,4 @@ func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemReques
 
 	return emptyResp, nil
 }
+
```
sameo: nit: Unneeded change.

protocols/grpc/utils.go
```diff
@@ -22,7 +22,14 @@ func copyValue(to, from reflect.Value) error {
 	}
 
 	if toKind == reflect.Ptr {
-		// If the destination is a pointer, we need to allocate a new one.
+		// Handle the case of nil pointers.
+		if fromKind == reflect.Ptr && from.IsNil() {
+			return nil
+		}
+
```
sameo: Good catch, thanks.

sboeuf: Sure ;)

grpc.go
sboeuf: Will do !

grpc.go
sboeuf: Several things here. I agree this would be a bit cleaner but we cannot apply this as you've described it, otherwise we won't hold the lock when setting the exit code channel, which is pretty important. We cannot release the lock after we've started the process, and until we have set the exit code channel. We need this lock both to prevent the subreaper to reap the first process started by libcontainer and second to prevent the subreaper to reap a terminated process before the exit code channel has been registered.
I could maybe split differently, by retrieving the pid and setting the exit code channel from `execProcess()`, and do the rest from `postStartProcess()`.
Another thing, I don't want to call into `postStartProcess()` for the `start` case because the only need to do is call `Exec()`, nothing else.

sameo: Right, I did not look at those details, but the idea was that `execProcess()` should only be used for starting and exec'ing. So if you can find the right split to do so, it would be cleaner indeed.

sboeuf:
@sameo comments addressed !

laijs:
LGTM

sameo:
LGTM

PR #117, opened 2018-01-26
agent: enable it to be the init process

### LINKED ISSUE #116: Enable agent starting as the init process
Just as the same ability as hyperstart, so that kata-agent can work as the agent of hyperd/runv.

laijs:
Just as the same ability as hyperstart, so that kata-agent
can work as the agent of hyperd/runv.

fixes #116

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -54,6 +54,7 @@ type sandbox struct {
 
 	id           string
 	running      bool
+	noPivotRoot  bool
 	containers   map[string]*container
 	channel      channel
 	network      network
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -423,6 +480,16 @@ func main() {
 		},
 	}
 
+	if len(os.Args) == 1 && filepath.Base(os.Args[0]) == "init" && os.Getpid() == 1 {
```
sboeuf: I would prefer if we don't force a specific name here. I think `os.Getpid() == 1` should be enough to know if this agent has been run as the init process.

agent.go
```diff
@@ -380,6 +382,61 @@ func (s *sandbox) teardown() error {
...
 
+type initMount struct {
+	typ, src, des string
+	options       []string
+}
+
+func initAgentAsInit() error {
+	agentLog.Infof("initAgentAsInit(), agent version: %s\n", version)
+
+	mounts := []initMount{
```
sboeuf: I'd be better to define this outside of this function. This should be const structure list.

agent.go
```diff
@@ -380,6 +382,61 @@ func (s *sandbox) teardown() error {
 	return s.channel.teardown()
 }
 
+type initMount struct {
+	typ, src, des string
+	options       []string
+}
+
+func initAgentAsInit() error {
```
sboeuf: Please add some verbose comments about the point of this function and explain roughly what it does.

laijs:
@sboeuf updated

agent.go
```diff
@@ -380,6 +381,64 @@ func (s *sandbox) teardown() error {
 	return s.channel.teardown()
 }
 
+type initMount struct {
+	typ, src, des string
```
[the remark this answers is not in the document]
laijs: updated

jshachm:
@laijs @WeiZhang555 
Do we need to consider the case that using `nsenter` to get into the `agent namespace` and do harm to guest filesystem like `cgroup` or `9p-share dir` ?

laijs:
@jshachm you can build your own rootfs.

sboeuf:
LGTM

bergwolf:
lgtm

PR #126, opened 2018-02-01
grpc: initial changes for networking protocol

egernst:
WIP changes for making the networking interface gRPC protocol declarative.

### For the proposal in this patch:
The number of messages for route and interface handling is reduced to
two with this patch: UpdateRoutes and UpdateInterfaces.

UpdateInterfaces operates on an array of network-interfaces.  Doing this
allows for removal of interfaces by sending a new update request with
the interface no longer included.  For example, initially
UpdateInterfaces is called for 'A,' 'B' and 'C.'  Later a network remove of
network interface 'B' is carried out by calling UpdateInterfaces with
just 'A' and 'C'.  While this simplifies the protocol and runtime handling, it does put a
burden on the agent to write out the (non-necessarily changed) interface
details for 'A' and 'C' and determine that 'B' was removed.  In the event that network disconnect is used for each of the existing interfaces, a nil array of Interfaces could be sent.  

UpdateRoutes also operates on an array of routes.

This patch doesn't work - just as an example which barely compiles to drive the discussion.

### Alternative approach:

We could allow for explicit removal of interfaces by either:
  1. defining a RemoveInterface message, or 
  2. adding a flag (or type: update|remove) that could be passed to the UpdateInterface function

If these options are used, the agent would work on the (network) interface granularity rather than on all of the interfaces per message call.

### Further details:
* Regardless of which approach we use, the interface or interfaces and routes are returned as well as error status (as discussed already)
* Interfaces will be identified by MAC-address (as discussed already)

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -716,44 +716,28 @@ func (a *agentGRPC) DestroySandbox(ctx context.Context, req *pb.DestroySandboxRe
 	return emptyResp, nil
 }
 
-func (a *agentGRPC) AddInterface(ctx context.Context, req *pb.AddInterfaceRequest) (*gpb.Empty, error) {
-	if err := a.sandbox.addInterface(nil, req.Interface); err != nil {
-		return emptyResp, err
-	}
+func (a *agentGRPC) AddInterface(ctx context.Context, req *pb.AddInterfaceRequest) (*pb.Interface, error) {
+	return a.sandbox.addInterface(nil, req.Interface)
+}
 
+func (a *agentGRPC) UpdateInterface(ctx context.Context, req *pb.UpdateInterfaceRequest) (*pb.Interface, error) {
+	return a.sandbox.updateInterface(nil, req.Interface)
+}
+
+func (a *agentGRPC) RemoveInterface(ctx context.Context, req *pb.RemoveInterfaceRequest) (*pb.Interface, error) {
+	return a.sandbox.removeInterface(nil, req.Interface)
+}
+
+func (a *agentGRPC) AddRoute(ctx context.Context, req *pb.AddRouteRequest) (*gpb.Empty, error) {
+	return emptyResp, a.sandbox.addRoute(nil, req.Route)
+}
+
+func (a *agentGRPC) UpdateRoute(ctx context.Context, req *pb.UpdateRouteRequest) (*gpb.Empty, error) {
 	return emptyResp, nil
 }
 
-func (a *agentGRPC) RemoveInterface(ctx context.Context, req *pb.RemoveInterfaceRequest) (*gpb.Empty, error) {
-	if err := a.sandbox.removeInterface(nil, req.Name); err != nil {
-		return emptyResp, err
-	}
-
-	return emptyResp, nil
-}
-
-func (a *agentGRPC) UpdateInterface(ctx context.Context, req *pb.UpdateInterfaceRequest) (*gpb.Empty, error) {
-	if err := a.sandbox.updateInterface(nil, req.Interface, req.Type); err != nil {
-		return emptyResp, err
-	}
-
-	return emptyResp, nil
-}
-
-func (a *agentGRPC) AddRoute(ctx context.Context, req *pb.RouteRequest) (*gpb.Empty, error) {
-	if err := a.sandbox.addRoute(nil, req.Route); err != nil {
-		return emptyResp, err
-	}
-
-	return emptyResp, nil
-}
-
-func (a *agentGRPC) RemoveRoute(ctx context.Context, req *pb.RouteRequest) (*gpb.Empty, error) {
-	if err := a.sandbox.removeRoute(nil, req.Route); err != nil {
-		return emptyResp, err
-	}
-
-	return emptyResp, nil
+func (a *agentGRPC) RemoveRoute(ctx context.Context, req *pb.RemoveRouteRequest) (*gpb.Empty, error) {
+	return emptyResp, a.sandbox.removeRoute(nil, req.Route)
 }
 
 func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemRequest) (*gpb.Empty, error) {
--- a/network.go
+++ b/network.go
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2018 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -53,73 +53,66 @@ func linkByHwAddr(netHandle *netlink.Handle, hwAddr string) (netlink.Link, error
 	return nil, fmt.Errorf("Could not find the link corresponding to HwAddr %q", hwAddr)
 }
 
-func updateInterfaceAddrs(netHandle *netlink.Handle, link netlink.Link, addrs []*pb.IPAddress, add bool) error {
-	for _, addr := range addrs {
-		netlinkAddrStr := fmt.Sprintf("%s/%s", addr.Address, addr.Mask)
+func updateLink(netHandle *netlink.Handle, link netlink.Link, iface *pb.Interface) error {
 
+	// As a first step, clear out any existing addresses associated with the link:
+	linkIPs, err := netlink.AddrList(link, netlink.FAMILY_V4)
+	if err != nil {
+		return fmt.Errorf("Could not check initial addresses for the link: %v", err)
+	}
+	for _, linkIP := range linkIPs {
+		if err := netlink.AddrDel(link, &linkIP); err != nil {
+			return fmt.Errorf("Could not delete existing addresses: %v", err)
+		}
+	}
+
+	// Set desired IP addresses:
+	for _, addr := range iface.IPAddresses {
+		netlinkAddrStr := fmt.Sprintf("%s/%s", addr.Address, addr.Mask)
 		netlinkAddr, err := netlink.ParseAddr(netlinkAddrStr)
+
 		if err != nil {
 			return fmt.Errorf("Could not parse %q: %v", netlinkAddrStr, err)
 		}
 
-		if add {
-			if err := netHandle.AddrAdd(link, netlinkAddr); err != nil {
-				return fmt.Errorf("Could not add %s to interface %v: %v",
-					netlinkAddrStr, link, err)
-			}
-		} else {
-			if err := netHandle.AddrDel(link, netlinkAddr); err != nil {
-				return fmt.Errorf("Could not remove %s from interface %v: %v",
-					netlinkAddrStr, link, err)
-			}
+		if err := netHandle.AddrAdd(link, netlinkAddr); err != nil {
+			return fmt.Errorf("Could not add %s to interface %v: %v",
+				netlinkAddrStr, link, err)
 		}
+	}
+
+	// set the interface name:
+	if err := netHandle.LinkSetName(link, iface.Name); err != nil {
+		return fmt.Errorf("Could not set name %s for interface %v: %v", iface.Name, link, err)
+	}
+
+	// set the interface MTU:
+	if err := netHandle.LinkSetMTU(link, int(iface.Mtu)); err != nil {
+		return fmt.Errorf("Could not set MTU %d for interface %v: %v", iface.Mtu, link, err)
 	}
 
 	return nil
 }
 
-func updateInterfaceName(netHandle *netlink.Handle, link netlink.Link, name string) error {
-	return netHandle.LinkSetName(link, name)
-}
-
-func updateInterfaceMTU(netHandle *netlink.Handle, link netlink.Link, mtu int) error {
-	return netHandle.LinkSetMTU(link, mtu)
-}
-
-func updateLink(netHandle *netlink.Handle, link netlink.Link, iface *pb.Interface, actionType pb.UpdateType) error {
-	switch actionType {
-	case pb.UpdateType_AddIP:
-		return updateInterfaceAddrs(netHandle, link, iface.IpAddresses, true)
-	case pb.UpdateType_RemoveIP:
-		return updateInterfaceAddrs(netHandle, link, iface.IpAddresses, false)
-	case pb.UpdateType_Name:
-		return updateInterfaceName(netHandle, link, iface.Name)
-	case pb.UpdateType_MTU:
-		return updateInterfaceMTU(netHandle, link, int(iface.Mtu))
-	default:
-		return fmt.Errorf("Unknown UpdateType %v", actionType)
-	}
-}
-
-func (s *sandbox) addInterface(netHandle *netlink.Handle, iface *pb.Interface) (err error) {
+func (s *sandbox) addInterface(netHandle *netlink.Handle, iface *pb.Interface) (resultingIfc *pb.Interface, err error) {
 	s.network.ifacesLock.Lock()
 	defer s.network.ifacesLock.Unlock()
 
 	if netHandle == nil {
 		netHandle, err = netlink.NewHandle()
 		if err != nil {
-			return err
+			return nil, err
 		}
 		defer netHandle.Delete()
 	}
 
 	if iface == nil {
-		return fmt.Errorf("Provided interface is nil")
+		return nil, fmt.Errorf("Provided interface is nil")
 	}
 
 	hwAddr, err := net.ParseMAC(iface.HwAddr)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	link := &netlink.Device{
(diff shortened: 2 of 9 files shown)


egernst:
@sboeuf @laijs @WeiZhang555  -- please let me know your feedback on above -- I want to make sure we're in a position to finish implementation based on whichever approach tomorrow so we can get networking up and working.  Thanks!

bergwolf:
Sorry but I have to argue one last time about using MAC address as interface identifier. MAC addresses can be changed by users. What is agent supposed to do if a user changes MAC address from inside of the guest?

jodh-intel:
# gRPC Traffic

I'm also concerned about the amount of potential gRPC traffic this design would generate. Have we considered the pathological cases?

- Maximum number of network interfaces (255?)
- Maximum number of routes (lots!)

Worst case, that sounds like a lot of bytes flying around just to remove a single if/route if the number of those is already large. Or put another way,

- *Every new container created results in more "setup traffic" than was required for the last container.*

Clearly, the ideal would be that a new container network request only requires us to send what is needed to create that new container network.

# Atomicity impacts

Also, if as @WeiZhang555 says these operations needs to be atomic (apply all or apply none), couldn't this theoretically have a negative impact on the system. For example,

- many routes exist and one needs to be removed.
- virtcontainers (for Clear Containers) sends the list of all routes to the agent, minus the one to be removed.
- the agent goes through the list applying the routes.
- the agent detects the last route has been deleted from the list and tries to delete that route.
- the route removal fails for some reason.
- the agent then has to undo all the changes.

## Design complexity

Note in that scenario that **both** the agent *and* virtcontainers need to keep track of the list of routes:

- virtcontainers needs to know the current and future routes to know *not* to send the route it wants the agent to delete.
- the agent needs to know the current and future routes so it can calculate "what changed" and therefore know to delete the route virtcontainers didn't send this time.

 If we want a simple design, shouldn't it just be a single entity that has to deal with this sort of logic?

egernst:
@jodh-intel - 
-For chattiness, it sounds like the alternative proposal, that is to add an ability to have a separate remove and then be able to work on an interface at a time, would address the issues you've brought up regarding chattiness. 
-As for atomicity, I think the issue would be reduced, and more manageable.  The idea was that we would return the interface/route that was created, not necessarily roll everything back. 

What do you think?

egernst:
@bergwolf - there was a lot of discussion regarding mac address on the last PR.  To me the prior conclusion still makes a lot of sense  to me.

sboeuf:
@egernst thanks for summarizing everything here. Based on @jodh-intel input about the traffic, my preference would be option 2:
```go
UpdateInterface(iface Interface, action ActionType)
UpdateInterface(route Route, action ActionType)
```
where `ActionType` would be
```go
const (
        UpdateType ActionType = iota
        RemoveType
        AddType
)
```
@bergwolf notice that this solution would allow you to rely on the structure `Interface` in any case, which could be used to pass the name in case you consider the MAC address non reliable.

jodh-intel:
@sboeuf - sounds good to me.

sameo:
@egernst So I think we should keep the `AddInterface()` and `RemoveInterface()` calls, it would be a little tricky to implement add and remove from the agent with only `UpdateInterfaces()`. Also `UpdateInterfaces()` should return the result of the update, i.e. the state of the set of interfaces after the update. It's up to the caller to decide if it wants to roll back, insist, etc...in case of failure.

As for the routes, I think we should do the same. Maybe we will eventually realize that the `AddRoute()` and `RemoveRoute()` calls are not needed and we will then remove them. This API is not even alpha, so we can have that freedom for now.

egernst:
@sameo -- thanks for the input.  

I'm now working on making an updated PR pulling in the various input.  

### Updated proposal:

For ease of use it makes sense to keep messages for the three general actions: add, update, remove.  I don't see any rationale for not allowing the addition, removal and/or update of multiple interfaces at a single time.   Based on this, I'll update the PR to have:

```go
AddInterfaces(ifaces Interfaces) Interfaces
UpdateInterfaces(ifaces Interfaces) Interfaces
RemoveInterfaces(ifaces Interfaces) Interfaces
```

 Same logic will be applied for routes as well.   Should have updated PR in a bit...

WeiZhang555:
The last one looks better

AddInterfaces(ifaces Interfaces) Interfaces
UpdateInterfaces(ifaces Interfaces) Interfaces
RemoveInterfaces(ifaces Interfaces) Interfaces

Except that the RemoveInterfaces may not need to return an Interface as it doesn’t have a middle state(does it?), and we can rename *Interfaces to *Interface as we operate on single interface now

egernst:
@WeiZhang555  -- check out my latest pull.  I'd prefer to leave the option to operate still on multiple interfaces at a time.  

The return value for each of them will still be interfaces.  

In the case of errors during removal, we'll return the interfaces which remain still along the list of items requested for removal.  If all goes well, then it'll return nil, nil.

jodh-intel:
Nice. +1 to the updated proposal: more flexibility and granularity.

grpc.go
```diff
@@ -718,44 +718,100 @@ func (a *agentGRPC) DestroySandbox(ctx context.Context, req *pb.DestroySandboxRe
...
+	for _, ifc := range req.Interfaces {
+		if ifc == nil {
+			continue
+		}
+
+		err := a.sandbox.removeInterface(nil, ifc)
+		if err != nil {
+			return &resultingInterfaces, err
+		}
+		resultingInterfaces.Interfaces = append(resultingInterfaces.Interfaces[:0], resultingInterfaces.Interfaces[1:]...)
```
WeiZhang555: This logic is wrong when the req.Interfaces = {A, nil, B} because `if ifc == nil {continue}`

This is removing always the first interface from list and it's not correct

And it's actually equal to 
`resultingInterfaces.Interfaces = resultingInterfaces.Interfaces[1:]`

grpc.go
```diff
@@ -718,44 +718,100 @@ func (a *agentGRPC) DestroySandbox(ctx context.Context, req *pb.DestroySandboxRe
...
+
+	for _, ifc := range req.Interfaces {
+		if ifc == nil {
+			continue
+		}
+
+		err := a.sandbox.updateInterface(nil, ifc)
+		if err != nil {
+			return &resultingInterfaces, err
+		}
```
WeiZhang555: This is  ambiguous, consider this situation

UpdateInterfaces(A, B, C)

For B, you want to update "mac" and "mtu" and "add ip" for B but it failed, so in this implementation, 
`err := a.sandbox.updateInterface(nil, ifc)` return non-nil error, and `UpdateInterfaces` returns "{A}" and `err` as result.

But Interface B in this VM has gone into an unknown state, right? How do we know exactly the right state of all interfaces so?

WeiZhang555:
After reading the implementation, I feel more -1 on this proposal now, check my comment: https://github.com/kata-containers/agent/pull/126#discussion_r165660701

I prefer the original design:

e.g. `UpdateInterface` handles one interface update at a time, if it fails, it returns final state of this interface, so that we still know the correct setting of it from grpc returned value. When this `UpdateInterface` extends to multiple interfaces situation, it will be so complicated and makes the API ambiguous, I don't think this is the declarative ways.

I hope I made myself clear. For the reason I said about, I'm -1 on this design.

egernst:
@WeiZhang555 - I definitely understand your point(s) -- the impl isn't close to done on it as is right now.

A couple of initial thoughts after reading your feedback:
*  Being able to create / update /remove a set of interfaces is a (good) option, though not necessary - you could send requests with just a single interface or route.
 * As I was initially writing this up, I had just stopped setup once we hit an error, and the return value just shows what has been successfully setup.
 * Perhaps it make more sense to return an error still but to continue to process the remaining interfaces on the list (which is simple in the event it is called with a single interface).
 * As you point out, it may make sense to also return the "resulting" state for the interface.  Let me think on this - it needs to be resolved regardless of whether we are operating on a single or multiple interfaces.

Thanks for the feedback.

egernst:
@WeiZhang555  I made the changes to return the resulting state for an interface in the event that an error is encountered during the update.  

I wanted to still operate on a slice of interfaces/routes, but was having a hard time decided how to best handle the error message which is returned.  For example, if I set ten interfaces but 2 of them fail, it'd be more useful to return the specific error for each rather than a generic "something went wrong somewhere" message and leave it up to the caller to read through.

While error cases here are not expected, as we get started I think working an interface and a route at a time is more straight forward, and we can work on a set of interfaces as an optimization "later".
WDYT?

Note, I still need to add tests - I expect more changes will be needed, but this should give you an idea for the suggested protocol.

WeiZhang555:
@egernst Yes, that would be better IMO.

Handling multiple error messages is another example of why operating on several interfaces/routes is not good idea. But we can start from single one, and extend the API scope if we finally find a good design for slice of inf/routes later. I'm open to discuss more about this later, maybe better in separate issue. 

Currently `working an interface and a route at a time is more straight forward` as you said.

sameo:
@egernst Yes, a single interface/route API look more sensible.

sameo:
Once the tests are fixed: LGTM

egernst:
@sameo @WeiZhang555 @sboeuf  @bergwolf 

Updated to include test covering updating and removal of a network interface, as well as add/remove route.  Please take a look.

I think that this could be adjusted / optimized later (ie, move to using UpdateRoutes instead of add/remove), but for now it'll allow for functioning networking and is ready to be considered for merge imo.  

Verified this also with kata-runtime-cc; see https://github.com/containers/virtcontainers/pull/595

sameo:
LGTM

network.go
```diff
@@ -235,16 +209,50 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *pb.Interface
...
+
+// getInterface will retrieve interface details from the provided link
+func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
+	var ifc pb.Interface
+	linkAttrs := link.Attrs()
+	ifc.Name = linkAttrs.Name
+	ifc.Mtu = uint64(linkAttrs.MTU)
+	ifc.HwAddr = linkAttrs.HardwareAddr.String()
+
+	addrs, _ := netHandle.AddrList(link, netlink.FAMILY_V4)
```
jodh-intel: Why are we not checking the returned error value? I'm guessing because this function doesn't return an error (because this function is called in an error scenario). But I'd still check the error and log it atleast.

network.go
```diff
@@ -235,16 +209,50 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *pb.Interface
...
+error_case:
+	// in the event that an error occurred during the interface update, make sure we return
+	// the resulting state instead of the requested state
+	if err != nil {
+		resultingIfc = getInterface(netHandle, link)
+	} else {
+		resultingIfc = iface
+	}
+
+	return resultingIfc, netHandle.LinkSetUp(link)
```
jodh-intel: Doesn't this mean we're calling `LinkSetUp()` twice for the non-error scenario?

egernst: That is a bug!  Will fix

network.go
egernst: Sure - will do.

egernst:
Adding some more tests to exercise some error cases -- WIP...

egernst:
@jodh-intel - I believe I addressed your initial comments.  Let me know if you have more feedback.

network_test.go
```diff
@@ -0,0 +1,111 @@
+//
+// Copyright (c) 2017 Intel Corporation
```
jodh-intel: Nit: 2018.

network.go
```diff
@@ -236,15 +219,56 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *pb.Interface
...
+	addrs, err := netHandle.AddrList(link, netlink.FAMILY_V4)
+	if err != nil {
+		agentLog.WithError(err).Error("getInterface() failed")
+	}
+	for _, addr := range addrs {
+		s := strings.Split(addr.String(), "/")
+
+		m := pb.IpAddress{
+			Address: s[0],
+			Mask:    s[1],
```
jodh-intel: This looks wrong as `addr.String()` returns a value of the form `ip/mask name` meaning `s[1]` will be `mask name`, not `mask` I think.

network_test.go
egernst: Ah, yes, happy new year to you as well.  Will update.

network.go
egernst: Yes, updating now...

egernst: @jodh-intel -- I updated this as well as the test to check the entire struct in the failure test (the test now would've caught this oversight).  Good catch - looks good now.

sboeuf:
@laijs @bergwolf @WeiZhang555 could you please review this :)

jodh-intel:
Thanks @egernst.

lgtm

network.go
```diff
@@ -53,73 +53,66 @@ func linkByHwAddr(netHandle *netlink.Handle, hwAddr string) (netlink.Link, error
 	return nil, fmt.Errorf("Could not find the link corresponding to HwAddr %q", hwAddr)
 }
 
-func updateInterfaceAddrs(netHandle *netlink.Handle, link netlink.Link, addrs []*pb.IPAddress, add bool) error {
-	for _, addr := range addrs {
-		netlinkAddrStr := fmt.Sprintf("%s/%s", addr.Address, addr.Mask)
+func updateLink(netHandle *netlink.Handle, link netlink.Link, iface *pb.Interface) error {
+
+	// As a first step, clear out any existing addresses associated with the link:
```
liwei: With clear and re-add we will lose any routes related to those addresses(e.g. default route via a gateway address in the same subnet). IMHO, we should compute diff of addresses and apply it instead.

liwei: With clear and re-add we will lose any routes related to those addresses(e.g. default route via a gateway address in the same subnet). IMHO, we should compute diff of addresses and apply it instead.

network_test.go
```diff
@@ -0,0 +1,115 @@
...
+	ifc.Mtu = 999999999999
+	resultingIfc, err = s.updateInterface(netHandle, &ifc)
+	// expecting this failed
+	assert.NotNil(t, err, "Expected failure")
+
+	ifc.Mtu = 500
+	assert.True(t, reflect.DeepEqual(resultingIfc, &ifc),
+		"Resulting inteface should have been unchanged: got %+v, expecting %+v", resultingIfc, ifc)
+	//	assert.False(t, resultingIfc.Mtu != 500,
+	//		"Resulting interface should have been unchanged: got MTU:  %d, expecting MTU: %d", resultingIfc.Mtu, 500)
```
WeiZhang555: Small nit: need to clean unused annotations.

This can be updated in following PRs

protocols/grpc/agent.proto
```diff
@@ -145,7 +145,7 @@ enum IPFamily {
 	v6 = 1;
 }
 
-message IPAddress {
+message IpAddress {
```
WeiZhang555: Why?

I think `IP` should be more common name and follows golint rules.

network.go
WeiZhang555: Yes, this could happen

WeiZhang555:
Sorry for late response.
Some small nits, others LGTM

protocols/grpc/agent.proto
egernst: Sure, will do.

network_test.go
egernst: Ah, should've done commit -a -s :).  Good catch

network.go
egernst: Good catch on the default route -- let me test this and update accordingly.

egernst: @liwei ; @WeiZhang555 I did some testing locally, and I'm not sure if this should be a concern here.  

The default route will get created still for the subset when adding the address.  I verified this in network_test.go.  Initially we have a single address, and then in later 'test' on this same interface we add two more.    When doing so, you'll see that the default route created still exists for the initial address, as well as for the newly created addresses.

The non-default routes need to be added once the addresses have been populated.  If you are changing an interfaces IP information after this point, then the final state of routes for that interface should be sent again as well (I don't see this as a normal use case...).  

I don't see how we'd lose routes.

egernst:
@WeiZhang555 @liwei  - after reviewing the scenario described, I think we shouldn't really hit that scenario.  The interface should be declarative -- we shouldn't have any assumptions about existing routes, etc, but should instead add the routes we want in final state explicitly.    See conversation at https://github.com/containers/virtcontainers/issues/500 as an example.

Ideally I'd rework this PR right now in order to make sure we instead operate on a set of routes and remove all routes as a first step, then applying each route requested.  For now, I think the existing should fit our needs, and we can address "better" route handling in a follow on PR to handle these non-standard scenarios (please correct me if I'm mistaken and this is a normal case).  Getting this in will enable us to have working networking, an important first step.  

WDYT?

liwei:
sounds reasonable and LGTM

sboeuf:
LGTM

sboeuf:
@WeiZhang555 if you're fine with the PR as is, please merge it ;)

laijs:
LGTM

PR #133, opened 2018-02-03
grpc: Handle hotplugged devices

### LINKED ISSUE #132: device: Pass host devices to the container
This feature does not work because the list of devices provided by the OCI spec structure has to be modified first, relying on the major/minor numbers found inside the VM, and after the device has been hotplugged.

sboeuf:
This commit adds a new Device structure in order to avoid confusion                                                          
    between what should be mounted and what is a device. When creating                                                           
    a new container, we expect the caller to pass a list of storages,                                                            
    and a list of devices.
    
When adding a host device to the container by using things like
    "docker run --device /dev/sda1:/dev/blk1", the runtime will not
    be able to anticipate the major/minor numbers for the device
    being hotplugged into the VM. This means the agent has to update
    the list of devices provided through the spec, by relying on the
    list of devices provided.

Fixes #132


sboeuf:
cc @sameo @jodh-intel @bergwolf @WeiZhang555 @laijs

bergwolf:
@sboeuf looking at libcontainer, it seems that it ignores major/minor in the spec when creating containers. Can you explain a bit why the agent needs to fill in device major/minor on its own?

grpc.go
```diff
@@ -322,6 +322,60 @@ func (a *agentGRPC) updateContainerConfig(spec *specs.Spec, config *configs.Conf
...
+	}
+
+	for _, mnt := range mounts {
+		if mnt == nil {
+			continue
+		}
+
+		if mnt.Driver == "" {
+			continue
+		}
```
laijs: In this case, what's the content of mnt.Driver?
I think this code needs to be moved to the corresponding storage driver.
The behavior of this function will be a part of the ABI between runtime <-> agent.
So it needs to be documented in protocol or storage driver.

sboeuf: Well I wanted to raise this PR to discuss if you were okay using this field in order to determine if we actually have a mount or a device. I am assuming an empty driver will mean we have a mount, vs a device if the driver field contains something. Let me know if you're okay with this (@laijs @bergwolf @jodh-intel @sameo), and I will update the doc in the protocol.
Another option would be to define a new field/structure `Device` that we would pass the same way we pass `Storage` for the mounts.

sboeuf:
@bergwolf if we don't provide libcontainer the appropriate major/minor numbers, it will create sort of a fake device because it won't actually be tied to the real device that we expect.
Let's say we have `/dev/sda1` on the host, and we want to mount that inside the container, all we need is to give the name and the major/minor numbers to libcontainer. As a block device, the device will show up as `/dev/vdaX` inside the VM (the name needs to be anticipated by the runtime), and then the agent has to find the numbers since they cannot be anticipated in that case.

bergwolf:
@sboeuf I'm afraid the place to assign major/minor of a device does not ensure it is the same device we'd expect to show up in the guest. In short, `/dev/vdaX` is not a reliable identifier of the device. Therefore the major/minor you get with it may refer to a different device other than `/dev/sda1` on the host.

In hyperstart, we solved the issue with scsi-id of the device. The workflow is that:
1. runtime specify the scsi id of the block device when hotplug it to the guest
2. hyperstart scans `/sys/class/scsi_disk/` directory to find out the device name in the guest
3. check device existence and wait for it if it is not present

The final wait is necessary because Linux kernel device initialization is asynchronous.

sboeuf:
@bergwolf I agree for SCSI device, unfortunately this case is not applicable for block devices.

sboeuf:
@bergwolf @laijs @sameo any comments other comments on this PR ? What should I do here ?

grpc.go
laijs: 1) register drivers implementations to the agent as a list
2) if the driver field contains something, search the driver from the list and call the handling function

sboeuf:
@laijs PR updated adding the list of handlers for different devices.

sboeuf:
ping @WeiZhang555 @bergwolf @laijs @sameo

protocols/grpc/agent.proto
```diff
@@ -194,7 +194,7 @@ message OnlineCPUMemRequest {
 }
 
 message Storage {
-	string driver = 1;           // empty in most cases. it will support "drbd", "bcache" ...
+	string driver = 1;           // Empty driver field means that we expect to do a mount. If driver field is not empty, we expect to do an operation related to a device. The type of device should be given through this field, triggering a specific function from the agent code.
```
laijs: please remove "Empty driver field means that we expect to do a mount. "
the driver field is needed when the agent's general mount code can't handle it,
for example special devices, fstypes...

sboeuf: PR updated !

device.go
```diff
@@ -0,0 +1,133 @@
...
+
+	stat := syscall.Stat_t{}
+	if err := syscall.Stat(device.Source, &stat); err != nil {
+		return err
+	}
+
+	major := int64(stat.Rdev / 256)
+	minor := int64(stat.Rdev % 256)
+
+	agentLog.Infof("Device: %s, Major: %d, Minor: %d", device.Source, major, minor)
```
jodh-intel: Could you use fields for this? Something like:

```
    "device-path": device.Source,
    "device-major": major,
    "device-minor": minor,
}).Info("handling block device")
```

device.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func waitForDevice(devicePath string) error {
+	deviceName := strings.TrimPrefix(devicePath, devPrefix)
+
+	if _, err := os.Stat(devicePath); err == nil {
+		return nil
+	}
+
+	uEvHandler, err := uevent.NewHandler()
```
jodh-intel: One of these handlers is created for every block device passed in the spec. Is that going to be a problem from a performance perspective do you think?

device.go
```diff
@@ -0,0 +1,133 @@
...
+		if d.Path == device.MountPoint {
+			spec.Linux.Devices[idx].Major = major
+			spec.Linux.Devices[idx].Minor = minor
+			updated = true
+			break
+		}
+	}
+
+	if !updated {
+		return fmt.Errorf("Should have found a matching device in the spec")
```
jodh-intel: I guess there is the previous log call, but it might be useful to atleast log `device.Source` here so it's clear which device wasn't found in the spec.

device.go
```diff
@@ -0,0 +1,133 @@
...
+	major := int64(stat.Rdev / 256)
+	minor := int64(stat.Rdev % 256)
+
+	agentLog.Infof("Device: %s, Major: %d, Minor: %d", device.Source, major, minor)
+
+	// Update the spec
+	updated := false
+	for idx, d := range spec.Linux.Devices {
+		if d.Path == device.MountPoint {
+			spec.Linux.Devices[idx].Major = major
```
jodh-intel: This is overwriting the original values right (https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices)? If so, should we log the old ones just in case?

grpc.go
jodh-intel: > I am assuming an empty driver will mean we have a mount, vs a device if the driver field contains something

If we agree on this, please can you add that comment to the definition to make this assumption clear to all:

- https://github.com/kata-containers/agent/blob/master/protocols/grpc/agent.proto#L197

device.go
sboeuf: done

device.go
sboeuf: I haven't added this code in this PR. I only did a copy/paste from `mount.go`.

device.go
sboeuf: done

device.go
sboeuf: done

sboeuf:
@jodh-intel I have updated the comment related to the driver field. Also, I am adding some unit tests for the new functions.

sboeuf:
@jodh-intel PR updated
@bergwolf @laijs PTAL

jodh-intel:
lgtm

device.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/kata-containers/agent/pkg/uevent"
+	pb "github.com/kata-containers/agent/protocols/grpc"
+	"github.com/sirupsen/logrus"
+)
+
+type deviceDriversHandler func(device pb.Storage, spec *pb.Spec) error
+
+var deviceDriversHandlerList = map[string]deviceDriversHandler{
+	"blk": blockDeviceHandler,
```
laijs: how about changing the driver name to "device"?

sboeuf: you mean instead of "blk" ?

sboeuf:
@bergwolf @laijs ping ! Could you please take a look, as I'd like to move forward with this.

device.go
bergwolf: Whatever the name is, it needs to be defined in the protocol instead of here.

device.go
```diff
@@ -0,0 +1,144 @@
...
+	agentLog.WithFields(logrus.Fields{
+		"device-path":  device.Source,
+		"device-major": major,
+		"device-minor": minor,
+	}).Info("handling block device")
+
+	// Update the spec
+	updated := false
+	for idx, d := range spec.Linux.Devices {
+		if d.Path == device.MountPoint {
```
bergwolf: `d.Path` in the OCI spec is the device path. How do you expect it to be a `MountPoint` path here?

It seems that you really need `d.Path == device.Source` instead? But again, this function only works for block devices that are also present in the storage spec (which is also useful since agent can mount the block devices for containers), whereas the commit message and #132 both intend to provide a generic way to pass host devices to the sandbox.

device.go
laijs: I think the code will not be only used for blk devices. so "blk" is not a proper name.
Even "device" doesn't provide enough information. IIUC, it is oci-style passing the devices which can be seen by the runtime/libcontainer into the container. Using "oci-device" or "device-passthrough" for the driver name instead of "blk" could be a good choice either.

Drivers are a part of the agent protocol, but I think just putting some comments on the documents is sufficient.

bergwolf: Yes, comments are fine to me. The point is that the supported values need to be documented there.

The association with `pb.Storage` makes the handler block device specific. I think we can:
1. remove the association with `pb.Storage` in the handler, and rename "blk` to a more generic name
2. make sure that if a block device `dev.Path` matches `pb.Storage.Source`, it will be mounted at `pb.Storage.Mountpoint`

device.go
sboeuf: No no, I need `MountPoint` because this is the expected destination for the device, and we expect this destination to be the source of the device which has been passed-through.

sboeuf:
@bergwolf sure, I can put a clear definition of what is expected into the protocol, this makes sense. But I'd like to understand what are __ALL__ the different values possible for `Driver` ?
@laijs using `device` instead of `blk` might work for now, but what if we pass a `scsi` device for instance, how will you make the difference with a block device using the flag `device` ? I am asking because they won't be processed exactly the same way.

device.go
bergwolf: In OCI spec, `spec.Linux.Devices.Path` is a device path like `/dev/sdb` and libcontainer would call `mknodat` to create it with the provided device numbers. While in kata agent protocol, `pb.Storage.MountPoint` the is mount point of the storage which is a directory outside of the container rootfs. How do you expect them to be equal?

bergwolf:
@sboeuf by **ALL**, I mean all the currently supported values. An empty string and "device" (or any proper name picked in this PR) are the only two currently supported. We can add more when necessary.

device.go
bergwolf: @sboeuf I guess we have some misunderstanding here. Can you please provide an example config of `spec.Linux.Devices` and `pb.Storage` that you intend to handle with this PR?

sboeuf: Ok so let me try to clarify here. In case we want to pass-through a block device such as `/dev/sdb` from the host to the container, we need to hotplug this device into the VM. After it has been hotplugged, it will show up as `/dev/vdX` name that we can anticipate from the runtime. That's why I expect the runtime to provide the right name `/dev/vdX` through the `Source` field, and the name expected by the OCI spec through the `Mountpoint` field, so that we can match with the spec which entry we should modify. Once we have found the device `/dev/vdX`, we get the __real__ major/minor numbers and we update them into the OCI spec for the corresponding device. If we don't update them, libcontainer will create a new device inside the container based on no real device, which means the device will be there but not linked to the real device.

sboeuf:
@laijs please let me know what you think about this comment:
>using device instead of blk might work for now, but what if we pass a scsi device for instance, how will you make the difference with a block device using the flag device ? I am asking because they won't be processed exactly the same way.

sboeuf:
sorry, I hit the wrong button...

laijs:
@sboeuf DriverOptions can be used in this case.

device.go
bergwolf: In that case, `Mountpoint` is not really a mount point of the device, but a device name inside the container. If we do want to make it so, I'd suggest changing the name from `Mountpoint` to something like `Target`, and document that it can either be a mount point of the device, or a device name to be represented inside the container, depending on the value of `Storage.Driver`.

And there must be 1:1 mapping between `pb.Storage.Target` and `spec.Linux.Devices`. Maybe we should make `pb.Storage.Target` a slice so that a device and be represented at multiple places in a container or to multiple containers in the sandbox?

sboeuf:
@laijs @sameo @jodh-intel @mcastelino @grahamwhaley @WeiZhang555 
After some discussion with @bergwolf , we think it would be clearer to add a structure `Device` additionally to `Storage`. This would avoid confusion by trying to make all the different cases fit into the structure `Storage`.

sboeuf:
@bergwolf @laijs PR updated ! Let me know what you think about it.

device.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type deviceDriversHandler func(device pb.Device, spec *pb.Spec) error
+
+var deviceDriversHandlerList = map[string]deviceDriversHandler{
+	"blk": blockDeviceHandler,
+}
+
+func blockDeviceHandler(device pb.Device, spec *pb.Spec) error {
```
bergwolf: Now this is no longer block device specific, maybe just a call it `genericDeviceHandler`?

laijs:
LGTM

grpc.go
```diff
@@ -340,6 +359,10 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	if err := addDevices(req.Devices, req.OCI); err != nil {
```
bergwolf: `addDevices` should be done **before** `addMounts` so that the latter can find the block device.

sboeuf: Yes I have definitely missed that, and this is needed since we expect to be able to mount only after the devices have been discovered inside the VM.

device.go
sboeuf: Not sure I get your point here. We still need a `deviceHandler` for each type of device that is going to be passed, right ?

grpc.go
sboeuf: PR updated.

device.go
bergwolf: True, but `blockDeviceHandler` is not really block device specific now, right? I was hoping that we can use it to handle some other types of devices as well. There are two ways of doing it:
1. rename `blk` to a more generic name for devices
2. define other types of `deviceDriversHandler` and use the same handler to handle them

I'm OK with either way but the handler function name needs also to be generic.

Anyway it is just a minor issue. We can get it in as is and change it later when adding other device types support.

bergwolf:
LGTM

sboeuf:
@bergwolf let me rebase on top of changes from @teawater about the `fmt.Errorf(` and I'll re-submit so that we can merge this.

protocols/grpc/agent.proto
```diff
@@ -196,6 +197,40 @@ message Storage {
...
+	// the device list listed inside the OCI spec. This is used in order
+	// to identify the right device in the spec and update it with the
+	// right options such as major/minor numbers as they appear inside
+	// the VM for instance. Note that an empty ctr_path should be used
+	// to make sure the device handler inside the agent is called, but
+	// no spec update needs to be performed. This has to happen for the
+	// case of rootfs, when a device has to be waited for after it has
+	// been hotplugged. An equivalent Storage entry should be defined if
+	// any mount needs to be performed afterwards.
+	string ctr_path = 4;
```
amshinde: I prefer calling this container_path instead of ctr_path here. We are losing readability for the sake of brevity here.

sboeuf: Addressed ! I have changed it into `container_path`

bergwolf:
It seems ci on Azure is still WiP. Merging.

Issue #140, opened 2018-02-11
disable log color


bergwolf:
The TTY is actually guest's and it malforms console output if save the log somewhere in a file.

PR #142, opened 2018-02-11
log: disable color

bergwolf:
So that the log is actually human readable when being redirected to
a file.

Fixes: #140


jodh-intel:
lgtm

WeiZhang555:
LGTM

PR #146, opened 2018-02-16
agent: Allow to retrieve the process of a stopped container

### LINKED ISSUE #145: agent: Allow a shim to proceed on an already stopped container
A container can be stopped because its lifetime was very short, and the shim might take some time to issue the call to `WaitProcess()`, for this reason, we should not prevent any caller from calling `(s *sandbox) getProcess()` on a `stopped` container.

sboeuf:
A container being in stopped state is not a valid reason for not
accepting a call to getProcess(). Indeed, the explanation behind
this is that we want to make sure a shim can connect after the
process has already terminated. Some processes have a very short
lifetime and the shim might end up calling into WaitProcess()
after that. This does not mean that we cannot retrieve the output
and the exit code, they are waiting for the shim for that. But we
need to let the shim by not returning an error in this case.

Fixes #145

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -195,15 +195,12 @@ func (s *sandbox) getProcess(cid, execID string) (*process, *container, error) {
 		return nil, nil, err
 	}
 
-	status, err := ctr.container.Status()
-	if err != nil {
-		return nil, nil, err
-	}
-
-	if status == libcontainer.Stopped {
-		return nil, nil, fmt.Errorf("Container %s is stopped", cid)
-	}
-
+	// A container being in stopped state is not a valid reason for not
+	// accepting a call to getProcess(). Indeed, we want to make sure a
+	// shim can connect after the process has already terminated. Some
+	// processes have a very short lifetime and the shim might end up
+	// calling into WaitProcess() after this happened. This does not mean
+	// we cannot retrieve the output and the exit code from the shim.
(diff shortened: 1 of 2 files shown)


sboeuf:
cc @sameo @chavafg @bergwolf @laijs @jodh-intel

agent.go
```diff
@@ -195,15 +195,6 @@ func (s *sandbox) getProcess(cid, execID string) (*process, *container, error) {
...
-	status, err := ctr.container.Status()
-	if err != nil {
-		return nil, nil, err
-	}
-
-	if status == libcontainer.Stopped {
-		return nil, nil, fmt.Errorf("Container %s is stopped", cid)
-	}
-
 	proc, err := ctr.getProcess(execID)
```
jodh-intel: It wouldn't hurt to add a comment explaining that this is still a valid call even if the container is stopped.

Also, could you create a unit test for the scenario where the container state is stopped?

sboeuf: I'll add the comment and I'll create a simple unit test working for every cases. The function does not need to be tested in particular for a stopped state as this is the point of the patch to allow this function to be called for any container state.

sboeuf:
@jodh-intel PR updated!

jodh-intel:
@sboeuf - thanks.

lgtm

sameo:
LGTM

PR #148, opened 2018-02-21
grpc: network: add UpdateRoutes

egernst:
To provide a true declarative interface, we
introduce an UpdateRoutes function to our agent protocol.

This function will clear out all existing routes and then attempt to
apply what has been passed over gRPC. In the event that a failure is
observed when applying routes, the error is logged, message is returned
as well as the final state of the ip routes in the guest.

The routes structure in the gRPC prtocol was enhanced in order to
provide handling of Source and Scope fields, which are necessary to be
able to apply functional fresh routes.

Signed-off-by: Eric Ernst <eric.ernst@intel.com>


egernst:
Tested with an updated kata-runtime, but still need to add functional tests before consideration for merge.

egernst:
@WeiZhang555 - FYI.

egernst:
FYI @bergwolf @laijs

protocols/grpc/agent.proto
```diff
@@ -42,6 +42,7 @@ service AgentService {
 	rpc RemoveInterface(RemoveInterfaceRequest) returns (Interface);
 	rpc AddRoute(AddRouteRequest) returns (google.protobuf.Empty);
 	rpc UpdateRoute(UpdateRouteRequest) returns (google.protobuf.Empty);
+	rpc UpdateRoutes(UpdateRoutesRequest) returns (Routes);
```
sboeuf: I am personally fine with this, but I don't see the reason why we would keep `UpdateRoute()` here, seems like it's overlapping.

WeiZhang555: Good point.

Maybe we can keep both APIs temporarily and see if `UpdateRoute` has its own scenario, I think with `UpdateRoutes`, other `route` related APIs including `AddRoute`/`RemoveRoute`/`UpdateRoute` aren't necessary any more.

network.go
```diff
@@ -272,6 +272,124 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+			return nil, err
+		}
+		defer netHandle.Delete()
+	}
+
+	//
+	// First things first, let's blow away all the existing routes.  The updateRoutes function
+	// is designed to be declarative, so we will attempt to create state matching what is
+	// requested, and in the event that we fail to do so, will return the error and final state.
+	//
```
WeiZhang555: The same issue:

I think it's also better to calculate differences with current route rules, and find routes to-delete/to-add, then only apply the adding and deleting for routes in smaller slices.

This can make sure we lost less if any error happens. E.g. after all routes are cleaned, we fail to add any route, so none route will exist.

protocols/grpc/agent.proto
egernst: Agreed others should be removed, including updateRoute, imo.

network.go
egernst: @WeiZhang555 -- but we don't expect this to happen, and we are in a bad state if it does, anyway, right?

WeiZhang555: @egernst Yes, you're right. It will be in bad state if any error happens, and runtime may need to redo everything whatever.

I don't have a strong opinion on how to handle this, this part lgtm.

egernst:
@jodh-intel ; @WeiZhang555 ; @sboeuf ; @sameo  -- Tests have now been added. PTAL

egernst:
@laijs @bergwolf ^^

network.go
```diff
@@ -273,6 +277,129 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+		}
+	}
+
+	//
+	// Set each of the requested routes
+	//
+	// First make sure we set the interfaces initial routes, as otherwise we
+	// won't be able to access the gateway
+	for _, reqRoute := range requestedRoutes.Routes {
+		if reqRoute.Gateway == "" {
```
sboeuf: Nit: Clearer indentation if you do:
```go
if reqRoute.Gateway == "" {
        continue
}
...
```

network.go
```diff
@@ -273,6 +277,129 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+				//If there was an error setting the route, return the
+				//error and the current list of routes on the system
+				goto error_case
+			}
+
+		}
+	}
+	// Take a second pass and apply the routes which include a gateway
+	for _, reqRoute := range requestedRoutes.Routes {
+		if reqRoute.Gateway != "" {
```
sboeuf: Nit: Clearer indentation if you do:
```go
if reqRoute.Gateway != "" {
        continue
}
...
```

network.go
```diff
@@ -273,6 +277,129 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+			if err != nil {
+				agentLog.WithError(err).Error("update Route failed")
+				//If there was an error setting the route, return the
+				//error and the current list of routes on the system
+				goto error_case
+			}
+		}
+	}
+
+error_case:
```
sboeuf: No need for `goto` here, and this code looks like C now lol.
I think doing something like this would work:
```go
func (s *sandbox) updateRoutes(netHandle *netlink.Handle, requestedRoutes *pb.Routes) (resultingRoutes *pb.Routes, err error) {
        defer func() {
                if err != nil {
                        routes, _ = getCurrentRoutes(netHandle)
                }
        }()

        ...
```
You only need to replace your calls to `goto error_case` with `return`

egernst: I was taught in school to never use gotos.  And here I am.  Let me checkout what I can do for defer.

network.go
```diff
@@ -273,6 +277,129 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+	} else {
+		routes = requestedRoutes
+	}
+
+	return routes, err
+}
+
+//getCurrentRoutes is a helper to gather existing routes in gRPC protocol format
+func getCurrentRoutes(netHandle *netlink.Handle) (*pb.Routes, error) {
+
```
sboeuf: Please just check the `netHandle` is not nil before starting using it.

network.go
```diff
@@ -273,6 +277,138 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+			if err != nil {
+				agentLog.WithError(err).Error("update Route failed")
+				//If there was an error setting the route, return the
+				//error and the current routes on the system via defer
+				return
+			}
+		}
+	}
+
+	resultingRoutes = requestedRoutes
```
sboeuf: Not needed, instead return:
```go
return requestedRoutes, nil
```

sboeuf:
LGTM

laijs:
The new API looks to me

network.go
```diff
@@ -273,8 +277,134 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+		return &routes, err
+	}
+
+	for _, route := range finalRouteList {
+		var r pb.Route
+		if route.Dst == nil {
+			r.Dest = ""
+		} else {
+			r.Dest = route.Dst.String()
+		}
```
WeiZhang555: A simplified version:

```
+		var r pb.Route
+		if route.Dst != nil {
 +			r.Dest = route.Dst.String()
 +		}
```

`r.Dest` is supposed to be "" by default when not set value

network.go
```diff
@@ -273,8 +277,134 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+
+		r.Gateway = route.Gw.String()
+		if r.Gateway == emptyRouteAddr {
+			r.Gateway = ""
+		}
```
WeiZhang555: If I understand this correctly, this should be equal to 

```
if route.Gw != nil {
    r.Gateway = route.Gw.String()
}
```

And `emptyRouteAddr` can be removed from everywhere

network.go
```diff
@@ -273,8 +277,134 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+
+		r.Source = route.Src.String()
+		if r.Source == emptyRouteAddr {
+			r.Source = ""
+		}
```
WeiZhang555: Same comment as above

network.go
```diff
@@ -273,8 +277,134 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) *pb.Interface {
...
+		return nil, err
+	}
+	for _, initRoute := range initRouteList {
+		// don't delete routes associated with lo:
+		link, _ := netHandle.LinkByIndex(initRoute.LinkIndex)
+		if link.Attrs().Name == "lo" || link.Attrs().Name == "::1" {
+			continue
+		}
+
+		err = netHandle.RouteDel(&initRoute)
```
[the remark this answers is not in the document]
WeiZhang555: > Does it temporarily disconnect the network even for the untouched rules?

Generally real user container with workloads will only start after network is ready in "pause" container, so I guess this won't be a critical problem.

network.go
egernst: @WeiZhang555  No, netlink returns a string of "<nil">

network.go
egernst: Not for destination. The String() operation will actually panic here otherwise.

network.go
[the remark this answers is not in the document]
egernst: @bergwolf -- yes it does.  As @WeiZhang555 stated, this is to be done at startup usually, I don't anticipate this to be a normal occurrence in a running workload.

@bergwolf  - can I open an issue for a follow on PR to investigate this as an optimization option? I think something functional would be good at this point, and then can address if there are performance improvements available...

network_test.go
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2018 Intel Corporation
```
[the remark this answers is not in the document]
egernst: Its in the commit message  -- I accidentally put 2017 when I created this file last week (it should have been 2018).

network.go
WeiZhang555: No, the "String()" func is called only when "route.Dst != nil", so no panic will occur.

network.go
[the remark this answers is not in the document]
bergwolf: @egernst @WeiZhang555 Yes we can optimize it later. A follow up issue sounds good to me.

network_test.go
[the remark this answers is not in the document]
bergwolf: ah, sorry I missed that.

bergwolf:
It lgtm. Please do not merge until @WeiZhang555 agrees too.

network.go
WeiZhang555: `route.Gw` is type `net.IP`, when `route.Gw == nil`, `route.Gw.String() == "<nil>"` according to my test.

According to this,  in your codes:
```
+		r.Gateway = route.Gw.String()
+		if r.Gateway == emptyRouteAddr {
 +			r.Gateway = ""
 +		}
```
is equal to 
```
+		r.Gateway = route.Gw.String()
+		if route.Gw == nil {
 +			r.Gateway = ""
 +		}
```

r.Gateway has a default value of "" so we only need to handle situation of `if route.Gw != nil`. A simplified version is 
```
if route.Gw != nil {
    r.Gateway = route.Gw.String()
}
```

Hope I'm making myself clear

network.go
egernst: Ah, that's what I get for checking the review so late at night :). 

Yes, this makes sense @WeiZhang555  -- I'll make that change now!

network.go
egernst: Thanks for spelling it out -- you're right -- let me make this change now...

egernst:
@WeiZhang555 -- tested and fixed - thanks for the help.

network_test.go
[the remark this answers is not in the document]
egernst: no problem -- I understand, normally that shouldn't be updated

network.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+const (
+	emptyRouteAddr = "<nil>"
+)
+
```
WeiZhang555: Unused variable. need to be removed.

WeiZhang555:
LGTM.

Need to remove unused variable in follow up PR, including optimization @bergwolf stated. 
I'll merge this first :smile:

PR #163, opened 2018-03-08
device: Add SCSI block device support

### LINKED ISSUE #162: device: Add SCSI block device support
The Kata agent has to support block devices (rootfs or device passthrough) that might be passed through virtio-scsi.

sboeuf:
Recently, the support for block devices (rootfs or device passthrough)
plugged through the VM with virtio-blk have been added.

This commit follows this by allowing block devices plugged through the
VM using virtio-scsi to be supported and properly processed from the
agent.

One thing to notice here is that latest versions of Qemu (>= 2.9) have
some stability issues using virtio-blk, and the use of virtio-scsi is
a good way to work around this current limitation.

Fixes #162


sboeuf:
@bergwolf @laijs @amshinde @sameo Please let me know what you think about this. I have it working with a corresponding patch on virtcontainers. I'll fix/add some unit tests to this PR in the meantime.

device.go
```diff
@@ -41,8 +62,51 @@ func blockDeviceHandler(device pb.Device, spec *pb.Spec) error {
...
+	devPath, err := getSCSIDevPath(device.Id)
+	if err != nil {
+		return err
+	}
+	device.VmPath = devPath
+
+	// If no ContainerPath is provided, this means we don't expect the
+	// device to be updated in the specification. Instead, this means
+	// this device is the rootfs, and the mount list should be updated
+	// accordingly.
```
laijs: this document should be put in the API's documents too.

sboeuf: Ok let me try to explain the situation here. In case of a rootfs, we need to both use `Device` and `Storage` because we need to detect the device being plugged inside the VM, and after it showed up, and that we have identified the name of the device, we need a `Storage` to mount the device to the expected rootfs path with the provided option such as `ext4` for instance.
This needs to be a 2 steps things IMO.

One way to avoid using empty `ContainerPath` (which I don't really like either), would be to add a new field `RootFs bool` to both structures like this:
```go
type Device struct {
        ...
        RootFs bool
        ...
}

...

type Storage struct {
        ...
        RootFs bool
        ...
}
```

WDYT ?

device.go
```diff
@@ -41,8 +62,51 @@ func blockDeviceHandler(device pb.Device, spec *pb.Spec) error {
...
+		return grpcStatus.Errorf(codes.Internal, "Mount list is empty")
+	}
+
+	for _, mnt := range mounts {
+		if mnt == nil {
+			continue
+		}
+
+		// An empty source here means this is the rootfs and it can
+		// be replaced with the provided value of the device VM path.
```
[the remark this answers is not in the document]
sboeuf: I agree the empty `Source` is not really clean here, and I think using a field `RootFs bool` into `Storage` structure could be easier to understand.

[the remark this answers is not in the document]
laijs: `Storage` has a `MountPoint` which defined by the runtime and it will be located in the VM.
After `Storage` handled in the agent, all the dir/files of MountPoints can be used and all the path (including oci.Root.Path & oci.Mount.Srouce) in the ocispec can use these paths(Storage.MountPoint). `RootFs bool` is needless.

device.go
laijs: > we need to both use Device and Storage because we need to detect the device being plugged inside the VM

using Storage only is enough, 'detecting the device being plugged inside the VM' should be done in the storage's driver.

sboeuf:
@laijs PR updated with what we've discussed, PTAL.

sboeuf:
@laijs @bergwolf I'll add/fix unit tests if you're okay with those changes.

protocols/grpc/agent.proto
```diff
@@ -189,31 +189,62 @@ message UpdateRoutesRequest {
...
+	string source = 3;
+	// Fstype represents the filesystem that needs to be used to mount the
+	// storage inside the VM. For instance, it could be "xfs" for block
+	// device, "9p" for shared filesystem, or "tmpfs" for shared /dev/shm.
+	string fstype = 4;
+	// Options describes the additional options that might be needed to
+	// mount properly the storage filesytem.
+	repeated string options = 5;
+	// MountPoint refers to the path where the storage should be mounted
+	// inside the container. This is used to determine which Mount from the
```
bergwolf: The original design is that `Mountpoint` is the mount point path outside of the container, and then it is bind mounted to the path inside the container (as specified by the `Mount` of the oci spec). This way, it is possible to share one `Storage` with multiple `Mount` so that we can mount one volume to different mount points in the pod.

protocols/grpc/agent.proto
```diff
@@ -189,31 +189,62 @@ message UpdateRoutesRequest {
...
+	// OCI specification should be updated, otherwise the source will be
+	// wrong.
+	// The rootfs case is specific since no Mount will be listed from the
+	// OCI specification.
+	string mount_point = 6;
+	// Rootfs is a simple flag indicating if the current storage represents
+	// a root filesystem or a simple volume.
+	bool rootfs = 7;
```
bergwolf: why do you need to differentiate them at the `Storage` resource level?

sboeuf: Because doing this from the agent code makes things less clean. If the caller can say this is a rootfs, it is clearly identified that we need to handle this a bit differently.

protocols/grpc/agent.proto
bergwolf: Then libcontainer has to understand all kinds of methods to mount the storage. E.g., can it understand an NFS share source?

Another drawback of the new approach is that we have to define multiple storage resources if we want to mount the same volume to different mount points in the container.

protocols/grpc/agent.proto
bergwolf: But we do not really need the special case for rootfs. If we stick with the original design, the `Rootfs` in OCI spec can be the `MountPoint` of the storage and there is no difference handling it in the agent.

protocols/grpc/agent.proto
laijs: MountPoint is the path in the VM.  (pb.Storage represents containerd/snapshotter docker/storage-driver.) ocispec.Root.Path oci.Mount.Source can use MountPoint or subdirectory of MountPoint.

sboeuf:
@laijs @bergwolf PR updated. I have removed `Rootfs` field, and now `MountPoint` is the path inside the VM, not the container.

sboeuf:
cc @amshinde

mount.go
```diff
@@ -23,6 +24,8 @@ const (
 	devPrefix      = "/dev/"
 	timeoutHotplug = 3
 	mountPerm      = os.FileMode(0755)
+
+	driver9pType = "9p"
```
bergwolf: nits: please define it together with other types like `driverBlkType` and `driverSCSIType`.

sboeuf: You would prefer that I put them all inside device.go ?

bergwolf: Or any other place you feel comfortable with. I just want them to be put together.

sboeuf: Ok fair enough. Thanks for the review, I'll fix unit tests on Monday.

sboeuf:
@bergwolf @laijs @sameo @amshinde PR is ready, please take a look.

amshinde:
@sboeuf Sure, Will take a look today.

mount.go
```diff
@@ -143,37 +143,78 @@ func parseMountFlagsAndOptions(optionList []string) (int, string, error) {
...
-	return mountList, nil
-}
-
-func removeMounts(mounts []string) error {
-	for _, mount := range mounts {
-		if err := syscall.Unmount(mount, 0); err != nil {
-			return err
+		if mountPoint != "" {
+			// Prepend mount point to mount list.
+			mountList = append([]string{mountPoint}, mountList...)
```
amshinde: Any reason why are you prepending here ? Why not just append with  `mountList = append(mountList, mountPoint)` ?

protocols/grpc/agent.proto
```diff
@@ -189,31 +189,55 @@ message UpdateRoutesRequest {
...
+	// DriverOptions allows the caller to define a list of options such
+	// as block sizes, numbers of luns, ... which are very specific to
+	// every device and cannot be generalized through extra fields.
+	repeated string driver_options = 2;
+	// Source can be anything representing the source of the storage. This
+	// will be handled by the proper handler based on the Driver used.
+	// For instance, it can be a very simple path if the caller knows the
+	// name of device inside the VM, or it can be some sort of identifier
+	// to let the agent find the device inside the VM.
+	string source = 3;
```
amshinde: Can you name it something else then if this could be a path or an identifier. "source" tends to imply a path.

protocols/grpc/agent.proto
```diff
@@ -189,31 +189,55 @@ message UpdateRoutesRequest {
...
+	// provided VmPath instead.
 	string id = 1;
-	// type defines the type of device described. This can be "blk",
+	// Type defines the type of device described. This can be "blk",
 	// "scsi", "vfio", ...
 	// Particularly, this should be used to trigger the use of the
 	// appropriate device handler.
 	string type = 2;
-	// vm_path can be used by the caller to provide directly the path of
+	// VmPath can be used by the caller to provide directly the path of
```
amshinde: Why have you changed this?

grpc.go
```diff
@@ -688,7 +668,7 @@ func (a *agentGRPC) CreateSandbox(ctx context.Context, req *pb.CreateSandboxRequ
 		}
 	}
 
-	mountList, err := addMounts(req.Storages)
+	mountList, err := addStorages(req.Storages)
```
amshinde: So the rootfs is passed as part of Storage. But one thing not clear to me is how we indicate the rootfs to libcontainer.  I remember OCI config has explicit field called "Rootfs". Looks like we are not populating that .

device.go
```diff
@@ -22,34 +24,67 @@ import (
...
+	devPath, err := getSCSIDevPath(device.Id)
+	if err != nil {
 		return err
 	}
+	device.VmPath = devPath
+
+	return updateSpecDeviceList(device, spec)
+}
 
+func updateSpecDeviceList(device pb.Device, spec *pb.Spec) error {
```
amshinde: Can you add documentation comments for this function as well.

device.go
```diff
@@ -7,6 +7,8 @@
 package main
```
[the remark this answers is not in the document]
sboeuf: I'll change that.

mount.go
```diff
@@ -143,37 +143,78 @@ func parseMountFlagsAndOptions(optionList []string) (int, string, error) {
...
+func mountStorage(storage pb.Storage) error {
+	flags, options, err := parseMountFlagsAndOptions(storage.Options)
+	if err != nil {
+		return err
+	}
+
+	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
+}
+
+func addStorages(storages []*pb.Storage) ([]string, error) {
```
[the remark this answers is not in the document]
sboeuf: Sure

protocols/grpc/agent.proto
sboeuf: Let's keep it with source here since this PR is not modifying it. I think `Source` is pretty generic and can both fit a path or and ID. But we could discuss this in a separate PR dedicated to this change.

protocols/grpc/agent.proto
sboeuf: I have changed it so that it matches the `agent.pb.go` variable format in Golang.

grpc.go
sboeuf: The OCI spec has a field `ociSpec.Root.Path` specifying the rootfs. But in this case, we don't need to make the diff between a rootfs and a generic storage since for both we expect to mount them somewhere. Different handlers (`blk`, `scsi`, `9p`) will act differently.

device.go
sboeuf: Yes sure.

sboeuf:
@amshinde PR updated.

grpc.go
amshinde: @sboeuf I assume the rootfs is mounted from the Storage[] depending on its location i.e it will be the first to be mounted. I do not particularly like this as this makes the code unclear and prone to errors. I would have wanted to see an explicit field that indicates the rootfs for a container. The last I saw libcontainer does a few setup steps for setting up rootfs. I am wondering if we are not using the libcontainer api the way it is meant to be used.

sboeuf:
@amshinde I have updated the PR, adding more verbose comments about rootfs, storages and devices.

laijs:
LGTM

mount.go
```diff
@@ -143,37 +144,106 @@ func parseMountFlagsAndOptions(optionList []string) (int, string, error) {
...
+
+	return nil
+}
+
+// storageDriversHandler is the type of callback to be defined to handle every
+// type of storage driver.
+type storageDriversHandler func(storage pb.Storage) (string, error)
+
+// storageDriversHandlerList lists the supported drivers.
+var storageDriversHandlerList = map[string]storageDriversHandler{
```
jodh-intel: How about making an `announce()` function that is called after `initLogger()` and that adds log entries for:

- `agentLog.WithField("version", version).Info()` (move it from `initLogger()`).
- `agentLog.WithField("storage-drivers", ...)` which shows the keys of this map.

device.go
```diff
@@ -21,29 +23,74 @@ import (
...
 type deviceDriversHandler func(device pb.Device, spec *pb.Spec) error
 
 var deviceDriversHandlerList = map[string]deviceDriversHandler{
-	"blk": blockDeviceHandler,
+	driverBlkType:  deviceBlkBlockDeviceDriverHandler,
+	driverSCSIType: deviceSCSIBlockDeviceDriverHandler,
 }
 
-func blockDeviceHandler(device pb.Device, spec *pb.Spec) error {
+func deviceBlkBlockDeviceDriverHandler(device pb.Device, spec *pb.Spec) error {
```
jodh-intel: `thisNameIsCrazyLongBlkBlockBlkBlahBleurk()` and a bit confusing (although precise :smile:)

How about changing the names to:

- `genericBlockDeviceHandler()`
- `scsiDeviceHandler()`

mount.go
```diff
@@ -143,37 +144,106 @@ func parseMountFlagsAndOptions(optionList []string) (int, string, error) {
...
+	driver9pType:   storage9pDriverHandler,
+	driverBlkType:  storageBlkBlockDeviceDriverHandler,
+	driverSCSIType: storageSCSIBlockDeviceDriverHandler,
```
jodh-intel: Again, these names are really long. Maybe this is a sign the code should be broken into discrete packages to avoid `crazyLongNamesToAvoidCollidingInTheGlobalMainNamespace`?

device.go
sboeuf: Yes, I'm going to simplify them

mount.go
jodh-intel: Yes, it's just an idea - when the agent starts it could "announce" a few details about itself in the log (version and what types of devices it can handle).

sboeuf:
@jodh-intel I have updated the function names, let me know if you think this is better now.

mount.go
jodh-intel: Sure, sounds good.

jodh-intel:
lgtm

sboeuf:
@jodh-intel could you go ahead and merge it please (since everybody approved the PR) ?

jodh-intel:
@sboeuf - we're seeing CI failures that might be related:

```
Stderr: docker: Error response from daemon: OCI runtime create failed: rpc error: code = InvalidArgument desc = Unknown storage driver "": unknown.
```
```
Stderr: losetup: /tmp/dd591751879: failed to set up loop device: Device or resource busy
```

@chavafg - or is this related to the container manager utility dyt?

sboeuf:
Problem is, we need virtcontainers to be updated. And virtcontainers needs this PR to land so that we can revendor the agent. We need to add to our CI a way to test a PR with a `Depends-on` flag.

jodh-intel:
It sounds like we're in a "chicken and egg" situation here and will have to either force-merge this PR (or https://github.com/containers/virtcontainers/pull/666 according to @sboeuf).

@kata-containers/agent - Can another team member ack the proposal to force-merge this PR please?

amshinde:
@jodh-intel @sboeuf I am ok with force merging this PR. We will be broken till the SCSI PR lands for virtcontainers.

jodh-intel:
... Wait!

Kidding.

Thanks @amshinde :smile:

amshinde:
Haha ..too late now @jodh-intel :)

PR #176, opened 2018-03-19
CI: Require pullapprove ack for protocol changes

### LINKED ISSUE #175: Require extra pullapprove approval for protocol changes
Extra care needs to be taken when changing a gRPC protocol buffer file (`*.proto`) since that defines the language the overall system uses to communicate.

Specifically, any changes are likely to require:

- careful testing.
- an update (agent re-vendor) for all other system components using the protocol
  (in other words, most of them).

See: https://github.com/kata-containers/kata-containers/issues/17.

jodh-intel:
A change to any of the gRPC protocol buffer files (`*.proto`) should
require an additional approval due to the potential impact it could
have across the system.

Fixes #175.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -41,3 +41,16 @@ groups:
           - "*.md"
         exclude:
           - "vendor/*"
+
+  protocol-changes:
+    required: 2
+    teams:
+      - architecture-committee
+      - builder
+      - packaging
+    conditions:
+      files:
+        include:
+          - "*.proto"
+        exclude:
+          - "vendor/*"


jodh-intel:
We might want to tweak the list of teams I've added here. Here's my rationale for the currently listed team for this change:

| Team | Rationale |
|--|--|
| `architecture-team` | Should be aware of any architectural changes. |
| `builder` | Need to be aware of such changes to potentially create a new image version containing the new agent. |
| `packaging` | Need to be aware of any system-breaking protocol change (to allow new packages to be generated [once we are actually producing packages]). |

sboeuf:
LGTM

.pullapprove.yml
```diff
@@ -41,3 +41,16 @@ groups:
           - "*.md"
         exclude:
           - "vendor/*"
+
+  protocol-changes:
+    required: 1
```
grahamwhaley: `required: 1` feels a little slim to me - if it is architectural change then I would not be against increasing this .

sboeuf: Oh yeah I've missed that actually. @jodh-intel could you please put `2` or `3` here ?
I think `2` is already the default for the agent repo, isn't it ?

jodh-intel: Ha - I put in `1` as I was expecting folk to fight against `>1` ;)

I''ve now changed it to `2`. The ideal in my mind would be `1` from the arch team and `1` from one of the other two teams. However, fwics, that would require a bit of extra bloat in the config as we'd need two "protocol-changes" groups - one just for the arch team and another for the other two teams.

sboeuf: `2` is fine for now, we'll introduce protocol-changes group in a follow-up PR I guess

laijs:
LGTM

PR #178, opened 2018-03-19
Skip more tests if not root

### LINKED ISSUE #177: Skip more tests when running as non-root user
Some of the tests are failing under Travis as they are running as a non-root user, but the tests require `root` privileges.

jodh-intel:
The unit tests in this repository are ultimately run by the
`go-test.sh` script in the `tests` repository. This script actually runs
the unit-tests *twice* **if** it is invoked as a non-`root` user: once
as the non-`root` user and once as `root`. This is done to cover all
possible scenarios: certain tests require super-user privileges whilst
other tests are only valid when running as a non-privileged user. The
script then combines the test results to calculate a coverage
percentage.

The Travis test runs are currently failing (in the first non-`root`
"run") as some of the tests require root privileges.

Make extra calls to `skipUnlessRoot()` for those failing tests to allow
them to be skipped when run as a non-`root` user and then run as `root`
in the second "run".

Fixes #177.

### DIFF
--- a/agent_test.go
+++ b/agent_test.go
@@ -20,6 +20,12 @@ const (
 	testContainerID = "testContainerID"
 )
 
+func skipUnlessRoot(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("Test disabled as requires root user")
+	}
+}
+
 func TestClosePostStartFDsAllNil(t *testing.T) {
 	p := &process{}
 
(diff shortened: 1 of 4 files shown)


sboeuf:
LGTM

sboeuf:
@bergwolf @laijs @egernst PTAL

laijs:
LGTM

PR #180, opened 2018-03-20
Add announce log call

jodh-intel:
Add a function that announces agent capabilities (version, device and storage drivers and total guest kernel memory).

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -75,11 +75,13 @@ type namespace struct {
 	exitCodeCh <-chan int
 }
 
-var agentLog = logrus.WithFields(logrus.Fields{
+var agentFields = logrus.Fields{
 	"name":   agentName,
 	"pid":    os.Getpid(),
 	"source": "agent",
-})
+}
+
+var agentLog = logrus.WithFields(agentFields)
 
 // version is the agent version. This variable is populated at build time.
 var version = "unknown"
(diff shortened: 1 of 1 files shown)


sboeuf:
@bergwolf @laijs PTAL

agent.go
```diff
@@ -416,7 +508,10 @@ var initRootfsMounts = []initMount{
 // initAgentAsInit will do the initializations such as setting up the rootfs
 // when this agent has been run as the init process.
 func initAgentAsInit() error {
-	fmt.Printf("initAgentAsInit(), agent version: %s\n", version)
+	err := announce()
```
bergwolf: This will fail because `/proc` is not mounted yet.

jodh-intel: Good catch @bergwolf! I've moved the call to after the mounts now. However, that might not be what you want since you won't get the version details displayed if a mount fails?

bergwolf: If a mount fails at this stage, it is mostly unrelated to the agent anyway, -- most likely a kernel bug or some kernel features are not enabled. We can get related info from runtime then. I can live with it.

bergwolf:
lgtm

agent.go
jodh-intel: ok - thanks @bergwolf.

PR #182, opened 2018-03-21
protocols/grpc: fix CPU hotplug race condition

### LINKED ISSUE #181: CPU hotplug race condition
all vCPUs are not connected in systems running low on resources,  for example in a virtual machine with 1 vCPU + 256 MB of RAM, 

```sh
docker run -ti --runtime kata-runtime --rm --cpu-quota 30000 --cpu-period 10000 busybox sh -c "sleep 5; nproc"
```
returns 2, when the expected result is 4 (1 cold plugged vCPU + 3 hot plugged vCPUs)

devimc:
With this patch the runtime will communicate to the agent the
number of vCPUs that were hot added, allowing to the agent online all vCPUs.
The agent will try to online all vCPUs 5 times, waiting 500 milliseconds
in each iteration, this is needed since when the runtime calls to QMP
`device_add`, QEMU doesn't allocate all vCPUs inmediatelly.

fixes #181


grpc.go
```diff
@@ -56,40 +58,85 @@ type onlineResource struct {
...
-	resourceList := []onlineResource{
-		{
-			sysfsOnlinePath: sysfsCPUOnlinePath,
-			regexpPattern:   cpuRegexpPattern,
-		},
-		{
-			sysfsOnlinePath: sysfsMemOnlinePath,
-			regexpPattern:   memRegexpPattern,
-		},
+var onlineCPUMemMux sync.Mutex
```
sboeuf: Looks like a multiplexer with the naming `...Mux`. Please call this `onlineCPUMemLock`

grpc.go
```diff
@@ -56,40 +58,85 @@ type onlineResource struct {
...
+
+const onlineCPUMemWaitTime = 500 * time.Millisecond
```
sboeuf: This looks like a lot ! We cannot afford to wait for so much time between two retries. Something like 50ms seems more appropriate, and maybe increase the number of retries if you think that in some cases it might take some time.

grpc.go
```diff
@@ -752,7 +799,7 @@ func (a *agentGRPC) UpdateRoutes(ctx context.Context, req *pb.UpdateRoutesReques
 }
 
 func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemRequest) (*gpb.Empty, error) {
-	go onlineCPUMem()
+	go onlineCPUMem(req)
```
sboeuf: I think this should not be spawned into a go routine. This is the caller responsibility to know if it should wait or not for the return of this function.
Also, if we look at this patch, we expect to be able to return a valid error in case we could not hotplug the amount of CPUs provided as parameter.

protocols/grpc/agent.proto
```diff
@@ -187,6 +187,8 @@ message UpdateRoutesRequest {
 }
 
 message OnlineCPUMemRequest {
+	// Specify the number of CPUs that were added
+	uint32 cpus = 1;
```
sboeuf: Was thinking that `nb_cpus` or something with number in it, might be more appropriate as the parameter here, since we're providing a number of CPUs.

grpc.go
```diff
@@ -56,40 +58,85 @@ type onlineResource struct {
...
+func onlineCPUMem(req *pb.OnlineCPUMemRequest) {
+	onlineCPUMemMux.Lock()
+	defer onlineCPUMemMux.Unlock()
+
+	cpuResource := onlineResource{
+		sysfsOnlinePath: sysfsCPUOnlinePath,
+		regexpPattern:   cpuRegexpPattern,
+	}
+
+	err := forceOnlineResources(cpuResource, req.Cpus, onlineCPUMaxTries)
```
sboeuf: Having functions `onlineCPUResources()` and `onlineMemResources()` seems more appropriate to make the code more readable.
```go
func onlineCPUResources(nbCPU int) error {
        resource := onlineResource{
		sysfsOnlinePath: sysfsCPUOnlinePath,
		regexpPattern:   cpuRegexpPattern,
	}

	var count uint32
	for i := uint32(0); i < onlineCPUMaxTries; i++ {
		r, err := onlineResources(resource)
		if err != nil {
			return err
		}
		count += r
		if count == nbCPU {
			return nil
		}
		time.Sleep(onlineCPUMemWaitTime)
	}

	return fmt.Errorf("only %d of %d were connected", count, expectedResources)
}

func onlineMemResources() error {
        resource := onlineResource{
		sysfsOnlinePath: sysfsMemOnlinePath,
		regexpPattern:   memRegexpPattern,
	}

        _, err := onlineResources(resource)
        return err
}
```

jodh-intel:
> when the runtime calls to QMP device_add, QEMU doesn't allocate all vCPUs inmediatelly.

I think this the piece of code you're referring to:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu.go#L712..L736

If so, I was about to say that QMP needs a transaction message, when -- guess what? -- I found that it already has!

- https://github.com/qemu/qemu/blob/master/qapi/transaction.json#L87..L151

Hence, we should be able to replace that loop with a call to something like:

```
{
  "execute": "transaction",                    
  "arguments": { "actions": [                  
    { 
      "type": "device_add",
      "data" : { 
        "id": "cpu-0", "socket-id": "...", "core-id": "...", "thread-id": "..." 
      }   
    },
    { 
      "type": "device_add",
      "data" : { 
        "id": "cpu-1", "socket-id": "...", "core-id": "...", "thread-id": "..." 
      }   
    },
    { 
      "type": "device_add",
      "data" : { 
        "id": "cpu-2", "socket-id": "...", "core-id": "...", "thread-id": "..." 
      }   
    } 
  ]                                            
}
```
It would be useful to have this feature in https://github.com/intel/govmm.

/cc @rarindam, @markdryan.

markdryan:
@jodh-intel I'd never seen that transaction command before.  I'd assumed it was new when you mentioned it but looking at the qmp docs it seems to have been around since 1.0.

The way we currently model QMP commands doesn't really lend itself to a generic transaction mechanism.  Perhaps this is a sign that the API is wrong and should be changed.  In the short term it should be possible to add a ExecuteAddDevices that adds multiple devices transactionally.

The other comment I would make is, would using transactions actually solve the race condition?  Does qmp generate an event when adding a device?  Perhaps ExecuteDeviceAdd just needs to wait for that.  Modifying ExecuteDeviceAdd might be a little risky, but we could add a new blocking version of that function and migrate to it over time.

devimc:
Hi @jodh-intel nice finding!
but I think `transactional` messages won't fix the race condition, since it only guarantee a failure if any operation fail, the problem with the race condition is that QEMU does not allocate and assign vCPUs intermediately, because it has to negotiate the resources with KVM.
For example, if 5 vCPUs are hot added to the VM, when `onlineCPUMem` is called,  probably only 3 vCPUs have been allocated and assigned to the VM, that means the agent will online only 3 vCPUs, the other two vCPUs will be allocated and assigned later, but the agent won't be able to online them.

devimc:

grpc.go
```diff
@@ -56,40 +58,93 @@ type onlineResource struct {
...
+			continue
+		}
 
-			onlinePath := filepath.Join(resource.sysfsOnlinePath, file.Name(), "online")
+		if strings.Trim(string(status), "\n\t ") == "0" {
 			ioutil.WriteFile(onlinePath, []byte("1"), 0600)
+			count++
+		}
+
+		if nbResources != -1 && count == uint32(nbResources) {
```
sboeuf:
```go
if nbResources > 0 && count == uint32(nbResources) {
```

grpc.go
```diff
@@ -56,40 +58,93 @@ type onlineResource struct {
...
+		onlinePath := filepath.Join(resource.sysfsOnlinePath, file.Name(), "online")
+		status, err := ioutil.ReadFile(onlinePath)
+		if err != nil {
+			// resource cold plugged
+			continue
+		}
 
-			onlinePath := filepath.Join(resource.sysfsOnlinePath, file.Name(), "online")
+		if strings.Trim(string(status), "\n\t ") == "0" {
 			ioutil.WriteFile(onlinePath, []byte("1"), 0600)
```
markdryan: You should probably check the error here.  What you should do I'm not sure.  Perhaps simply continue.  Looking and the old code no action was taken after this line, so although the error wasn't discarded, ignoring it was harmless.  But with these changes I guess it's not.

grpc.go
```diff
@@ -56,40 +58,93 @@ type onlineResource struct {
...
+
+func onlineCPUResources(nbCpus uint32) error {
+	resource := onlineResource{
+		sysfsOnlinePath: sysfsCPUOnlinePath,
+		regexpPattern:   cpuRegexpPattern,
+	}
+
+	var count uint32
+	for i := uint32(0); i < onlineCPUMaxTries; i++ {
+		r, err := onlineResources(resource, int32(nbCpus-count))
```
markdryan: Do the directory contents of resource.sysfsOnlinePath change between invocations of this function?  If not do you need to read the directory up to 10 times and check the online files 10 times.  I'm suppose I'm thinking of a situation in which in the first invocation you online 4 out of 5 cpus.  If I read the code correctly you'd then re-read the online files associated with those CPUs up to 9 more times.

devimc: In systems running low on resources, yes. Read the directory in each iteration is needed

grpc.go
```diff
@@ -56,40 +61,95 @@ type onlineResource struct {
...
+		if err != nil {
+			// resource cold plugged
+			continue
+		}
 
-			if !matched {
-				continue
+		if strings.Trim(string(status), "\n\t ") == "0" {
+			if err := ioutil.WriteFile(onlinePath, []byte("1"), 0600); err != nil {
+				return count, err
```
markdryan: If you get one write error you'll give up entirely, i.e., the for loop in onlineCPUResources will terminate.  Is this intentional?

grpc.go
```diff
@@ -752,7 +807,10 @@ func (a *agentGRPC) UpdateRoutes(ctx context.Context, req *pb.UpdateRoutesReques
 }
 
 func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemRequest) (*gpb.Empty, error) {
-	go onlineCPUMem()
+	if !req.Wait {
```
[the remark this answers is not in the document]
devimc: In order to have a fast boot time, the go routine should run inside the VM, otherwise the runtime fails with next error

```
rpc error: code = Unavailable desc = transport is closing.
```

grpc.go
devimc: fixed, thanks

markdryan: I'm not sure it's fixed correctly though.  If you get one write error you'll give up entirely, i.e., the for loop in onlineCPUResources will terminate. Is this intentional?

grpc.go
devimc: you're right, would be better to log the error and continue

devimc:
@markdryan @sboeuf  PTAL

devimc:
:slightly_frowning_face:

sboeuf:
@bergwolf @laijs PTAL

grpc.go
```diff
@@ -56,40 +61,96 @@ type onlineResource struct {
...
+			if err := ioutil.WriteFile(onlinePath, []byte("1"), 0600); err != nil {
+				agentLog.WithField("online-path", onlinePath).WithError(err).Errorf("Could not online resource")
 				continue
 			}
+			count++
+		}
 
-			onlinePath := filepath.Join(resource.sysfsOnlinePath, file.Name(), "online")
-			ioutil.WriteFile(onlinePath, []byte("1"), 0600)
+		if nbResources > 0 && count == uint32(nbResources) {
```
bergwolf: nits: move it to after `count++`. It only makes sense to check when `count` increases.

devimc:
@bergwolf changes applied , thanks

devimc:
ping @kata-containers/agent

grpc.go
```diff
@@ -43,10 +45,13 @@ const (
 
 // CPU and Memory hotplug
 const (
+	cpuRegexpPattern = "cpu[0-9]*"
+	memRegexpPattern = "memory[0-9]*"
```
jodh-intel: Nit: shouldn't these be strictly?:

```
cpuRegexpPattern = "cpu[0-9][0-9]*"
memRegexpPattern = "memory[0-9][0-9]*"
```

jodh-intel: It's fine as long as files called `cpu` or `memory` are never created in that directory.

grpc.go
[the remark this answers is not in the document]
sboeuf: Ok makes sense!

protocols/grpc/agent.proto
```diff
@@ -187,6 +187,13 @@ message UpdateRoutesRequest {
 }
 
 message OnlineCPUMemRequest {
+	// Wait specifies if the caller waits for the agent to online all resources.
+	// If true the agent returns until all resources have been connected, else all
```
sboeuf: s/`until`/`once`/

sboeuf: s/`else`/`otherwise`/

protocols/grpc/agent.proto
```diff
@@ -187,6 +187,13 @@ message UpdateRoutesRequest {
...
+	// resources are connected asynchronously and the agent returns immediately
```
sboeuf: s/`immediately`/`immediately.`/

protocols/grpc/agent.proto
```diff
@@ -187,6 +187,13 @@ message UpdateRoutesRequest {
...
+	bool wait = 1;
+
+	// NbCpus Specifies the number of CPUs that were added and the agent has to online
```
sboeuf: s/`Specifies`/`specifies`/
s/`online`/`online.`/

sboeuf:
@markdryan could you give this a last review before we merge ?

grpc.go
```diff
@@ -56,40 +61,99 @@ type onlineResource struct {
...
-			regexpPattern:   memRegexpPattern,
-		},
+var onlineCPUMemLock sync.Mutex
+
+const onlineCPUMemWaitTime = 100 * time.Millisecond
+
+const onlineCPUMaxTries = 10
+
+// online resources, nbResources specifies the maximum number of resources to online.
+// If nbResources is -1 then there is no limit, all resources are connected
```
markdryan: Looking at the code, there's no limit if nbResources==0 either.

markdryan:
@sboeuf Looks good to me.  I think my comments have been addressed.  I made one small comment on a comment.

sboeuf:
@devimc please address both comments from @markdryan and myself, and we're good to merge this :)

devimc:
@markdryan @sboeuf  changes applied, thanks

sboeuf:
Thanks @devimc, let's merge it !

Issue #183, opened 2018-03-22
verbose grpc request details


bergwolf:
When agent log is set to debug level, we can print details of each grpc request to better trace what's happening between runtime and agent.

PR #184, opened 2018-03-22
grpc: print every grpc request detail on debug

bergwolf:
When agent log debug is enabled, print every grpc request so that
we get clear idea of what is requested.

Fixes: #183

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	gpb "github.com/gogo/protobuf/types"
 	pb "github.com/kata-containers/agent/protocols/grpc"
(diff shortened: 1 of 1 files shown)


jodh-intel:
I like the idea but a couple of thoughts:

- It almost feels like we need another thin layer to add in the debug calls before the functions in `grpc.go` get called. Or some sort of template code that `protoc` could inject maybe?

- `spew` seems to be designed for multi-line human readable output (it includes lots of newlines which introduces noise in the logs, even though they are encoded). I'm wondering if we could just do the following as that would be more "in keeping" with the structured logging approach:

  ```
  func Dump(v interface{}) string {              
    bytes, _ := json.Marshal(v)                
    return string(bytes)                       
  }

    "grpc-function", "create-container",
    "grpc-request", Dump(req),
  }).Debug()
  ```

devimc:
lgtm

Issue #188, opened 2018-03-22
ci: lib: allow tests repo url to be overwridden


grahamwhaley:
Only use the default if a tests_repo is not already set.

PR #189, opened 2018-03-22
ci: lib: allow override of tests_repo

grahamwhaley:
If a tests_repo is already set, use that.

Fixes: #188

Signed-off-by: Graham whaley <graham.whaley@intel.com>


devimc:
lgtm

PR #192, opened 2018-03-26
device: VmPath can be empty if an Id is provided

### LINKED ISSUE #191: device: VmPath can be empty if an Id is provided
It is not an error to have an empty `VmPath`. The error is to have both `VmPath` and `Id` being empty.

sboeuf:
The statement returning an error in case VmPath is empty is wrong
because an Id can be provided instead. This patch fixes this behavior
and generates an error only if VmPath and Id are both empty.

Fixes #191

### DIFF
--- a/device.go
+++ b/device.go
@@ -310,9 +310,9 @@ func addDevice(device *pb.Device, spec *pb.Spec) error {
 			"invalid type for device %v", device)
 	}
 
-	if device.VmPath == "" {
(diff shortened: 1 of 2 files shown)


sboeuf:
@jodh-intel @bergwolf @egernst @laijs

bergwolf:
lgtm

jodh-intel:
lgtm

PR #194, opened 2018-03-28
grpc: implement ListProcesses

### LINKED ISSUE #193: Implement list processes
# Description of problem

```
kata-containers ps $CID
```
returns nothing

# Expected result

```
kata-runtime ps $CID
UID         PID   PPID  C STIME TTY          TIME CMD
root        134    107  0 20:07 ?        00:00:00 bash
```

# Actual result

```
kata-runtime ps $CID
```

devimc:
ListProcesses returns a list of running processes inside
the container, this function should be called by the runtime
in the ps command implementation.

fixes #193


grpc.go
```diff
@@ -536,6 +541,91 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
+		return resp, fmt.Errorf("invalid format option")
+	}
+
+	psArgs := req.Args
+	if len(psArgs) == 0 {
+		psArgs = []string{"-ef"}
+	}
+
+	cmd := exec.Command("ps", psArgs...)
+	output, err := cmd.CombinedOutput()
```
sboeuf: Have you tried to run this intensively ? I ask because I am afraid this command might error sometimes, due to the subreaper. Let's say you run your `ps` process from the agent, the subreaper might be the one reaping the process when it terminates, meaning the `cmd.Process.Wait()` called from `cmd.CombinedOutput()` might return an error because it didn't properly got the exit code from the command.
@laijs @bergwolf any input on that ?

protocols/grpc/agent.proto
```diff
@@ -97,6 +98,18 @@ message WaitProcessResponse {
...
+// ListProcessesRequest contains the options used to list running processes inside the container 
+message ListProcessesRequest {
+	string container_id = 1;
+	string format = 2;
+	repeated string args = 3;
+}
+
+// ListProcessesResponse represents the list of running processes inside the container
+message ListProcessesResponse {
+	bytes processList = 1;
```
bergwolf: `process_list`

grpc.go
```diff
@@ -536,6 +541,91 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
+	default:
+		return resp, fmt.Errorf("invalid format option")
+	}
+
+	psArgs := req.Args
+	if len(psArgs) == 0 {
+		psArgs = []string{"-ef"}
+	}
+
+	cmd := exec.Command("ps", psArgs...)
```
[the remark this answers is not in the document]
devimc: nop, `c.container.Processes()` returns the list of process

grpc.go
```diff
@@ -536,6 +541,91 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
+
+	switch req.Format {
+	case "table":
+	case "json":
+		resp.ProcessList, err = json.Marshal(pids)
+		return resp, err
+	default:
+		return resp, fmt.Errorf("invalid format option")
+	}
+
```
sboeuf: @devimc everything after this looks like it is not reachable. Am I missing something ?

grpc.go
devimc: you're right, fixed

grpc.go
devimc: do you mean if the format is not supported ?

devimc:
@bergwolf @sboeuf changes applied, thanks

grpc.go
sboeuf: I mean that if you look at this `switch` statement, there are `return` for each statement, including `default`. How do you expect the code further down to be reached ?

devimc: no,  `table` doesn't have a `return`

```go
......
	case "table":
	case "json":
		resp.ProcessList, err = json.Marshal(pids)
		return resp, err
.......
```

sboeuf: Oh yes I was assuming a `fallthrough` here...

jodh-intel: @sboeuf - I think the code is validating `table` as a legitimate format type (but with no associated statement). However, the actual handling for `table` is below the `switch`.

grpc.go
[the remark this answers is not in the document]
bergwolf: The `ps` command is going to run in agent's pidns, which is not the container's pidns AFAICT.

devimc:
@kata-containers/agent I'll address code coverage in this pr https://github.com/kata-containers/agent/pull/207

grpc.go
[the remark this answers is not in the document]
devimc: yes, and that's ok,  or I'm missing something.. ?

devimc:
can we merge this?, I'll fix the code coverage in #207

devimc:
rebased  :smile:

grpc.go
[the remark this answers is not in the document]
amshinde: @devimc If we want to return a list of processes running inside a container, should'nt we be running the `ps` command inside container's PID namespace.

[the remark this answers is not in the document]
devimc: @amshinde nop, `c.container.Processes()` already does that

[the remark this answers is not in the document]
bergwolf: I see. `c.container.Processes()` returns pids in the caller's namespace. So it works if the caller is in the `init_ns`, which is true for the agent. Thanks for the explanation @devimc !

bergwolf:
lgtm

grpc.go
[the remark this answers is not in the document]
amshinde: @devimc Can you add explain/ add comments why you need to run the ps command in the agent namespace when you already have all the pids from the container?
Cant you use a predefined ps table format?

grpc.go
```diff
@@ -623,6 +627,91 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
+		}
+
+		p, err := strconv.Atoi(fields[pidIndex])
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert pid to int: %s", fields[pidIndex])
+		}
+
+		// appends pid line
+		for _, pid := range pids {
+			if pid == p {
```
amshinde: You are comparing the pid from the container namespace to the one in the agent pid namespace. Can you explain why you do this? I would imagine not processes from the container pid namespace would not be visible in the agent namespace.

devimc: I added more comments in the code, like this

```go
// Get the list of processes that are running inside the containers.
// the PIDs match with the system PIDs, not with container's namespace
pids, err := c.container.Processes() 
```

devimc:
@amshinde changes applied, thanks

devimc:
almost done @kata-containers/agent  one more review please  :smiley:

sboeuf:
@grahamwhaley @egernst @jcvenegas ping ! PTAL

sboeuf:
@devimc @chavafg any idea why centos build is stuck here ?

chavafg:
I see the centos job working correctly, maybe we hit the limit of concurrent jobs and this one was waiting.

grpc.go
```diff
@@ -623,6 +627,97 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
+	psArgs := req.Args
+	if len(psArgs) == 0 {
+		psArgs = []string{"-ef"}
+	}
+
+	// All container's processes are visibles from agent's namespace.
+	// pids already contains the list of processes that are running
+	// inside a container, now we have to use that list to filter
+	// ps output and return just container's processes
+	cmd := exec.Command("ps", psArgs...)
```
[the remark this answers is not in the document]
amshinde: Agree with gwhaley. Also, we should really avoid execing if possible.

grpc.go
amshinde: The fact that you are able to match the pids from the agent pid namespace makes me think that this is returning the system pids.

devimc: sorry I don't understand your concerns , is the agent running inside a namespace? because I tested this patch and works, `ps` returns all the processes, even I see the PID 1 (systemd)

grahamwhaley: I think that is sort of the crux - when you do a 'ps' of a container, does it show you the systemd process '1' of the (for want of a better phrase..) mini-OS that is in the VM. If I am understanding correctly, a `ps` of a container should only show you the processes in that container - in that workload - we should not see anything, such as systemd pid1, from the mini-OS.

Maybe you can post an example output of what happens to for a 'busybox sh' container or similar to clarify what we get? Also, maybe, a `ps` from inside the container (`docker exec ps`) so we can compare?

grpc.go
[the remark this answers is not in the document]
devimc: see https://github.com/opencontainers/runc/blob/master/ps.go#L63

the idea behind runtime's `ps` sub-command is to list the processes using linux [ps](https://linux.die.net/man/1/ps) command,  _"peruse the /proc"_ is not an option since the user decides the output format and the information to show. Would be nice to have an implementation of [ps](https://linux.die.net/man/1/ps) in `go` to include it here, but as far as  I know there is no. If you concern is that we don't include `ps` command in the image, don't worries I already filed an [issue](https://github.com/kata-containers/osbuilder/issues/81) for that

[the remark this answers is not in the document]
grahamwhaley: Ah, so the user can supply arguments to `ps` from the docker command line? In which case, yes, it makes more sense to use `ps` directly.
I will note, you might find that some implementations of `ps` (notably busybox, which then also implies Alpine I believe), may act differently and/or not support all options that you get with the normal 'full ps'. So, you may wish to beware and test that out.

grpc.go
devimc: with runc

```
$ docker run --runtime=runc -dti debian bash
d7d8f6ed73faef744b3961efdef780a4b94642b6d396bd73afa5699c32ee7d01
$ sudo docker-runc ps d7d8f6ed73faef744b3961efdef780a4b94642b6d396bd73afa5699c32ee7d01
UID        PID  PPID  C STIME TTY          TIME CMD
root      4337  4320  0 08:31 pts/0    00:00:00 bash
```

with kata-runtime

```
$ docker run --runtime=kata-runtime -dti debian bash
32b6d8eb33c750c8ebf960464a0a40860e138d24bfd13a45ee45704dbd1462bc
$ sudo kata-runtime ps 32b6d8eb33c750c8ebf960464a0a40860e138d24bfd13a45ee45704dbd1462bc
UID         PID   PPID  C STIME TTY          TIME CMD
root        147    115  0 13:44 ?        00:00:00 bash
```

devimc:
rebased

grpc.go
devimc: docker exec with runc

```
$ docker run --runtime=runc -dti busybox sh
5e93234a965d7e64d7a2ab918b9eaace72712137f08b6a43ce61162f45752ad0
$ docker exec 5e93234a965d7e64d7a2ab918b9eaace72712137f08b6a43ce61162f45752ad0 ps
PID   USER     TIME  COMMAND
    1 root      0:00 sh
    6 root      0:00 ps
```


docker exec with kata-runtime
```
$ docker run --runtime=kata-runtime -dti busybox sh
9e1070e509f5574117e86319aff9c1c2d64a00f2d27a40b974c46f59971266c3
$ docker exec 9e1070e509f5574117e86319aff9c1c2d64a00f2d27a40b974c46f59971266c3 ps
PID   USER     TIME  COMMAND
    1 root      0:00 sh
    5 root      0:00 ps
```

grahamwhaley: Ooh. I see. That is interesting - that it seems `docker ps` does not use the PIDS from inside the container user namepace, but indeed shows the host space PIDS. Not really what I expected at all (but maybe there are sound reasons for `runc` - that maybe you can then manipulate those processes from the guest).
Things are a little different for kata, as even with that 'guest PID', that is really the VM root PID, so there is not really much you can do with that from the guest - but, showing that is probably as near to `runc` behaviour as we can get. We could debate if that is meaningful for VM containers though.
@amshinde , you agree here that this is fine then?

amshinde: Yeah, I was not expecting to see the system processes here. Looks fine to me then.

sboeuf:
@grahamwhaley I think you can merge this now (after you've ack'ed it !)

PR #196, opened 2018-03-30
agent: Rollback properly when container creation fails

### LINKED ISSUE #195: Properly rollback when container creation fails
We need to make sure the mounts performed during the container creation are completely removed in case of a failure.

sboeuf:
In case the container creation fails, we need a proper rollback
regarding the mounts previously performed.

Fixes #195

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -323,18 +323,37 @@ func (a *agentGRPC) updateContainerConfig(spec *specs.Spec, config *configs.Conf
 	return a.updateContainerConfigPrivileges(spec, config)
 }
 
-func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (*gpb.Empty, error) {
+// rollbackFailingContainerCreation rolls back important steps that might have
+// been performed before the container creation failed.
+// - Destroy the container created by libcontainer
+// - Delete the container from the agent internal map
+// - Unmount all mounts related to this container
+func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container, err error) {
+	if err != nil {
+		if ctr.container != nil {
+			ctr.container.Destroy()
+		}
+
+		a.sandbox.deleteContainer(ctr.id)
+
+		if err2 := removeMounts(ctr.mounts); err2 != nil {
+			agentLog.WithError(err2).Error("rollback failed")
+		}
+	}
+}
+
+func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (resp *gpb.Empty, err error) {
 	if a.sandbox.running == false {
 		return emptyResp, grpcStatus.Error(codes.FailedPrecondition, "Sandbox not started, impossible to run a new container")
 	}
 
-	if _, err := a.sandbox.getContainer(req.ContainerId); err == nil {
+	if _, err = a.sandbox.getContainer(req.ContainerId); err == nil {
 		return emptyResp, grpcStatus.Errorf(codes.AlreadyExists, "Container %s already exists, impossible to create", req.ContainerId)
 	}
 
 	// re-scan PCI bus
 	// looking for hidden devices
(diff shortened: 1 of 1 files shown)


sboeuf:
@egernst @bergwolf PTAL

devimc:
lgtm

sboeuf:
@laijs @bergwolf PTAL

grpc.go
```diff
@@ -359,6 +359,27 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
+	// Make sure we rollback the mounts performed in case the container
+	// creation failed.
+	defer func() {
+		if err != nil {
+			if ctr.container != nil {
+				ctr.container.Destroy()
+			}
+
+			a.sandbox.deleteContainer(req.ContainerId)
+			removeMounts(ctr.mounts)
```
jodh-intel: Shouldn't that be:

```
err = removeMounts(ctr.mounts)
```

sboeuf: Well not really since I don't think an error from the rollback part executed from `defer` should not gate anything.

jodh-intel: We're already in the error path but we should probably atleast combine the errors so they get logged at a higher level since we don't expect.

sboeuf: This looks overkill to me... We're already into the error path, meaning an error is going to be logged, but you also want the potential error of the error ? What should I log ? A `Debug()` one ?

jodh-intel: I can see that point of view and I also see that it's going to complicate the code a little, but why are we doing rollback? Answer: to ensure we *undo* the operations the agent just performed.. So we shouldn't fully ignore operations that return errors imho as that would hide bugs relating to the undo.

As a minimum, I think we should do something like:

```
err2 := removeMounts(ctr.mounts)
if err2 != nil {
    agentLog.WithError(err2).Error("rollback failed")
}
```

That way atleast we get something useful in the logs.

btw, the naming is a little random - it's not very obvious at first glance that the `defer` basically undoes `addStorages()`. Maybe the code block could be put into a `removeStorages()` to make that clearer?

sboeuf: Yeah makes sense, let me rework a little bit

sboeuf:
@jodh-intel PR updated.

jodh-intel:
lgtm

grpc.go
```diff
@@ -359,6 +378,18 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
 
+	ctr := &container{
+		id:        req.ContainerId,
+		processes: make(map[string]*process),
+		mounts:    mountList,
+	}
+
+	a.sandbox.setContainer(req.ContainerId, ctr)
+
+	// Make sure we rollback the mounts performed in case the container
```
grahamwhaley: it's not just the mounts - it is 'anything that needs undoing' (and we do a touch more cleanup there I think).

grpc.go
```diff
@@ -388,32 +419,22 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
-	}
-
-	a.sandbox.setContainer(req.ContainerId, container)
-
-	if err := a.execProcess(container, container.initProcess, true); err != nil {
+	if err = a.execProcess(ctr, ctr.initProcess, true); err != nil {
 		return emptyResp, err
 	}
 
-	return emptyResp, a.postExecProcess(container, container.initProcess)
```
grahamwhaley: doing this little rename cleanup in a separate commit would have made review a touch easier ;-)

grpc.go
sboeuf: let me change this

sboeuf: done

jodh-intel:
lgtm.

I guess we'll need to test this via https://github.com/kata-containers/tests somehow.

sboeuf:
Can we merge this even with this codecov red cross ? I am not sure how I can add unit testing for this.

jodh-intel:
I say yes as we haven't checked the results coming back from codecov yet (nor have we looked at thresholds).

PR #197, opened 2018-04-02
mount: Correct error message with mount failure.

amshinde:
The function mount() performs a generic mount and calls into
mount(2). The current error message is misleading as it refers
to an error in bind-mounting.


devimc:
lgtm

jodh-intel:
lgtm

Issue #209, opened 2018-04-09
network: failure on unsupported netlink protocol


bergwolf:
The go netlink package has added support for more netlink protocols than we need. We need to specify NETLINK_ROUTE when grabbing new netlink handle to avoid failure due to unsupported netlink protocols.

PR #210, opened 2018-04-09
network: specify netlink protocol

bergwolf:
The go netlink package has added more netlink socket protocols
than we need. Specify one to avoid failure due to unsupported
protocols.

Fixes: #209

### DIFF
--- a/network.go
+++ b/network.go
@@ -12,6 +12,8 @@ import (
 	"reflect"
 	"sync"
 
+	"golang.org/x/sys/unix"
+
 	pb "github.com/kata-containers/agent/protocols/grpc"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
@@ -101,7 +103,7 @@ func (s *sandbox) addInterface(netHandle *netlink.Handle, iface *pb.Interface) (
 	defer s.network.ifacesLock.Unlock()
 
 	if netHandle == nil {
-		netHandle, err = netlink.NewHandle()
+		netHandle, err = netlink.NewHandle(unix.NETLINK_ROUTE)
 		if err != nil {
 			return nil, err
 		}
@@ -146,7 +148,7 @@ func (s *sandbox) removeInterface(netHandle *netlink.Handle, iface *pb.Interface
 	defer s.network.ifacesLock.Unlock()
 
 	if netHandle == nil {
-		netHandle, err = netlink.NewHandle()
+		netHandle, err = netlink.NewHandle(unix.NETLINK_ROUTE)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


nitkon:
LGTM. I tested it out on my x86 machine and the kata container is getting created with the hyperstart kernel using the initrd approach. 

sudo docker run -ti --runtime kata-runtime busybox sh
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
f70adabe43c0: Already exists 
Digest: sha256:58ac43b2cc92c687a32c8be6278e50a063579655fe3090125dcb2af0ff9e1a64
Status: Downloaded newer image for busybox:latest
/ # ls 
bin   dev   etc   home  proc  root  sys   tmp   usr   var

devimc:
@bergwolf  does this change is to remove XFRM support from the kernel?

bergwolf:
@devimc actually it was `NETLINK_NETFILTER` that caused the docker failure.

sboeuf:
@grahamwhaley PTAL.

devimc:
@bergwolf  ok, then IP_SET is not needed, right?

sboeuf:
@devimc could you give more context about what you're talking about ? I am asking because I don't understand the relation between your comments and this change.

sboeuf:
@devimc thanks for the explanation here. So because this change makes only use of `NETLINK_ROUTE`, you're trying to reduce our Kata Containers kernel.

devimc:
@sboeuf  yes

sboeuf:
@devimc Ok then I think we could open an issue for that, as I expect that we will merge this PR very soon. This way we can continue the discussion.

devimc:
@sboeuf  https://github.com/kata-containers/linux/pull/5 has not been merged, so I'll update the PR

bergwolf:
@devimc We are apparently not depending on `iptables` for setting up the guest filtering rules. So it looks like `NETLINK_NETFILTER` (and `IP_SET` that depends on it) is not required at the moment. But will we depend on `iptables` at some point in the future?

devimc:
@bergwolf good question, @mcastelino  @amshinde WDYT ?

mcastelino:
@devimc we can wait till we see the need for NETLINK_NETFILTER. 

@bergwolf have you run into a situation where you have set iptable rules inside the VM. I noticed in your runv/frakti code do had a option to setup port forwarding rules inside the VM? When did you use this?

amshinde:
@mcastelino We may end up needing iptables for supporting docker compose in the future.

bergwolf:
@mcastelino in runv, we use iptables to setup port mapping and to support docker CNM.

@all whether we need NETLINK_XFRM and NETLINK_NETFILTER is not relevant to this patch. Can we go ahead and merge it and have another issue to discuss kernel config options?

sboeuf:
I am all for merging this. @egernst @grahamwhaley PTAL.
And let's continue the discussion on a separate issue or PR (here kata-containers/linux#5 as indicated by @devimc).

bergwolf:
agreed @grahamwhaley. Let's merge ignoring the codecov failure (which actually reports `increase coverage by 0.67%.`).

PR #211, opened 2018-04-09
agent: add server interceptor to log grpc requests

bergwolf:
When agent_log is set to `debug` level, trace all grpc requests and
responses.

Fixes: #183

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -22,11 +22,13 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
 	pb "github.com/kata-containers/agent/protocols/grpc"
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	_ "github.com/opencontainers/runc/libcontainer/nsenter"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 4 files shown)


agent.go
```diff
@@ -440,6 +443,26 @@ func (s *sandbox) initChannel() error {
 	return s.channel.setup()
 }
 
+func grpcTracer(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
```
sboeuf: Please let the code breathes a little bit into this function, otherwise the code is harder to read.

bergwolf:
Jenkins failed on kata-containers/tests/issues/228.

agent.go
```diff
@@ -440,6 +443,29 @@ func (s *sandbox) initChannel() error {
...
+	message := req.(proto.Message)
+	agentLog.WithFields(logrus.Fields{
+		"request": info.FullMethod,
+		"req":     message.String()}).Debug("new request")
```
devimc: I wonder if changing log format will make log-parser happy

```go
	"request": info.FullMethod,
	"req":     "new request"}).Debug(message.String())
```

bergwolf: I don't think so. The problem is that kernel prints one line to the console while agent is printing the logrus format log and they mixed up. It is not even related to this patch at all. The situation can happen at any time. The reason we only see it until now is because it introduces a lot of agent logs while there are not much agent logs at all even if debug is on before this PR.

jodh-intel:
Very nice!

lgtm

Issue #213, opened 2018-04-11
support serial port reconnection


bergwolf:
When using a serial port, we rely on yamux to do multiplexing but there is one drawback of yamux -- it only allow for one underlying connection. So when kata runtime drops the serial port connection, either due to restart or due to some errors, the agent grpc service will quit and the runtime cannot reconnect anymore. This also blocks the implementation of `Sandbox.Release()` API.

PR #214, opened 2018-04-11
support agent grpc reconnection over serial port

bergwolf:
For yamux over serial connections, there can be only one serial connection, and when kata proxy quits, such connection is closed and thus the agent grpc server quits.

To support kata proxy reconnection, we need to listen and accept on the serial file again after existing grpc server quits.

Fixes: #213


channel.go
```diff
@@ -86,6 +93,59 @@ func (c *serialChannel) setup() error {
...
+	}
+	defer syscall.Close(epfd)
+
+	event.Events = syscall.EPOLLOUT | syscall.EPOLLET&0xffffffff
+	event.Fd = int32(fd)
+	if err = syscall.EpollCtl(epfd, syscall.EPOLL_CTL_ADD, int(fd), &event); err != nil {
+		return err
+	}
+	defer func() {
+		syscall.EpollCtl(epfd, syscall.EPOLL_CTL_DEL, int(fd), nil)
```
devimc: Q. why not just ```defer syscall.EpollCtl(epfd, syscall.EPOLL_CTL_DEL, int(fd), nil)```

channel.go
```diff
@@ -86,6 +93,59 @@ func (c *serialChannel) setup() error {
...
+		if err != nil {
+			return err
+		}
+
+		for i := 0; i < nev; i++ {
+			ev := events[i]
+			if ev.Fd == int32(fd) {
+				agentLog.WithFields("events", ev.Events).Debug("New serial channel event")
+				// Do not handle HUP if there is data available
+				if ev.Events&syscall.EPOLLOUT != 0 {
```
devimc: commend is `// Do not handle HUP if there is data available` but this is `EPOLLOUT` ?

channel.go
```diff
@@ -86,6 +93,59 @@ func (c *serialChannel) setup() error {
...
+					return nil
+				}
+				// Peer point closed before data is avaialbe
+				if ev.Events&syscall.EPOLLERR != 0 {
+					return fmt.Errorf("serial port IO failure")
+				}
+				if ev.Events&syscall.EPOLLHUP != 0 {
```
devimc: `EPOLLOUT` or `EPOLLHUP` ?

channel.go
```diff
@@ -86,6 +93,59 @@ func (c *serialChannel) setup() error {
...
+		return fmt.Errorf("serial port IO closed")
+	}
+
+	epfd, err := syscall.EpollCreate1(syscall.EPOLL_CLOEXEC)
+	if err != nil {
+		return err
+	}
+	defer syscall.Close(epfd)
+
+	event.Events = syscall.EPOLLOUT | syscall.EPOLLET&0xffffffff
```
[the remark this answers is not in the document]
bergwolf: Sure. `syscall.EPOLLET` is negative and does not fit in golang `uint32`. I'll add comments about it.

channel.go
bergwolf: Good point!

channel.go
bergwolf: Sorry, comment outdated. I'll update.

channel.go
bergwolf: `EPOLLOUT` is already handled above.

agent.go
```diff
@@ -437,17 +438,10 @@ func (s *sandbox) initChannel() error {
 
 	s.channel = c
 
-	return s.channel.setup()
+	return nil
 }
 
 func (s *sandbox) startGRPC() error {
```
devimc: with this patch now `startGRPC` always returns nil, wouldn't be better return nothing

bergwolf: good point!

agent.go
```diff
@@ -461,18 +455,47 @@ func (s *sandbox) startGRPC() error {
...
 
-	return s.channel.teardown()
+			var l net.Listener
+			l, err = s.channel.listen()
+			if err != nil {
+				agentLog.WithError(err).Warn("Failed to create agent grpc listener")
+				return
+			}
+
+			s.grpcListener = l
```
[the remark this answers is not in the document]
bergwolf: Yeah, it seems we can just remove `grpcListener`.

bergwolf:
@devimc updated. PTAL.

bergwolf:
@devimc are you OK with the PR now?

bergwolf:
@kata-containers/agent CI green. PTAL.

PR #217, opened 2018-04-20
uevent: Fix netlink error while assigning pid in netlink client

### LINKED ISSUE #216: Make sure that it is possible to open two netlink sockets to listen to udev events.
With our current implementation of uevent package, if two netlink sockets are opened at the same time for listening to uevents, we get a netlink error that the address is in use. This is because we use the Pid of the process in assigning the socket address, which results in conflicts.
https://github.com/kata-containers/agent/blob/06ce85905e514591dc9036ae1b7e32acd2f48022/pkg/uevent/uevent.go

 Instead allow kernel to choose socket address.

amshinde:
unix.SockaddrNetlink represents a netlink client with the Pid
being the netlink socket address. This can be assigned the actual
pid of the process, but in case we have two netlink sockets opened
at the same time, this results in errors in binding to the netlink
socket. Assign this to zero which means the kernel takes care of
assigning it.

See http://man7.org/linux/man-pages/man7/netlink.7.html

Fixes #216

### DIFF
--- a/pkg/uevent/uevent.go
+++ b/pkg/uevent/uevent.go
@@ -9,7 +9,6 @@ package uevent
 import (
 	"bufio"
 	"io"
-	"os"
 	"strings"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


pkg/uevent/uevent.go
```diff
@@ -35,7 +34,7 @@ type ReaderCloser struct {
 func NewReaderCloser() (io.ReadCloser, error) {
 	nl := unix.SockaddrNetlink{
 		Family: unix.AF_NETLINK,
-		Pid:    uint32(os.Getpid()),
+		Pid:    uint32(0),
```
jodh-intel: I appreciate you've documented this in the commit, but I'd also be tempted to put a comment above this line as it looks "odd" ;)

jodh-intel:
lgtm

bergwolf:
lgtm!

PR #220, opened 2018-04-23
grpc: update cpuset cgroup

### LINKED ISSUE #219: not all vcpus are available in the containers
# Description of problem

```
docker run --cpus 5 -ti busybox sh -c "sleep 5; nproc"
```
don't show the right number of vCPUs

> now that our kernel supports cpuSet libcontainer places the container inside a CPU cgroup with the number of vCPUs available when it is created. VCPUs are hotplugged and connected asynchronously, not all vCPUs are available when the container is created causing this issue1

# Expected result

`6`   1 cold added + 5 hot added

# Actual result

random, sometimes 3 or 4

see https://github.com/kata-containers/tests/issues/250

devimc:
Each time a container is created, it is placed inside a cpuset cgroup
with the actual number of vCPUs, when more vCPUs are hot added this cgroup
MUST BE updated to allow containers accessing to these new resources. The
cpuset cgroup is not updated if the container already have one defined by user.

fixes #219

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -71,6 +71,15 @@ const onlineCPUMemWaitTime = 100 * time.Millisecond
 
 const onlineCPUMaxTries = 10
 
+// handleError will log the specified error if wait is false
+func handleError(wait bool, err error) error {
+	if !wait {
+		agentLog.WithError(err).Error()
+	}
+
+	return err
+}
+
 // Online resources, nbResources specifies the maximum number of resources to online.
 // If nbResources is <= 0 then there is no limit and all resources are connected.
 // Returns the number of resources connected.
--- a/grpc.go
+++ b/grpc.go
@@ -54,8 +54,10 @@ const (
 )
 
 var (
-	sysfsCPUOnlinePath = "/sys/devices/system/cpu"
-	sysfsMemOnlinePath = "/sys/devices/system/memory"
+	sysfsCPUOnlinePath     = "/sys/devices/system/cpu"
+	sysfsMemOnlinePath     = "/sys/devices/system/memory"
+	sysfsConnectedCPUsPath = filepath.Join(sysfsCPUOnlinePath, "online")
+	sysfsDockerCpusetPath  = "/sys/fs/cgroup/cpuset/docker/cpuset.cpus"
 )
 
 type onlineResource struct {
@@ -79,6 +81,8 @@ func handleError(wait bool, err error) error {
 
 	return err
 }
+
+const dockerCpusetMode = 0644
 
 // Online resources, nbResources specifies the maximum number of resources to online.
 // If nbResources is <= 0 then there is no limit and all resources are connected.
@@ -154,19 +158,55 @@ func onlineMemResources() error {
 	return err
 }
 
-func onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
+func (a *agentGRPC) onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
 	if req.NbCpus <= 0 {
-		return fmt.Errorf("requested number of CPUs '%d' must be greater than 0", req.NbCpus)
+		return handleError(req.Wait, fmt.Errorf("requested number of CPUs '%d' must be greater than 0", req.NbCpus))
 	}
 
 	onlineCPUMemLock.Lock()
 	defer onlineCPUMemLock.Unlock()
 
 	if err := onlineCPUResources(req.NbCpus); err != nil {
-		return err
+		return handleError(req.Wait, err)
 	}
 
-	return onlineMemResources()
+	if err := onlineMemResources(); err != nil {
+		return handleError(req.Wait, err)
+	}
+
+	// At this point all CPUs have been connected, we need to know
+	// the actual range of CPUs
+	cpus, err := ioutil.ReadFile(sysfsConnectedCPUsPath)
+	if err != nil {
+		return handleError(req.Wait, fmt.Errorf("Could not get the actual range of connected CPUs: %v", err))
+	}
+	connectedCpus := strings.Trim(string(cpus), "\t\n ")
+
+	// In order to update container's cpuset cgroups, docker's cpuset cgroup MUST BE updated with
+	// the actual number of connected CPUs
+	if err := ioutil.WriteFile(sysfsDockerCpusetPath, []byte(connectedCpus), dockerCpusetMode); err != nil {
+		return handleError(req.Wait, fmt.Errorf("Could not update docker cpuset cgroup '%s': %v", connectedCpus, err))
+	}
+
+	// Now that we know the actual range of connected CPUs, we need to iterate over
+	// all containers an update each cpuset cgroup. This is not required in docker
+	// containers since they don't hot add/remove CPUs.
+	for _, c := range a.sandbox.containers {
+		contConfig := c.container.Config()
+		// Don't update cpuset cgroup if one was already defined.
+		if contConfig.Cgroups.Resources.CpusetCpus != "" {
+			continue
+		}
+		contConfig.Cgroups.Resources = &configs.Resources{
+			CpusetCpus: connectedCpus,
+		}
+		if err := c.container.Set(contConfig); err != nil {
+			return handleError(req.Wait, err)
+		}
+
+	}
+
+	return nil
 }
 
 func setConsoleCarriageReturn(fd int) error {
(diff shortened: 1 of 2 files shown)


grpc.go
```diff
@@ -157,7 +161,41 @@ func onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
...
+	connectedCpus := strings.Trim(string(cpus), "\t\n ")
+
+	// In order to update container's cpuset cgroups, docker's cpuset cgroup MUST BE updated with
+	// the actual number of connected CPUs
+	if err := ioutil.WriteFile(sysfsDockerCpusetPath, []byte(connectedCpus), dockerCpusetMode); err != nil {
+		return fmt.Errorf("Could not update docker cpuset cgroup '%s': %v", connectedCpus, err)
+	}
+
+	// Now that we know the actual range of connected CPUs, we need to iterate over
+	// all containers an update each cpuset cgroup, this is not required in docker
```
jodh-intel: Nit: "and update each cpuset group. This is not ...".

grpc.go
```diff
@@ -935,9 +973,9 @@ func (a *agentGRPC) UpdateRoutes(ctx context.Context, req *pb.UpdateRoutesReques
 
 func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemRequest) (*gpb.Empty, error) {
 	if !req.Wait {
-		go onlineCPUMem(req)
+		go a.onlineCPUMem(req)
```
jodh-intel: This isn't strictly related to this PR (since the code already uses a go routine), but if `Wait` is `true`, fwics the agent will never log any errors from `onlineCPUMem()`.

But since `req` is passed to that function, it could log all error scenarios so atleast something is captured to the journal. It would make the code a bit less clean, but something like:

```
// handleError will log the specified error if wait is true
func handleError(wait bool, err error) error {
    if wait {
        agentLog.WithError(err).Error()
    }

    return err
}

func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemRequest) (*gpb.Empty, error) {

    :

    err := doSomething()
    if err != nil {
        return handleError(req.Wait, err)
    }

    :
}
```

devimc:
@bergwolf  for docker that could be a solution, but for k8s I don't think so, because new CPUs can be hot added each time a new container is created inside the POD and  these resources should be visible for all containers of course if the container does not have a cpuset constraint.

bergwolf:
@devimc I agree we should fix it for adding new containers to existing sandbox. But we should also make sure a container is created **after** all of its resources are hot plugged. It's more of a container creation workflow issue. Do you agree?

bergwolf:
@devimc hmm, that is indeed a good point. I agree we should do these operation in parallel, especially vCPU and memory. For network and storage we should wait but that is not relevant to this PR. Thanks for point it out!

bergwolf:
lgtm!

grpc.go
```diff
@@ -145,19 +158,55 @@ func onlineMemResources() error {
...
+	// the actual number of connected CPUs
+	if err := ioutil.WriteFile(sysfsDockerCpusetPath, []byte(connectedCpus), dockerCpusetMode); err != nil {
+		return handleError(req.Wait, fmt.Errorf("Could not update docker cpuset cgroup '%s': %v", connectedCpus, err))
 	}
 
-	return onlineMemResources()
+	// Now that we know the actual range of connected CPUs, we need to iterate over
+	// all containers an update each cpuset cgroup. This is not required in docker
+	// containers since they don't hot add/remove CPUs.
+	for _, c := range a.sandbox.containers {
```
jcvenegas: why should be updated on each container in the sandbox, the cpu update is not an operation per container ? In that case we should receive the container that will be updated, isn't it?

grpc.go
```diff
@@ -54,8 +54,10 @@ const (
 )
 
 var (
-	sysfsCPUOnlinePath = "/sys/devices/system/cpu"
-	sysfsMemOnlinePath = "/sys/devices/system/memory"
+	sysfsCPUOnlinePath     = "/sys/devices/system/cpu"
+	sysfsMemOnlinePath     = "/sys/devices/system/memory"
+	sysfsConnectedCPUsPath = filepath.Join(sysfsCPUOnlinePath, "online")
+	sysfsDockerCpusetPath  = "/sys/fs/cgroup/cpuset/docker/cpuset.cpus"
```
jcvenegas: dump question why 'docker'?

devimc: because we are using `libcontainer` and by default it creates the cgroup with that name

jcvenegas: would be nice to have it in a comment, but thanks for clarify.

grpc.go
devimc: nop, in the case of a POD we can have several containers, when new CPUs are added, these new resources need to be visible for all containers

jcvenegas: thanks.

devimc:
@bergwolf can we merge this pr ?

devimc:
@kata-containers/agent  please take a look :smile:

egernst:
@devimc  -- will take care of this once jenkins finishes up.  Thanks!

grpc.go
sboeuf: @devimc if libcontainer creates this, is there any way we could retrieve this from libcontainer itself instead of hardcoding it ?

grpc.go
```diff
@@ -71,6 +73,17 @@ const onlineCPUMemWaitTime = 100 * time.Millisecond
 
 const onlineCPUMaxTries = 10
 
+const dockerCpusetMode = 0644
+
+// handleError will log the specified error if wait is true
+func handleError(wait bool, err error) error {
```
sboeuf: This is different than what is described in the commit message. Could you please split this PR into two separate commits ?

grpc.go
devimc: I didn't find a way to get this from `libcontainer`

devimc:

grpc.go
sboeuf: @devimc Thanks for splitting into two commits, but two things:
- You should provide an issue ID through `Fixes #XXX` in the new commit message
- Please add, as part of this commit, all the code that has been changed with the introduction of this new function `handleError()`. This way, it's easier to review the commit related to `cpuset` only.

grahamwhaley:
CI 16.04 seems to be 'stuck' in some sort of limbo state. CI 17.10 failed with this:
```
# time="2018-04-30 22:47:00.171433657Z" level=debug msg="ExecRequest &ExecRequest{ContainerId:1ca7cfabdaed87476e69940a45ed1d89afc38bf914111744a5e6026782174b8d,Cmd:[env],Tty:false,Stdin:false,Stdout:true,Stderr:true,}" 
# time="2018-04-30T22:47:00Z" level=fatal msg="execing command in container failed: unable to upgrade connection: 404 page not found"
# 1
# rm: cannot remove '/tmp/tmp.VlKm2Ti41e/crio/devicemapper/mnt/78d2456bb8577f2a910251abab747ccae9363e3152bbafeeb2a593f3c935f765': Device or resource busy
# rm: cannot remove '/tmp/tmp.VlKm2Ti41e/crio/devicemapper/mnt/9c92005fc289b5686ed1f2c41e17fdb28da3aa44e5fd0b1ec40c543cacc3e18b': Device or resource busy
# rm: cannot remove '/tmp/tmp.VlKm2Ti41e/crio-run/devicemapper-containers/093a74994c9794ddf86eec81a4a652591f41e41c4c5e191eb31f8e58bb0eec05/userdata/shm': Device or resource busy
Build timed out (after 5 minutes). Marking the build as aborted.
Build was aborted
```

@chavafg - is that a failure mode we know of?
I'll re-kick 17.10, and see if I can get 16.04 to do anything useful...

grahamwhaley:
Odd, 16.04 looks like it passed some time back:
http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-agent-ubuntu-16-04-PR/102/
I've given it a rebuild to see if we can force it to update the status here (and take into account the new updates).

grahamwhaley:
16.04 CI seems happy enough - but the 'feedback link' back to this Issue seems to be bust, so is not updating - so, we will need to force merger past that once we get another ack I think:
http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-agent-ubuntu-16-04-PR/122/console

@fuentess - I do see this in the logs - is that something we need to worry about?:
```
Archiving artifacts
WARN: No artifacts found that match the file pattern "kata-proxy_*,kata-runtime_*,kata-shim_*,crio_*,docker_*,kubelet_*,kata-collect-data_*". Configuration error?
WARN: ‘kata-proxy_*’ doesn’t match anything
```

chavafg:
@grahamwhaley no need to worry about that output, as the logs are collected when there is a failure. As this finished correctly, then we don't have any artifact to archive.

chavafg:
@grahamwhaley about the other failure that occurred before, we already have an issue: https://github.com/kata-containers/tests/issues/238, we need to investigate why this happens, but seems like a network issue on azure vms.

Issue #221, opened 2018-04-24
API:Add the StatsContainer api for `events cli`


jshachm:
As [runtime#188,](https://github.com/kata-containers/runtime/issues/186) we need to implement `events cli` , first thing is to add `StatsContainer` api.

PR #222, opened 2018-04-24
grpc :Add the StatsContainer api for `events cli`

jshachm:
Add `StatsContainer` API to display container events
Fixes: #221

Signed-off-by: c00416947 <caihaomin@huawei.com>

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -817,6 +817,35 @@ func (a *agentGRPC) UpdateContainer(ctx context.Context, req *pb.UpdateContainer
 	return emptyResp, c.container.Set(config)
 }
 
+func (a *agentGRPC) StatsContainer(ctx context.Context, req *pb.StatsContainerRequest) (*pb.StatsContainerResponse, error) {
+	c, err := a.sandbox.getContainer(req.ContainerId)
+	if err != nil {
+		return nil, err
+	}
+
+	stats, err := c.container.Stats()
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := json.Marshal(stats.CgroupStats)
+	if err != nil {
+		return nil, err
+	}
+
+	var cgroupStats pb.CgroupStats
+	err = json.Unmarshal(data, &cgroupStats)
+	if err != nil {
+		return nil, err
+	}
+	resp := &pb.StatsContainerResponse{
+		CgroupStats: &cgroupStats,
+	}
+
+	return resp, nil
+
+}
+
 func (a *agentGRPC) RemoveContainer(ctx context.Context, req *pb.RemoveContainerRequest) (*gpb.Empty, error) {
 	ctr, err := a.sandbox.getContainer(req.ContainerId)
 	if err != nil {
--- a/protocols/grpc/agent.proto
+++ b/protocols/grpc/agent.proto
@@ -30,6 +30,7 @@ service AgentService {
 	rpc WaitProcess(WaitProcessRequest) returns (WaitProcessResponse); // wait & reap like waitpid(2)
 	rpc ListProcesses(ListProcessesRequest) returns (ListProcessesResponse);
 	rpc UpdateContainer(UpdateContainerRequest) returns (google.protobuf.Empty);
+	rpc StatsContainer(StatsContainerRequest) returns (StatsContainerResponse);
 
 	// stdio
 	rpc WriteStdin(WriteStreamRequest) returns (WriteStreamResponse);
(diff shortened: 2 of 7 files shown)


sboeuf:
@jshachm I'll try to review this when I get a chance. In the meantime, could you please fix the CI.

protocols/grpc/health.pb.go
```diff
@@ -108,7 +110,10 @@ func init() {
 }
 func (this *CheckRequest) Equal(that interface{}) bool {
 	if that == nil {
-		return this == nil
+		if this == nil {
```
bergwolf: The check is useless yet it is auto-generated. Are you using an old version of protoc?

protocols/grpc/agent.proto
```diff
@@ -110,6 +111,87 @@ message ListProcessesResponse {
 	bytes process_list = 1;
 }
 
+message StatsContainerRequest {
+    string container_id = 1;
+}
+
+message CpuUsage {
```
bergwolf: Do we have a reference about why these stats are collected? IOW, are they sufficient to support the CRI metrics as well as the kata cli events?

protocols/grpc/health.pb.go
jshachm: I am try to using the latest `protoc` to re-generated it :-)

jshachm:
@sboeuf @bergwolf  which  version of `protoc` can we use to re-generate `pb`. I have tried with latest release and latest `gogo-protoc` still have problems with the different codes result.

Should we provide a stable version of `protoc` tools to avoid this problem?

bergwolf:
@jshachm Did you use the helper script `hack/update-generated-agent-proto.sh` to re-generate grpc code?

jshachm:
@bergwolf  I use this `hack/update-generated-agent-proto.sh`  and using latest `protoc`  src make binary from `master` branch and latest `gogo-protoc` from `master` branch .Still have problems .

```
$ protoc --version
libprotoc 3.5.1
```

jshachm:
@bergwolf That's weird. My local compile env is the same as yours. Maybe related to `golang` version, I myself using `1.10.1`.  

Let me find out where the problem and will upgrade the pr.

jshachm:
@sboeuf  CI is happy now

grpc.go
```diff
@@ -718,6 +718,34 @@ func (a *agentGRPC) ListProcesses(ctx context.Context, req *pb.ListProcessesRequ
 	return resp, nil
 }
 
+func (a *agentGRPC) StatsContainer(ctx context.Context, req *pb.StatsContainerRequest) (*pb.StatsContainerResponse, error) {
```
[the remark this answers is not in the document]
jshachm: Back from holiday, WIP. Thx for reviewing.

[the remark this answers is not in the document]
jshachm: @sboeuf   If we use `libcontainer.Stats` directly  as the return of a `StatsContainerResponse` or sth else like the grpc response. There will be error if we don't convert `libcontainer.Stats` into `StatsContainerResponse`.

Besides, if we want to use `libcontainer.Stats` directly, we need to convert `GO struct` into `protoc IDL`. And I think this is why here I introduce a struct `reponse` here and what's my work have done.

Glad to hear your views and suggestions. If anything I ignore, plz let me know.

WeiZhang555:
There're some conflicts now, sorry for you. Please rebase first.

WeiZhang555:
And you can try to run `make proto` to generate the pb files, it makes sure everyone is using same protoc version and there's no unnecessary modifications will be introduced due to different protoc version.

jshachm:
@WeiZhang555 done~

jshachm:
CI fails randomly. But I think it's not this patch's problem.
@sboeuf @bergwolf @WeiZhang555  PTAL 
Any other suggestions? So I can go forward to `runtime cli` implement.

protocols/grpc/agent.proto
```diff
@@ -116,6 +117,87 @@ message UpdateContainerRequest {
 	LinuxResources resources = 2;
 }
 
+message StatsContainerRequest {
+    string container_id = 1;
+}
+
+message CpuUsage {
+	uint64 total_usage = 1;
+	repeated uint64 percpu_usage = 2;
```
jodh-intel: Is CPU hotplug supported here? If it is in principle, I think we'll need some tests to ensure we dtrt.

/cc @devimc

protocols/grpc/agent.proto
```diff
@@ -116,6 +117,87 @@ message UpdateContainerRequest {
 	LinuxResources resources = 2;
 }
 
+message StatsContainerRequest {
+    string container_id = 1;
+}
+
+message CpuUsage {
+	uint64 total_usage = 1;
```
jodh-intel: Presumably this is in bytes? Same question for lots of the other fields.

protocols/grpc/agent.proto
```diff
@@ -116,6 +117,87 @@ message UpdateContainerRequest {
...
+	uint64 max_usage = 2;
+	uint64 failcnt = 3;
+}
+
+message CgroupStats {
+    CpuStats cpu_stats = 1;
+    MemoryStats memory_stats  = 2;
+    PidsStats pids_stats = 3;
+    BlkioStats blkio_stats = 4;
+    map<string, HugetlbStats> hugetlb_stats = 5; // the map is in the format "size of hugepage: stats of the hugepage"
```
jodh-intel: Why is the key of type `string` if it's being used to store "size of hugepage"? Again, is this in bytes?

protocols/grpc/agent.proto
```diff
@@ -116,6 +117,87 @@ message UpdateContainerRequest {
...
+
+}
+
+message StatsContainerResponse {
+    CgroupStats cgroup_stats = 1;
```
jodh-intel: Something to consider: what if the guest kernel doesn't provide all the cgroup values the agent is expecting to be able to read for the workload? Do we fail, or return some sort of "default" value when we couldn't query it. In fact, maybe we could consider adding a special `--test` option to the agent where it will check to see if all the cgroup paths exists as expected and error if they don't exist (and maybe even display the appropriate `CONFIG_*` name(s) to help the user configure their guest kernel appropriately).

protocols/grpc/agent.proto
devimc: libcontainer checks the cgroups and each time a new CPU is added, the cgroup is updated, in theory this should work

protocols/grpc/agent.proto
jshachm: checkout from libcontainer/cgroups/stats.go
To make it easier to transport into `kata event` display value, it should be in the same type with what `runc` does.

protocols/grpc/agent.proto
jshachm: `string` is the type of the hugetlb size like `2MB` or `1GB`  it should be in `string` as what `runc` has done.

WeiZhang555:
LGTM

@sboeuf Do you need to take another look at this？

jodh-intel:
Thanks for explaining @jshachm.

lgtm

/cc @egernst as @sboeuf is afk.

PR #225, opened 2018-04-26
agent: Simplify tearing down the network

### LINKED ISSUE #226: agent: network: Routes and Interfaces list are unproperly modified when parsing them
We need to make sure we don't modify a list that is being looped over by the caller. That's what happened with network.go when the list of routes and interfaces are removed by `removeNetwork()`.

### LINKED ISSUE #230: network: Don't remove the network routes when destroying the sandbox
The network routes will always be removed when the associated interfaces are going to be removed. It is safer to leave this to each interface removal.

sboeuf:
This PR addresses the issues related to the way the agent currently tears down the network.
First it fixes an issue due to a modification of a list containing pointers instead of copies, while at the same time, this list is being parsed.

Then, with the idea of simplifying the way the network is removed, it introduces a commit removing the network removal code that was called when the sandbox was destroyed. Indeed, such a case means the VM is going to be shut down, leading to the removal of the network by the VM anyway.

Fixes #226
Fixes #230

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -853,6 +853,7 @@ func (a *agentGRPC) CreateSandbox(ctx context.Context, req *pb.CreateSandboxRequ
 
 	a.sandbox.id = req.Hostname
 	a.sandbox.containers = make(map[string]*container)
+	a.sandbox.network.ifaces = make(map[string]*pb.Interface)
 	a.sandbox.network.dns = req.Dns
 	a.sandbox.running = true
 
--- a/network.go
+++ b/network.go
@@ -25,10 +25,10 @@ import (
 // related information.
 type network struct {
 	ifacesLock sync.Mutex
-	ifaces     []*pb.Interface
+	ifaces     map[string]*pb.Interface
 
 	routesLock sync.Mutex
-	routes     []*pb.Route
+	routes     []pb.Route
 
 	dns []string
 }
@@ -139,7 +139,7 @@ func (s *sandbox) addInterface(netHandle *netlink.Handle, iface *pb.Interface) (
 	}
 
 	// Update sandbox interface list.
-	s.network.ifaces = append(s.network.ifaces, iface)
+	s.network.ifaces[iface.Name] = iface
 
 	return iface, nil
 }
@@ -172,12 +172,7 @@ func (s *sandbox) removeInterface(netHandle *netlink.Handle, iface *pb.Interface
 	}
 
 	// Update sandbox interface list.
-	for idx, sIface := range s.network.ifaces {
-		if sIface.Name == iface.Name {
-			s.network.ifaces = append(s.network.ifaces[:idx], s.network.ifaces[idx+1:]...)
-			break
-		}
-	}
+	delete(s.network.ifaces, iface.Name)
 
 	return nil, nil
 }
@@ -465,7 +460,7 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *pb.Route, add bo
 		}
 
 		// Add route to sandbox route list.
-		s.network.routes = append(s.network.routes, route)
+		s.network.routes = append(s.network.routes, *route)
 	} else {
 		if err := netHandle.RouteDel(netRoute); err != nil {
 			return grpcStatus.Errorf(codes.Internal, "Could not remove route dest(%s)/gw(%s)/dev(%s): %v",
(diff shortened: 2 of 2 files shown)


network.go
```diff
@@ -508,15 +508,29 @@ func (s *sandbox) removeNetwork() error {
 	}
 	defer netHandle.Delete()
 
+	// Create a duplicate of the route list, with no pointers,
+	// otherwise this leads to a mix when routes are removed.
+	var routeList []pb.Route
 	for _, route := range s.network.routes {
-		if err := s.removeRoute(netHandle, route); err != nil {
+		routeList = append(routeList, *route)
```
amshinde: @sboeuf Where else in the code are we updating this list? 
Why not just store objects instead of pointers in the list? We can always modify the objects itself when a slice is passed to a function.

sboeuf: This involves a bigger change regarding how things are handled if we move from pointer to object. I am fine with any solution but I want some feedback before to go ahead and implement it.
@bergwolf @jodh-intel @egernst @grahamwhaley PTAL

sboeuf:
@egernst @amshinde I have simplified the network removal, PTAL.

sboeuf:
@bergwolf @grahamwhaley PTAL

network.go
bergwolf: This is because a slice is not really a good data structure for dynamical element removal, - you cannot avoid data copy if you want to remove within a loop.

OTOH, if we add an ID to pb.Route to uniquely identify a route for a sandbox, we can use a map for it and it is perfectly safe to do
```
for _, r := range routes {
     delete(routes, r.ID)
}
```

The same applies to pb.Interface where we can use interface.Name to uniquely identify an interface and thus use a map instead of a slice to store it.

network.go
```diff
@@ -25,10 +25,10 @@ import (
 // related information.
 type network struct {
 	ifacesLock sync.Mutex
-	ifaces     []*pb.Interface
+	ifaces     []pb.Interface
```
bergwolf: As I mentioned in https://github.com/kata-containers/agent/pull/225/files#r184835917, we can use pb.Interface.Name to uniquely identify an interface and thus use a map instead of a slice to store it. Then it is perfectly safe to call delete in a loop.

grpc.go
```diff
@@ -892,10 +892,6 @@ func (a *agentGRPC) DestroySandbox(ctx context.Context, req *pb.DestroySandboxRe
 	}
 	a.sandbox.Unlock()
 
-	if err := a.sandbox.removeNetwork(); err != nil {
```
bergwolf: Based on the same logic, can we simplify `DestroySandbox` by removing `removeMounts` and `teardownSharedPidNs` as well? Assuming that we do not reuse a sandbox after destroying it, the only thing must being done here is necessary pre-condition check and syscall.Sync().

sboeuf: We need to discuss about this, but if it's done, it's gonna be done in a follow up PR. This PR is more about the network.

network.go
sboeuf: Either way is fine with me. Do you really want that we introduce maps here, or you're okay with this way ?

grpc.go
bergwolf: Yes, agreed.

network.go
bergwolf: Please change to use a map for it. It has a few advantages:
- suitable for delete within a loop
- ensuring each interface is unique

network.go
```diff
@@ -25,10 +25,10 @@ import (
...
 
 	routesLock sync.Mutex
-	routes     []*pb.Route
+	routes     []pb.Route
```
bergwolf: We should be able to use a map here as well, by introducing a name/id field for each route rule. Then we can delete the rule in a loop safely. But it's ok if you want to defer it to a later PR.

sboeuf: Yes, this is in the scope for another PR here, because we don't have such thing as route ID, and we actually check for the entire route fields to detect if this is the same route.

network.go
sboeuf: Ok, it works in this case because the name is already what we use to identify what we need to remove.

sboeuf:
@bergwolf PR updated PTAL

sboeuf:
@egernst PTAL

sboeuf:
@egernst PR updated.

egernst:
@sboeuf  - I think for paranoia sake it would be wise to keep just the interface removal part (I don't see need for removing routes/dns).

sboeuf:
@egernst I have updated the PR. I am no longer removing the code removing the interfaces.
Do you want me to remove the code removing DNS ?

egernst:
@sboeuf - yes, please remove that as well.  LGTM otherwise - I'm happy with merging one you make this update.

sboeuf:
@egernst done !

sboeuf:
@grahamwhaley PTAL

PR #227, opened 2018-04-26
Use PCI Addresses to determine the device names for virtio-blk devices

### LINKED ISSUE #198: Get rid of prediction of device names on the host for virtio-blk
In case of virtio-blk, we currently predict disk names on the host side, instead handle finding out the correct disk name based on the pci address space. This will help to get rid of errors in predicting names.

amshinde:
This PR gets rid of the device name prediction needed on the host side for block devices. This can be used for device handler that we add in the future including VFIO.

Fixes #198

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"runtime/debug"
 	"strings"
@@ -23,6 +24,7 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/kata-containers/agent/pkg/uevent"
 	pb "github.com/kata-containers/agent/protocols/grpc"
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/configs"
@@ -74,6 +76,8 @@ type sandbox struct {
 	mounts          []string
 	subreaper       reaper
 	server          *grpc.Server
+	pciDeviceMap    map[string]string
+	deviceWatchers  map[string](chan string)
 }
 
 type namespace struct {
@@ -303,6 +307,55 @@ func (s *sandbox) teardownSharedPidNs() error {
 	s.sharedPidNs = namespace{}
 
 	return nil
+}
+
+func (s *sandbox) listenToUdevEvents() {
+	fieldLogger := agentLog.WithField("subsystem", "udevlistener")
+
+	uEvHandler, err := uevent.NewHandler()
+	if err != nil {
+		fieldLogger.Warnf("Error starting uevent listening loop %s", err)
+		return
+	}
+	defer uEvHandler.Close()
+
+	for {
+		uEv, err := uEvHandler.Read()
+		if err != nil {
+			fieldLogger.Error(err)
+			continue
+		}
+
+		fieldLogger = fieldLogger.WithFields(logrus.Fields{
+			"uevent-action":    uEv.Action,
+			"uevent-devpath":   uEv.DevPath,
+			"uevent-subsystem": uEv.SubSystem,
+			"uevent-seqnum":    uEv.SeqNum,
+			"uevent-devname":   uEv.DevName,
+		})
+
+		// Check if device hotplug event results in a device node being created.
+		if uEv.DevName != "" && uEv.Action == "add" && strings.HasPrefix(uEv.DevPath, rootBusPath) {
+			// Lock is needed to safey read and modify the pciDeviceMap and deviceWatchers.
+			// This makes sure that watchers do not access the map while it is being updated.
+			s.Lock()
+
+			// Add the device node name to the pci device map.
+			s.pciDeviceMap[uEv.DevPath] = uEv.DevName
+
+			// Notify watchers that are interested in the udev event.
+			// Close the channel after watcher has been notified.
+			for devPCIAddress, ch := range s.deviceWatchers {
+				if ch != nil && strings.HasPrefix(uEv.DevPath, filepath.Join(rootBusPath, devPCIAddress)) {
+					ch <- uEv.DevName
+					close(ch)
+					delete(s.deviceWatchers, uEv.DevName)
+				}
+			}
+
+			s.Unlock()
+		}
+	}
 }
 
 // This loop is meant to be run inside a separate Go routine.
@@ -643,8 +696,10 @@ func main() {
 		running:    false,
 		// pivot_root won't work for init, see
 		// Documention/filesystem/ramfs-rootfs-initramfs.txt
-		noPivotRoot: os.Getpid() == 1,
-		subreaper:   r,
+		noPivotRoot:    os.Getpid() == 1,
+		subreaper:      r,
+		pciDeviceMap:   make(map[string]string),
+		deviceWatchers: make(map[string](chan string)),
 	}
 
 	if err = s.initLogger(); err != nil {
@@ -664,6 +719,8 @@ func main() {
 
 	// Start gRPC server.
 	s.startGRPC()
+
+	go s.listenToUdevEvents()
 
 	s.wg.Wait()
 }
--- a/device.go
+++ b/device.go
@@ -29,6 +29,8 @@ const (
 	driverSCSIType = "scsi"
 )
 
+const rootBusPath = "/devices/pci0000:00"
+
 // SCSI variables
 var (
 	// Here in "0:0", the first number is the SCSI host number because
--- a/device.go
+++ b/device.go
@@ -31,6 +31,12 @@ const (
 
 const rootBusPath = "/devices/pci0000:00"
 
+var (
+	sysBusPrefix     = "/sys/bus/pci/devices"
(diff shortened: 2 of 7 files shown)


agent.go
```diff
@@ -305,6 +309,58 @@ func (s *sandbox) teardownSharedPidNs() error {
 	return nil
 }
 
+func (s *sandbox) listenToUdevEvents() {
+	uEvHandler, err := uevent.NewHandler()
+	if err != nil {
+		return
```
devimc: please log the error

agent.go
```diff
@@ -305,6 +309,58 @@ func (s *sandbox) teardownSharedPidNs() error {
...
+			s.pciDeviceMap[uEv.DevPath] = uEv.DevName
+
+			// Notify watchers that are interested in the udev event.
+			// Close the channel after watcher has been notified.
+			var matchedWatchers []string
+			for devPCIAddress, ch := range s.deviceWatchers {
+				if ch != nil && strings.HasPrefix(uEv.DevPath, filepath.Join(rootBusPath, devPCIAddress)) {
+					ch <- uEv.DevName
+					close(ch)
+					matchedWatchers = append(matchedWatchers, uEv.DevName)
```
devimc: I think you can remove here to the watchers that have been notified, right? , hence you don't need `var matchedWatchers []string`

device.go
```diff
@@ -42,31 +47,110 @@ var (
...
-	// First need to make sure the expected device shows up properly,
-	// and then we need to retrieve its device info (such as major and
-	// minor numbers), useful to update the device provided
-	// through the OCI specification.
-	devName := strings.TrimPrefix(device.VmPath, devPrefix)
-	checkUevent := func(uEv *uevent.Uevent) bool {
-		return (uEv.Action == "add" &&
-			filepath.Base(uEv.DevPath) == devName)
+func getDevicePCIAddress(pciID string) (string, error) {
+	tokens := strings.Split(pciID, "/")
```
devimc: can `pciID` have more than 2 `/` ?

device.go
```diff
@@ -42,31 +47,110 @@ var (
...
+
+	busNum := len(files)
+	if busNum != 1 {
+		return "", fmt.Errorf("Expected an entry for bus in %s, got %d entries instead", bridgeBusPath, busNum)
+	}
+
+	bus := files[0].Name()
+
+	// Device address is based on the bus of the bridge to which it is attached.
+	// We do not pass devices are multfunction, hence the trailing 0 in the address.
```
devimc: do you mean "We do not pass devices _as_ multfunction" ?

device.go
```diff
@@ -42,31 +47,110 @@ var (
...
+	if len(tokens) != 2 {
+		return "", fmt.Errorf("PCI Identifier for device should be of format [bridgeAddr/deviceAddr], got %s", pciID)
+	}
+
+	bridgeId := tokens[0]
+	deviceId := tokens[1]
+
+	// Deduce the complete bridge address based on the bridge address identifier passed
+	// and the fact that bridges are attached on the main bus with function 0.
+	pciBridgeAddr := fmt.Sprintf("0000:00:%s.0", bridgeId)
```
devimc: have you tested this with two or more bridges?

devimc: I think this will fail if devices are attached to a second bridge, but I'm not sure

agent.go
amshinde: Will do

agent.go
amshinde: I dont think it is safe to delete an entry in the map while looping through it.

device.go
amshinde: No, this is supposed to be "bridgeAddr/deviceAddr" passed by the runtime.

device.go
amshinde: Yes, typo there. Will fix.

device.go
amshinde: Why do you think it should fail? All bridges are added on the main/root bus.

agent.go
devimc: ohh, you're right  :smile:

device.go
devimc: my spider-sense alerted me about that, would be nice to have a test for that

device.go
egernst: is it multifunction or multifunction?

amshinde: Fixed

agent.go
amshinde: Done

device.go
amshinde: @devimc Have tested this, since all  bridges are attached to the main bus, they are expected to have a syspath under  `/sys/devices/pci0000:00/`

amshinde:
I see that the CI is failing :
```
=== RUN   TestUpdateRoutes
ERRO[0002] update Route failed                           error="rpc error: code = Internal desc = Could not add route dest()/gw(192.168.0.1)/dev(ifc-name): network is unreachable" name=kata-agent pid=10876 source=agent
```

I compared this with previous runs and I see that this test was skipped earlier as it needs root privileges. This test is now being run as root and is failing.
@chavafg @egernst Can you take a look?

amshinde:
@sboeuf @egernst @bergwolf PTAL

agent.go
amshinde: @devimc I just came across documentation that mentions that it is safe to delete from a a map while iterating through it. So now I am doing this within the loop itself.

devimc:
thanks @amshinde for testing with more than 2 bridges

device.go
```diff
@@ -42,31 +47,110 @@ var (
...
 	}
-	if err := waitForDevice(device.VmPath, devName, checkUevent); err != nil {
+
+	// If device is not found in the device map, hotplug event has not
+	// been received yet, create and add channel to the watchers map.
+	// The key of the watchers map is the device we are interested in.
+	// Note this is done inside the lock, not to miss any events from the
+	// global udev listener.
+	if devName == "" {
+		notifyChan := make(chan string)
```
sboeuf: Probably better to create a buffered channel of 1 here. This way, the uevent listener won't get stuck waiting for someone to read the event.

agent.go
```diff
@@ -305,6 +309,53 @@ func (s *sandbox) teardownSharedPidNs() error {
...
+			"uevent-action":    uEv.Action,
+			"uevent-devpath":   uEv.DevPath,
+			"uevent-subsystem": uEv.SubSystem,
+			"uevent-seqnum":    uEv.SeqNum,
+			"uevent-devname":   uEv.DevName,
+		})
+
+		// Check if device hotplug event results in a device node being created.
+		if uEv.DevName != "" && uEv.Action == "add" && strings.HasPrefix(uEv.DevPath, rootBusPath) {
+			s.Lock()
```
sboeuf: You could add a quick comment explaining the lock is needed to safely read and modify the `pciDeviceMap` map.

sboeuf:
@bergwolf PTAL

amshinde:
@sboeuf Changes applied.

sboeuf:
@amshinde maybe we should not merge this until we get the equivalent PR on the runtime also ready to be merged. I am saying this because by merging this PR, we will basically agree on breaking `virtio-blk` support, right ? Or am I wrong and there is still the backward compatibility to find the device based on the prediction ?

agent.go
```diff
@@ -665,5 +720,7 @@ func main() {
 	// Start gRPC server.
 	s.startGRPC()
 
+	go s.listenToUdevEvents()
```
bergwolf: please let s.wg.Wait() also wait for the udev listener goroutine.

device.go
```diff
@@ -42,31 +50,114 @@ var (
...
+			return "", grpcStatus.Errorf(codes.DeadlineExceeded,
+				"Timeout reached after %ds waiting for device %s",
+				timeoutHotplug, pciAddr)
+		}
+	}
+
+	return filepath.Join(systemDevPath, devName), nil
+}
+
+func virtioBlkDeviceHandler(device pb.Device, spec *pb.Spec, s *sandbox) error {
```
bergwolf: Now `device.Id` for both virtio-scsi and virtio-blk are well formatted addresses. Could you please add some comments above both the two handlers documenting the format of the device id?

device.go
```diff
@@ -42,31 +50,114 @@ var (
...
+		notifyChan := make(chan string, 1)
+		s.deviceWatchers[pciAddr] = notifyChan
+	}
+	s.Unlock()
+
+	if devName == "" {
+		fieldLogger.Info("Waiting on channel for device notification")
+		select {
+		case devName = <-notifyChan:
+		case <-time.After(time.Duration(timeoutHotplug) * time.Second):
```
bergwolf: Please remove the watcher and close its notification channel on timeout. Otherwise the channel may exist forever if the device never shows up.

device.go
amshinde: Have added in comments.

agent.go
amshinde: The udev listener routine does not quit for any reason, thats why I have not added the wait for it.
It should quit when the main thread quits.

device.go
amshinde: Fixed.

sboeuf:
@chavafg tested this, so we can go ahead and merge it.

PR #233, opened 2018-05-10
namespace: Share pid namespace of containers

### LINKED ISSUE #59: [RFC] Always create sandbox init process and knobs for joining

amshinde:
In case of k8s, containers in a pod are always supposed
to be in the same IPC, network, and UTC namespace. Sharing pid
namespace is a configuration option, containers are supposed to
be in their own PID namespace by default. This info needs to be
passed by the runtime.

The shared namespace is supposed to be the namespace of the
infra container. However, when we were sharing pid namespace we
were creating an additional pause process. This is not required
as the infra container is supposed to run the pause process.
The very first container will be the infra/pause container.
Make other containers join the namespaces of the infra container.

Also, remove `SandboxPidns` from CreateSandbox and move this to
CreatContainer request. This is because the knowledge that a container
needs to join a shared PID namespace is not available to the runtime
at the time of creating a sandbox. This information can be deduced
only while launching a container.

Fixes #59

Related PR for runtime : https://github.com/kata-containers/runtime/pull/298


egernst:
Nice, Archana.

grpc.go
```diff
@@ -394,29 +394,66 @@ func (a *agentGRPC) postExecProcess(ctr *container, proc *process) error {
 // path set by the spec, since we will always ignore it. Indeed, it makes no
 // sense to rely on the namespace path provided by the host since namespaces
 // are different inside the guest.
-func (a *agentGRPC) updateContainerConfigNamespaces(config *configs.Config) error {
+func (a *agentGRPC) updateContainerConfigNamespaces(config *configs.Config, ctr *container) {
+	pidNs := false
+	utcNs := false
+	ipcNs := false
```
jshachm: can it be 
`var pidNs, utcNs, ipcNs = false, false, false `
to make it simple

bergwolf:
Hi @amshinde, the infra container only exists because of docker's infamous pid1 problem, which does not exist with kata because container processes are always cleaned up by the kata agent. IOW, the infra container is not necessary for a kata deployment.

If an infra container is created, it exists same as other containers. I think it works without letting users notice any difference. Do you see any issue with it?

`SandboxPidns` is to specify whether the runtime wants to put all containers in the same pidns or not, which is controlled by the runtime. k8s would let the cri shims know this information and it got lost with crio/cri-containerd because of limitation of OCI spec -- which does not have a sandbox concept. But with frakti, we would be able to pass such information down to the runtime.

AFAIU, the only missing piece of #59 is the sandbox level net/uts/ipc ns for new containers to join.

And I agree `useSandboxPidNs` makes sense for kata cli case since crio/cri-containerd does not pass the sandbox namespace requirements down. Although the best way to fix it is to introduce yet another annotation to ask them pass down the sandbox namespace requirements :)

grpc.go
bergwolf: Or simply var foo, bar bool. The default value for bool is false.

bergwolf:
> the other containers are expected to join the namespace of the sandbox/infra container.
With sandbox level namespaces, all containers share the same namespaces and therefore the other containers are still joining the same namespaces as the infra container. The only difference is that the infra container does not own the namespaces any more.

> Thats why I got rid of the pause process that we were explicitlu starting for shared namespaces.
The agent subreaper makes sure we always clean up dead processes. What you did only works if there is a infra container. In other cases, there is nothing waiting for the zombie processes. It is an advantage of kata over docker to be free of the pid1 problem. If we take your change, I'm afraid we are copying docker's defects.

/cc @sboeuf @laijs @sameo who used to design this together.

amshinde:
@bergwolf fyi, I am going to raise a separate PR for solely handling IPC and UTS namespaces. These can be created as persistent namespaces and all containers need to share these. So we can create these before and make all containers join these. We can continue the discussion for handling PID namespace here, I will modify this PR accordingly.

Regarding the point 3) in your comment above, I looked a bit at k8s and I dont see a way to disable the infra container. One can replace this with a custom image, but I did not find a way to disable this altogether.
Both cri-containerd and crio-o rely on the infra-container for settting up namespaces.

Related PR in the runtime: https://github.com/kata-containers/runtime/pull/298

bergwolf:
@amshinde How about modifying `SandboxPidns` in create sandbox semantics to just mean creating internal pause process to hold the sandbox level pidns, and that otherwise the first container in a sandbox leads the sandbox level pidns?

Then upon each `CreateContaienr()`, we check container spec for NEWPID and the pidns path:
1. no NEWPID: the container joins the sandbox level pidns
2. NEWPID with empty pidns path: the container lives in its own pidns
3. NEWPID with non-empty pidns path: the container joins the sandbox level pidns

So you can remove the extra "pause" process by just setting `SandboxPidns` to false in `CreateSandbox()` in kata runtime in case of infra container. The `SandboxPidns` flag can be exported via the kata runtime APIs and the decision will be made by kata runtime cli.

This way, the kata agent and virtcontainers library both remain flexible enough to properly support non-infra container cases.

egernst:
I could use the IPC and UTS namespace portion of this discussion, as this is a priority request from some users. @amshinde do you have a PR which you can push or that I missed?

amshinde:
@egernst @bergwolf Finally got around to implement IPC and UTS sharing, here is the PR: https://github.com/kata-containers/agent/pull/249
@bergwolf  I like your suggestion about handling `SandboxPidns`, I'll modify this PR accordingly.

amshinde:
@bergwolf Branch updated to handle just pid namespace sharing. PTAL.

agent.go
```diff
@@ -53,33 +53,36 @@ type process struct {
...
-	config      configs.Config
-	processes   map[string]*process
-	mounts      []string
+	id              string
+	initProcess     *process
+	container       libcontainer.Container
+	config          configs.Config
+	processes       map[string]*process
+	mounts          []string
+	useSandboxPidNs bool
```
bergwolf: Can you drop this `useSandboxPidNs` field? As stated in my earlier comment, I think it can be determined by the container spec instead:
 - no NEWPID: the container joins the sandbox level pidns
 - NEWPID with empty pidns path: the container lives in its own pidns
 - NEWPID with non-empty pidns path: the container joins the sandbox level pidns

agent.go
```diff
@@ -53,33 +53,36 @@ type process struct {
...
+	wg               sync.WaitGroup
+	sharedPidNs      namespace
+	mounts           []string
+	subreaper        reaper
+	server           *grpc.Server
+	pciDeviceMap     map[string]string
+	deviceWatchers   map[string](chan string)
+	sharedUTSNs      namespace
+	sharedIPCNs      namespace
+	infraContainerId string
```
bergwolf: I don't think you need the `infraContainerId` field. You can determine if the container is the very first container in a sandbox and use it to set the sandbox level pidns path if `sandboxPidNs` is false.

amshinde: Yes, I just added this field to make this explicit. I can take this out.

agent.go
amshinde: I added this field since currently we are clearing out the namespace paths in the runtime before they are passed to the agent. I suppose this was done so as to not try to use the host namespaces accidentally in the agent (which we may need to check for mount, user namespaces?).
I also like the fact that this configuration is sent explicitly by the runtime. The runtime actually checks that the namespace for a container is the same as the very first container, and then sets the `useSandboxPidNs` field. This way we check that it is not just some other namespace path on the host. (Have raised a separate PR for this in the runtime.)
Whats do you think? Do you see any issues in passing this field explicitly?

bergwolf: Fair enough. I thought it duplicates with config.NEWPID but kata runtime is the only user and it clears the namespace path, I agree that explicit setting is better. Besides we are already diverging from OCI spec in the namespace handling, it is better not to use the spec to avoid confusion. Would you please add some comments on the proto change to make it clear that the field overrides whatever is set in config.NEWPID?

> which we may need to check for mount, user namespaces

Yes, we should. All namespace paths in the config should be cleared IMO. Otherwise they are potential attach surfaces.

agent.go
bergwolf: Thanks!

amshinde: Fixed.

amshinde:
@devimc @sboeuf can you take a look at this PR?

amshinde:
@bergwolf I have addressed your comments. PTAL.

protocols/grpc/agent.proto
```diff
@@ -59,6 +59,11 @@ message CreateContainerRequest {
 	repeated Device devices = 4;
 	repeated Storage storages = 5;
 	Spec OCI = 6;
+
+	// This field is used to indicate if the container needs to join
+	// sandbox shared pid ns. This is needed since we clear out all the
+	// namespace paths before they are sent to the agent.
```
bergwolf: Please be specific that it overrides NEWPID settings in container's OCI spec, not just the NEWPID namespace path.

grpc.go
```diff
@@ -415,16 +415,21 @@ func (a *agentGRPC) postExecProcess(ctr *container, proc *process) error {
...
 		if ns.Type == configs.NEWPID {
-			// In case the path is empty because we don't expect
-			// the containers to share the same PID namespace, a
-			// new PID ns is going to be created.
-			config.Namespaces[idx].Path = a.sandbox.sharedPidNs.path
+			// Use shared pid ns if SandboxPidns has been set in either
+			// the CreateSandbox request or CreateContainer request.
+			// Else set this to empty string so that a new pid namespace is
+			// created for the container.
+			if ctr.useSandboxPidNs || a.sandbox.sandboxPidNs {
```
bergwolf: hmm, so `a.sandbox.sandboxPidNs` still means all containers will join the kata agent pause process held sandbox level pidns. Is it intentional? I don't see any issue with it but want to make sure we are on the same page about its semantics. And whatever we decide, we need to document it in the proto file.

grpc.go
```diff
@@ -442,6 +447,9 @@ func (a *agentGRPC) updateContainerConfigNamespaces(config *configs.Config) erro
 	// If no NEWPID type was found, let's make sure we add it. Otherwise,
 	// the container could end up in the same PID namespace than the agent
 	// and we want to prevent this for security reasons.
+	// The container would then join the shared pid namespace of either the
+	// pause process created by the agent or the pid namespace of the infra
+	// container.
 	if !pidNs {
```
bergwolf: Without NEWPID settings in the config, we should still let `ctr.useSandboxPidNs` decide if the container will join the sandbox level pidns. Otherwise `ctr.useSandboxPidNs` is not working if container spec has no NEWPID settings. The mix of sandbox pidns, container use sandbox pidns and NEWPID settings is complicating the protocol semantics. So I want to simplify it by moving NEWPID settings out of the picture.

grpc.go
```diff
@@ -603,9 +607,30 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
 	return emptyResp, a.postExecProcess(ctr, ctr.initProcess)
 }
 
+func updateSharedPidNs(s *sandbox, ctr *container) error {
+	// Populate the shared pid path only if this is an infra container and the
+	// shared path is empty. This means SandboxPidns has not been
+	// passed in the CreateSandbox request and a separate pause process has not been
+	// created. We treat the first container created as the infra container in that case
+	// and use its pid namespace in case it needs to be shared.
+	if s.sharedPidNs.path == "" && len(s.containers) == 1 {
```
bergwolf: Please use `a.sandbox.sandboxPidNs` to do the checking, and clear `s.sharedPidNs.path` when removing the last container from sandbox. Then you don't need an empty `s.sharedPidNs.path` to indicate `a.sandbox.sandboxPidNs` being false and the check works when removing the last container from sandbox.

grpc.go
amshinde: Typically, when `ctr.useSandboxPidNs` is set, NEWPID will be set because that bool indicates that indicates that the container needs to join a namespace belonging to the sandbox and this will be provided in the OCI spec file.
This is also in accordance with your earlier comment
`no NEWPID: the container joins the sandbox level pidns
NEWPID with empty pidns path: the container lives in its own pidns
NEWPID with non-empty pidns path: the container joins the sandbox level pidns`

no NEWPID-> container joins the sandbox level pidns (which is what is done here)

grpc.go
amshinde: I dont follow. Let me explain what I have done here.
If `a.sandbox.sandboxPidNs`, then we create our own pause process and  use that as the `s.sharedPidNs.path`. This is to satisfy the frakti scenario.
Usually either `a.sandbox.sandboxPidNs` will be set(frakti) or `ctr.useSandboxPidNs`(CRI).

If `a.sandbox.sandboxPidNs` is set then we continue using agent's pause process pidns as the shared pidns. If this is not set, and if `ctr.useSandboxPidNs` is set, then the first process is treated as the infra container and its pidns is shared. Because in this case a separate pause process has not been created.
Thats why I check for the blank value for `s.sharedPidNs.path`. It is populated once either by the pause process that agent creates or by the very first container(infra container, which is the last one to exit)

amshinde: Also my runtime PR just sets `ctr.useSandboxPidNs`  in the cri-o and cri-containerd use cases. The flag `a.sandbox.sandboxPidNs` which existed previously is not set at all, I assume this will be set when called through frakti.

grpc.go
bergwolf: My comment as you quoted above is to remove `ctr.useSandboxPidNs`. If you keep it, we should not rely on `NEWPID` settings here. The point is that we should not depend on two separate flags for one thing. 

> Typically, when ctr.useSandboxPidNs is set, NEWPID will be set

We cannot assume it. This is a protocol API and we need to make it clear what is in effect and what is not.

grpc.go
amshinde: @bergwolf When pid namespace is shared, this is very much true for k8s. The first container will be the last one to exit, since it is the pid 1 process of the sandbox. I have verified this for both cri-containerd and crio. If the pid 1 exits, the other containers will be terminated. 
For frakti which does not really on an infra container, this will be handled by creating a pause process by the agent by setting `a.sandbox.sharedPidNs` to true and in that case `s.sharedPidNs.path` will be populated with the pause process pidns and not that of any container.

bergwolf: @amshinde The kata agent should be self-contained and not depend on its users to be correct internally. The kata agent is robust on its own, not JustWork(R) with k8s/docker/frakti. And the sandbox life cycle and container life cycle is different by design. That's why we have CreateSandbox and DestroySandbox APIs.

sboeuf:
@amshinde I was about to review this PR, but I have realized that you've been discussing on #kata-dev with @bergwolf about modifying the PR. I'll review it tomorrow when you're done with reworking it !

grpc.go
amshinde: @bergwolf Fair enough. Have fixed this.

grpc.go
```diff
@@ -603,9 +601,31 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
 	}
 
+	if err := updateSharedPidNs(a.sandbox, ctr); err != nil {
+		return emptyResp, err
+	}
+
 	return emptyResp, a.postExecProcess(ctr, ctr.initProcess)
 }
 
+func updateSharedPidNs(s *sandbox, ctr *container) error {
```
sboeuf: The following prototype might be more appropriate, right ?
```go
func (a *agentGRPC) updateSharedPidNs(ctr *container) error
```

amshinde:
@sboeuf Pushed the function member change. PTAL

sboeuf:
Thanks @amshinde 
LGTM

sboeuf:
Let's merge it when CI gets green !

amshinde:
@sboeuf CI has passed. What is the codecov/patch though? I ask because tests have been added which increase the overall code coverage.

sboeuf:
@amshinde I don't know... @bergwolf any idea where this comes from ?

sboeuf:
@egernst @devimc @jodh-intel we need one more approval here !

sboeuf:
@amshinde codecov failure might be related to the tests files of the protocol not being properly generated. How did you generate the protocol after you did  the modifications ?

amshinde:
@sboeuf I used the `hack/update-generated-agent-proto.sh` script to generate the protocol. Do you see any issues? Can you generate at your end and let me know if you see any discrepencies.

PR #236, opened 2018-05-14
agent: Add support for ephemeral volumes

harche:
Ephemeral volumes needs to mounted inside VM
and not passed as 9pfs mounts.

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	type9pFs       = "9p"
+	typeTmpFs      = "tmpfs"
 	devPrefix      = "/dev/"
 	timeoutHotplug = 3
 	mountPerm      = os.FileMode(0755)
@@ -95,26 +96,28 @@ func mount(source, destination, fsType string, flags int, options string) error 
 		return fmt.Errorf("need mount FS type")
 	}
 
-	if fsType != type9pFs {
-		var err error
-
+	var err error
+	switch fsType {
+	case type9pFs:
+		if err = createDestinationDir(destination); err != nil {
+			return err
+		}
+		absSource = source
+	case typeTmpFs:
+		absSource = source
+	default:
 		absSource, err = filepath.EvalSymlinks(source)
 		if err != nil {
 			return grpcStatus.Errorf(codes.Internal, "Could not resolve symlink for source %v", source)
 		}
 
-		if err := ensureDestinationExists(absSource, destination, fsType); err != nil {
+		if err = ensureDestinationExists(absSource, destination, fsType); err != nil {
 			return grpcStatus.Errorf(codes.Internal, "Could not create destination mount point: %v: %v",
 				destination, err)
 		}
-	} else {
-		if err := createDestinationDir(destination); err != nil {
-			return err
-		}
-		absSource = source
 	}
 
-	if err := syscall.Mount(absSource, destination,
+	if err = syscall.Mount(absSource, destination,
 		fsType, uintptr(flags), options); err != nil {
 		return grpcStatus.Errorf(codes.Internal, "Could not mount %v to %v: %v",
 			absSource, destination, err)
@@ -187,9 +190,20 @@ type storageHandler func(storage pb.Storage, s *sandbox) (string, error)
 
 // storageHandlerList lists the supported drivers.
 var storageHandlerList = map[string]storageHandler{
-	driver9pType:   virtio9pStorageHandler,
-	driverBlkType:  virtioBlkStorageHandler,
-	driverSCSIType: virtioSCSIStorageHandler,
+	driver9pType:        virtio9pStorageHandler,
+	driverBlkType:       virtioBlkStorageHandler,
+	driverSCSIType:      virtioSCSIStorageHandler,
+	driverEphemeralType: ephemeralStorageHandler,
+}
+
+func ephemeralStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
+	if _, err := os.Stat(storage.MountPoint); os.IsNotExist(err) {
+		if err = os.MkdirAll(storage.MountPoint, os.ModePerm); err == nil {
+			_, err = commonStorageHandler(storage)
+		}
+		return "", err
+	}
+	return "", nil
 }
 
 // virtio9pStorageHandler handles the storage for 9p driver.
--- a/device.go
+++ b/device.go
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	driver9pType   = "9p"
-	driverBlkType  = "blk"
-	driverSCSIType = "scsi"
+	driver9pType        = "9p"
+	driverBlkType       = "blk"
+	driverSCSIType      = "scsi"
+	driverEphemeralType = "ephemeral"
 )
 
 const rootBusPath = "/devices/pci0000:00"
--- a/mount_test.go
+++ b/mount_test.go
@@ -30,6 +30,22 @@ func createSafeAndFakeStorage() (pb.Storage, error) {
 	}, nil
 }
 
+func TestEphemeralStorageHandlerSuccessful(t *testing.T) {
+	skipUnlessRoot(t)
+
+	storage, err := createSafeAndFakeStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer syscall.Unmount(storage.MountPoint, 0)
+	defer os.RemoveAll(storage.MountPoint)
+
+	storage.Fstype = typeTmpFs
+	storage.Source = typeTmpFs
+	_, err = ephemeralStorageHandler(storage, &sandbox{})
+	assert.Nil(t, err, "ephemeralStorageHandler() failed: %v", err)
+}
+
 func TestVirtio9pStorageHandlerSuccessful(t *testing.T) {
 	skipUnlessRoot(t)
 


harche:
Corresponding PR in `runtime`, https://github.com/kata-containers/runtime/pull/307

mount.go
```diff
@@ -98,15 +99,23 @@ func mount(source, destination, fsType string, flags int, options string) error
 	if fsType != type9pFs {
 		var err error
 
-		absSource, err = filepath.EvalSymlinks(source)
-		if err != nil {
-			return grpcStatus.Errorf(codes.Internal, "Could not resolve symlink for source %v", source)
+		if fsType == typeTmpFs {
```
egernst: Since we have more than a couple cases now for the fsType now, I wonder if it'd warrant throwing this into a case now?  WDYT?

grpc.go
```diff
@@ -515,6 +515,34 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	for idx, ociMnt := range ociSpec.Mounts {
+		splitSourceSlice := strings.Split(ociMnt.Source, "/")
+		if len(splitSourceSlice) > 1 {
+			k8sStorageType := splitSourceSlice[len(splitSourceSlice)-2]
+			if k8sStorageType == "emphemeral" {
+				// Only mount tmpfs once on the host, and reuse the the mount destination
```
egernst: Just my own ignorance likely -- can you clarify "mount temps once on the host" ?  The agent is running in the guest.

grpc.go
```diff
@@ -515,6 +515,34 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	for idx, ociMnt := range ociSpec.Mounts {
+		splitSourceSlice := strings.Split(ociMnt.Source, "/")
+		if len(splitSourceSlice) > 1 {
+			k8sStorageType := splitSourceSlice[len(splitSourceSlice)-2]
+			if k8sStorageType == "emphemeral" {
```
jodh-intel: - Is this a reliable test? This appears to be looking for the word "emphemeral" in the `ociMnt.Source` *path* (and that "magic tag" might change at some point.
- This block uses deep indentation. You could save one level by restructuring slightly:
  ```
  if k8sStorageType != "emphemeral" {
      continue
  }

  ...
  ```

grpc.go
```diff
@@ -515,6 +515,34 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	for idx, ociMnt := range ociSpec.Mounts {
+		splitSourceSlice := strings.Split(ociMnt.Source, "/")
+		if len(splitSourceSlice) > 1 {
+			k8sStorageType := splitSourceSlice[len(splitSourceSlice)-2]
```
jodh-intel: You could create a variable for `len(splitSourceSlice)` as you've scanned the string twice here.

grpc.go
jshachm: Maybe we shouldn't mount `tmpfs` in the host. And direct mount it in the `guest vm`

mount.go
harche: Yes, `switch case` makes more sense here.

grpc.go
harche: @egernst Sorry I will update that comment in the code to avoid the confusion. It should read, `Only mount tmpfs once inside the VM, and reuse the the mount destination`


`Ephemeral Volumes` live and die with the pod. They are created and mounted on the host by kublet and then binded mounted to every container in the pod that wants to use the volume. In case of kata, the pod is our VM Sandbox. So the volume has to be confined within the VM context. This tmpfs has to be created inside the VM, hence you see the code to create tmpfs in the `agent`. This way we increase the isolation provided by kata to the containers. The containers that want to share the data using ephemeral volumes won't have to `leak` their data to the host because the backing volume stays within VM memory. 

@jshachm By the time control comes to kata, kublet has already provision tmpfs on the host. We can choose to ingore that and provision our tmpfs inside the guest as this patching achieving to do.

bergwolf:
Instead of checking k8sStorageType, we should add an ephemeral storage driver type and let ociSpec.Mounts just reference it. For example,

Let req.stroage have
Strorage{Driver:"ephemeral", fstype:"tmpfs", MountPoint: <sandbox-mountpoint-foo>}

And in ociSpec.Mounts, let ociSpec.Mounts[i].Mount have
Mount{Source:<sandbox-mountpoint-foo>, Destination:< ephemeral-volume-inside-of-container>, Type: "bind"}

The main change is in addStorages, we create the tmpfs mountpoint at <sandbox-mountpoint-foo>, with a new storage driver handler.

grpc.go
harche: @jodh-intel Is it very reliable to check for `k8sStorageType == "emphemeral"` condition. The string `emphemeral` is introduced by this patch, https://github.com/kata-containers/runtime/pull/307/files#diff-8a56f5a1fa5971cc43e650ffa24f1e2bR376 

That patch in `runtime` loops over the given OCI mounts to look for the mounts used by k8s for ephemeral storage. It then replaces those mount paths on the host with the mount paths in the guest by adding `"/ephemeral"` to the path. So it's very much in our control the existance of that string. Right now without this patch, an ephemeral volume is passed to guested via 9p. 

e.g. A typical `ephemeral volume` path looks like, `/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume`

But this resides on the host. We want that volume to be provisioned inside our VM. We can keep the same path or we can make it simpler since we have the total control on what happens inside the sandbox VM. So I replace that original source path with much simpler `"/ephemeral"`. Inside the VM that above path becomes, `/ephemeral/cache-volume`

grpc.go
harche: sure.

grpc.go
harche: @jodh-intel I will add some unit test cases.

harche:
@bergwolf could you give me a pointer to the code? When the sandbox is coming up it has no idea of any storage that containers are attached to. Ideally, if sandbox was aware of the ephemeral volume (from pod configuration) we could provision a tmpfs inside the sandbox VM and have that bind mounted to the containers after sanbox init is complete.

harche:
@bergwolf Sounds good, I will give it a shot and update the PR.

sboeuf:
@bergwolf I completely agree this should be the way to do it. We have designed `Storage` and `Device` for this purpose.
@harche please let us know when the code is ready :)

sboeuf:
@harche just to clarify about ephemeral volumes. Do you expect anything to be present on the host at a location like `/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume` ?

I ask because if this is basically an empty volume, then great, a simple volume backed by tmpfs on the guest will do the trick.
But if we expect some data to be there initially, then we would also have to share them, or copy them inside the guest as a first step.

Just want to make sure we're on the same page !

harche:
@sboeuf There isn't any data initially. Ephemeral volumes are used only for sharing the data between the containers of the same pod. So the volume is empty to begin with unless a container from the respective pod writes something in it.

sboeuf:
@harche okay thanks for clarification ;)

harche:
@sboeuf @bergwolf @jodh-intel I have updated this PR as well as corresponding [PR](https://github.com/kata-containers/runtime/pull/307) in runtime.

bergwolf:
@harche Thanks for the change. It
LGTM!

mount.go
```diff
@@ -187,9 +188,21 @@ type storageHandler func(storage pb.Storage, s *sandbox) (string, error)
...
+	driverEphemeralType: ephemeralStorageHandler,
+}
+
+func ephemeralStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
+	if _, err := os.Stat(storage.MountPoint); os.IsNotExist(err) {
+		os.MkdirAll(storage.MountPoint, os.ModePerm)
+		_, err = commonStorageHandler(storage)
+		return "", err
+
+	} else {
```
caoruidong: You can delete this else to make CI happy.

harche: sure.

jodh-intel:
Thanks @harche!

lgtm

mount.go
```diff
@@ -187,9 +188,19 @@ type storageHandler func(storage pb.Storage, s *sandbox) (string, error)
...
-	driverSCSIType: virtioSCSIStorageHandler,
+	driver9pType:        virtio9pStorageHandler,
+	driverBlkType:       virtioBlkStorageHandler,
+	driverSCSIType:      virtioSCSIStorageHandler,
+	driverEphemeralType: ephemeralStorageHandler,
+}
+
+func ephemeralStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
+	if _, err := os.Stat(storage.MountPoint); os.IsNotExist(err) {
+		os.MkdirAll(storage.MountPoint, os.ModePerm)
```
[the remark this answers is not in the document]
jodh-intel: Good catch @sboeuf!

mount.go
```diff
@@ -95,9 +96,14 @@ func mount(source, destination, fsType string, flags int, options string) error
 		return fmt.Errorf("need mount FS type")
 	}
 
-	if fsType != type9pFs {
-		var err error
-
+	var err error
```
amshinde: nit: you dont need to declare err here, since you use a new instance in the inner scope.

mount.go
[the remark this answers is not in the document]
amshinde: @harche You do not even need to check if the directory exists before calling `MkdirAll`.  From the docs,  "If path is already a directory, MkdirAll does nothing and returns nil."

mount.go
```diff
@@ -95,9 +96,14 @@ func mount(source, destination, fsType string, flags int, options string) error
...
+	switch fsType {
+	case typeTmpFs, type9pFs:
+		if err := createDestinationDir(destination); err != nil {
```
amshinde: @harche Are you not already creating the mount destination directory in the "ephemeral" storage handler?

mount.go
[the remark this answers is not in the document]
sboeuf: @amshinde Well I think you need to check this because in case it does exist, you don't need to call into `commonStorageHandler()`.

mount.go
harche: yes

mount.go
harche: Yes, I will update the PR. Thanks.

mount.go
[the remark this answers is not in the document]
amshinde: and why are we not calling then `commonStorageHandler()` then? It may happen that the directory exists, but we skip mounting altogether.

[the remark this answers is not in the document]
harche: @amshinde the tmpfs mount point has to be mounted on the guest only once. Then rest of the containers will bind mount that directory. 

e.g. If all the containers need to have an ephemeral volume in the pod they will bind mount, say /sharedMount directory on the VM. But before that this /sharedMount has to be mounted on the VM backed by tmpfs (but only once)

The above code is tested with a k8s yaml with 3 containers sharing the mount point. 

```
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  labels:
    app: myapp
spec:
  containers:
  # Application Container
  - name: myapp-container
    volumeMounts:
    - mountPath: /cache
      name: cache-volume
    imagePullPolicy: Never
    image: app_image:latest
    command: ['sh', '-c', 'cat /cache/test && sleep 3600']
  initContainers:
  # Copy Data container
  - name: data-container
    volumeMounts:
    - mountPath: /cache
      name: cache-volume
    imagePullPolicy: Never
    image: data_image:latest
    command: ['sh', '-c', 'cp /test /cache/']
  # Append Data container
  - name: init-append-data
    volumeMounts:
    - mountPath: /cache
      name: cache-volume
    imagePullPolicy: Never
    image: test_image:latest
    command: ['sh', '-c', 'echo test >> /cache/test']
  volumes:
  - name: cache-volume
    emptyDir:
      medium: "Memory"
``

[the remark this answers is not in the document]
harche: @amshinde It's is intentionally skipped mounting if dir exists because then every single container end up pointing towards a seperate tmpfs volume and they don't end up sharing the volume. So only once (for the first time) you have to make sure the dir exists and then mount it on guest backed by tmpfs. The corresponding enteries in all other container's OCI spec already point towards bind mounting that dir.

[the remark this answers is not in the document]
amshinde: @harche Thanks for the explanation, my question was based on the assumption that you would be adding storage just once for the first container.

amshinde:
@harche CI is failing,can you take a look so that this can be merged.

sboeuf:
@harche please fix the unit tests.

harche:
@sboeuf @amshinde done.

Issue #240, opened 2018-05-16
Support Epoll on Arm64


Pennyzct:
# Description of problem

` sudo docker run -ti --runtime kata-runtime busybox sh
`
# Expected result

run successfully

# Actual result

log info from qemu output:
```
announce: name="kata-agent",pid='\x01',source="agent",version="0.0.1-8856b3acb43a90fe30f120b3e083807720829e4f",device-handlers="scsi,blk",storage-handlers="scsi,9p,blk",system-memory="2054184 kB"
announce: name="kata-agent",pid='\x01',source="agent",version="0.0.1-8856b3acb43a90fe30f120b3e083807720829e4f",device-handlers="blk,scsi",storage-handlers="9p,blk,scsi",system-memory="2054184 kB"
time="2018-05-10T07:26:31.52054312Z" level=info msg="agent grpc server starts" name=kata-agent pid=1 source=agent
init[332]: syscall 1069
Code: f94023e4 f94027e5 f9400fe8 d4000001 (b13ffc1f) 
CPU: 0 PID: 332 Comm: init Not tainted 4.9.51-hyper #1
Hardware name: linux,dummy-virt (DT)
task: ffffffc008c20100 task.stack: ffffffc008cc8000
PC is at 0x473cf8
LR is at 0x473cd8
pc : [<0000000000473cf8>] lr : [<0000000000473cd8>] pstate: 20000000
sp : 0000004420041df0
x29: 0000007f7fffe950 x28: 0000004420001b00 
x27: 0000000000de9880 x26: 0000004420041e00 
x25: 00000000008abcf8 x24: 0000007f8e2d8000 
x23: 0000007f8e2dc000 x22: 0000007f7ffff8d0 
x21: 000000000000005a x20: 0000000000000000 
x19: 000000000000003c x18: 0000000000000000 
x17: 0000004420041e90 x16: 0000004420041e08 
x15: 0000000000000e10 x14: 0000000000000008 
x13: 000000441fffdbcb x12: 0000000000000000 
x11: 0000000000000000 x10: 00000000000000f3 
x9 : 0000004420048690 x8 : 000000000000042d 
x7 : 0000000000000030 x6 : 0000000000000001 
x5 : 0000000000000000 x4 : 0000000000000000 
x3 : ffffffffffffffff x2 : 0000000000000001 
x1 : 0000004420041ee0 x0 : 0000000000000006 

[...]

PR #241, opened 2018-05-16
channel: Support Epoll on Arm64

Pennyzct:
syscall.EpollWait() gives ENOSYS: function not implemented error
(from Linux kernel), thus because epoll_wait is deprecated, and
only the newer epoll_pwait is provided on Linux/arm64. Golang has
fixed this issue(https://github.com/golang/go/issues/10235).
However, the implementation exits in x/sys/unix, not in the
standard Go syscall package.

Fixes: #240

Signed-off-by: Penny Zheng <penny.zheng@arm.com>
Signed-off-by: Wei Chen <wei.chen@arm.com>

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -13,7 +13,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/hashicorp/yamux"
 	"github.com/mdlayher/vsock"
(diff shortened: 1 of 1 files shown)


bergwolf:
In that case, how about just switching to epoll_pwait?

bergwolf:
hmm, I wanted to favor the syscall package over the unix package but I guess we are bound to linux anyway. So,
LGTM! Thanks for the patch!

Pennyzct:
Quite a few favor the `syscall` over `x/sys/unix`, however under the discussion of the implementation of deprecated syscall on Arm64 in golang community([issue 10235](https://github.com/golang/go/issues/10235)), they chose x/sys/unix over syscall. So, it's quite the only legit solution for us. thanks for understanding.😊

jodh-intel:
Thanks @Pennyzct.

lgtm

PR #244, opened 2018-05-21
Get cgroup mounts info from /proc/cgroups

keloyang:
Get cgroup mounts info from /proc/cgroups, this can get accurate cgroup subsystem info, if kernel don't surport to mount "/sys/fs/cgroup/cpu,cpuacct", agent should mount cpu and cpuacct subsystem separately.

Fix https://github.com/kata-containers/agent/issues/250
Signed-off-by: Shukui Yang <yangshukui@huawei.com>

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -7,6 +7,7 @@
 package main
 
 import (
+	"bufio"
 	"errors"
 	"flag"
 	"fmt"
@@ -37,7 +38,19 @@ import (
 	grpcStatus "google.golang.org/grpc/status"
 )
 
-const meminfo = "/proc/meminfo"
+const (
+	procCgroups = "/proc/cgroups"
+	cgroupPath  = "/sys/fs/cgroup"
+	meminfo     = "/proc/meminfo"
+)
+
+var initRootfsMounts = []initMount{
+	{"proc", "proc", "/proc", []string{"nosuid", "nodev", "noexec"}},
+	{"sysfs", "sysfs", "/sys", []string{"nosuid", "nodev", "noexec"}},
+	{"devtmpfs", "dev", "/dev", []string{"nosuid"}},
+	{"tmpfs", "tmpfs", "/dev/shm", []string{"nosuid", "nodev"}},
+	{"devpts", "devpts", "/dev/pts", []string{"nosuid", "noexec"}},
+}
 
 type process struct {
 	id          string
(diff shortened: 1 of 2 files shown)


devimc:
@keloyang please file and issue, add a body to your commit message and `fixes #NNN`

keloyang:
@devimc thanks, I have added the issue  https://github.com/kata-containers/agent/issues/250

agent.go
```diff
@@ -622,44 +623,116 @@ type initMount struct {
...
+		}
+		cgroupMounts = append(cgroupMounts, initMount{"cgroup", "cgroup",
+			filepath.Join(cgroupPath, cgroup), []string{"nosuid", "nodev", "noexec", "relatime", cgroup}})
+	}
+
+	if err = scanner.Err(); err != nil {
+		return []initMount{}, err
+	}
+
+	cgroupMounts = append(cgroupMounts, initMount{"tmpfs", "tmpfs",
```
bergwolf: cgroup tmpfs is added twice.

agent.go
```diff
@@ -622,44 +623,116 @@ type initMount struct {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func cgroupsMount() error {
+	cgroups, err := getCgroupMounts(procCgroups)
+	if err != nil {
+		cgroups = initCgroupMounts
```
bergwolf: I don't think we need to fallback here. If getCgroupMounts fails, there is no cgroups support in the kernel at all and kata agent should fail to start. And the fallback will fail as well. So there is no need for it.

agent.go
```diff
@@ -622,44 +623,116 @@ type initMount struct {
...
+			continue
+		}
+		cgroupMounts = append(cgroupMounts, initMount{"cgroup", "cgroup",
+			filepath.Join(cgroupPath, cgroup), []string{"nosuid", "nodev", "noexec", "relatime", cgroup}})
+	}
+
+	if err = scanner.Err(); err != nil {
+		return []initMount{}, err
+	}
+
```
bergwolf: please add a check to make sure devices cgroup is supported.

agent.go
keloyang: It is not repetition ， it remove /sys/fs/cgroup to ro.

agent.go
keloyang: may be there is no need to check devices cgroup, if not supported, kata-cantainer start failed and return err.   on the other hand ，if check , agent also return a err or skip， kata-cantainer  still start fail.

agent.go
keloyang: Whether getCgroupMounts  is successful or not depend on mount /proc before getCgroupMounts ,if agent getCgroupMounts  before mount /proc,  getCgroupMounts failed， but the kernel still support cgroup.

in this pr getCgroupMounts comes after mounting /proc, so we need not fallback，need me remove it ?

agent.go
bergwolf: We follow the fail fast rule here. It's better to find out a failure sooner than later, especially the failure is so fatal that even busybox cannot start.

agent.go
bergwolf: oops, I missed the `ro` field. Sorry for the noise.

agent.go
keloyang: Remoted, thanks.

jshachm:
Can we have a new unit test  for this change ? @keloyang

agent.go
bergwolf: @keloyang what about this one?

agent.go
```diff
@@ -622,44 +635,83 @@ type initMount struct {
...
-// initAgentAsInit will do the initializations such as setting up the rootfs
-// when this agent has been run as the init process.
-func initAgentAsInit() error {
+func mountToRootfs(m initMount) error {
+	if err := os.MkdirAll(m.dest, os.FileMode(0755)); err != nil {
+		return err
+	}
+
+	if flags, options, err := parseMountFlagsAndOptions(m.options); err != nil {
+		return grpcStatus.Errorf(codes.Internal, "Could parseMountFlagsAndOptions(%v)", m.options)
```
jodh-intel: s/Could parseMountFlagsAndOptions(%v)/Could not parse mount flags %v/

agent.go
```diff
@@ -622,44 +635,83 @@ type initMount struct {
...
+	defer f.Close()
+
+	cgroupMounts := []initMount{{"tmpfs", "tmpfs", cgroupPath, []string{"nosuid", "nodev", "noexec", "mode=755"}}}
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		text := scanner.Text()
+		fields := strings.Split(text, "\t")
+		cgroup := fields[0]
+
+		if cgroup == "" || cgroup[0] == '#' || (len(fields) > 3 && fields[3] == "0") {
```
jodh-intel: This could do with a comment explaining what fields[3] is (`0`: cgroup enabled, `1`: disabled).

agent.go
keloyang: oh, sorry, I missed out it , thanks.

keloyang:
@bergwolf @jodh-intel @jshachm thanks for your review，PTAL

sboeuf:
@jodh-intel please take another look and merge if you're fine with this change !

jodh-intel:
lgtm

PR #253, opened 2018-05-25
grpc: signal frozen containers

### LINKED ISSUE #252: remove paused containers takes ~10 seconds
# Description of problem

```
$ time bash -c "docker run --name foo -dti debian bash && docker pause foo && docker rm -f foo"
6a16ba8b7a937fcedaa555a1ae3d852c986bdadb654e573063d13e6cbc7fb6cf
foo
foo

real	0m12.092s
user	0m0.168s
sys	0m0.033s
```

# Expected result

remove paused container shouldn't take too much time

devimc:
Allow frozen containers receive signals in order to avoid runtime
waits for shim to finish.

fixes #252

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -656,7 +656,8 @@ func (a *agentGRPC) SignalProcess(ctx context.Context, req *pb.SignalProcessRequ
 	// processes inside the container.
 	// If the process is the container process, let's use the container
 	// API for that.
-	if req.ExecId == "" {
+	// Frozen processes are thawed when `all` is true, allowing them to receive and process signals.
+	if req.ExecId == "" || status == libcontainer.Paused {
 		return emptyResp, ctr.container.Signal(signal, true)
 	} else if ctr.initProcess.id == req.ExecId {
 		return emptyResp, ctr.container.Signal(signal, false)


grpc.go
```diff
@@ -656,7 +656,10 @@ func (a *agentGRPC) SignalProcess(ctx context.Context, req *pb.SignalProcessRequ
 	// processes inside the container.
 	// If the process is the container process, let's use the container
 	// API for that.
-	if req.ExecId == "" {
+	// Frozen processes don't receive signals, but we'll do an exception
+	// because of we don't know why docker containers can, see https://github.com/moby/moby/issues/37137.
```
sboeuf: The change looks fine to me, but I really don't like this kind of comment.
The simple explanation is that all processes have to be thawed before we can signal them, and that's why we need to make sure we use `allProcesses` flag as `true` for this specific use case.

devimc:
@sboeuf done

jshachm:
LGTM good catch~

jodh-intel:
lgtm.

However, please could you point to tests we have to test sending signals to frozen containers? Or, raise an issue in the appropriate repo (referencing this PR) so we don't forget this.

devimc:
@jodh-intel done
https://github.com/kata-containers/agent/issues/254

PR #258, opened 2018-05-31
Fix update cpus - cpuset

### LINKED ISSUE #256: vCPUs not detected if cpuset fails before update the number of vCPUs
# Description of problem

```
└─ $ docker run --rm --name foo -dti debian bash
917b7fb254364fe8f4d905a18ebde4813f67c311fd9562298e885a253dc6c115
└─ $ docker update --cpuset-cpus "0-4" foo
Error response from daemon: Cannot update container 917b7fb254364fe8f4d905a18ebde4813f67c311fd9562298e885a253dc6c115: /usr/bin/kata-runtime did not terminate sucessfully: rpc error: code = Unknown desc = failed to write 0-4 to cpuset.cpus: write /sys/fs/cgroup/cpuset/docker/917b7fb254364fe8f4d905a18ebde4813f67c311fd9562298e885a253dc6c115/cpuset.cpus: invalid argument
: unknown
└─ $ docker update --cpus 4 foo
foo
└─ $ docker attach foo
root@917b7fb25436:/# nproc
1
```

also the communication with the container is closed, I found these messages in the journal

```
2018/05/30 19:28:22 [ERR] yamux: keepalive failed: i/o deadline reached
"agent grpc server quits" error="keepalive timeout"
"agent grpc channel teardown failed" error="close /dev/vport0p1: file already closed"
```

# Expected result
5

# Actual result
1

### LINKED ISSUE #257: Could not update twice cpuset
# Description of problem

```
docker run --rm --name foo -dti debian bash;
docker update --cpuset-cpus "0-4" foo;
docker update --cpus 4 foo;
docker update --cpuset-cpus "0-5" foo;
docker update --cpuset-cpus "0-1" foo; 
docker update --cpus 7 foo; 
docker update --cpuset-cpus "0-6" foo;
docker exec foo nproc
```

should print 7

# Expected result
7

# Actual result
2

devimc:
Don't modify container's config on update since if container.Set
fails, rollback won't be possible.
fixes #256

All the cpuset cgroup sub-directories MUST BE updated before setting the
new configuration, in that way `container.Set` won't fail and the expected
number of vCPUs will be available in the container.
fixes #257


grpc.go
```diff
@@ -848,43 +848,48 @@ func (a *agentGRPC) UpdateContainer(ctx context.Context, req *pb.UpdateContainer
 		return emptyResp, err
 	}
 
-	config := c.container.Config()
```
sboeuf: `c.container.Config()` returns a copy of type `configs.Config`. Why do you need to duplicate this copy ?
I would assume that if something goes wrong during `c.container.Set(config)`, this is libcontainer's responsibility to handle a proper rollback.
The issue is not very clear to me, could you elaborate ?

grpc.go
```diff
@@ -182,7 +220,7 @@ func (a *agentGRPC) onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
 	connectedCpus := strings.Trim(string(cpus), "\t\n ")
 	agentLog.WithField("range-of-vcpus", connectedCpus).Debug("connecting vCPUs")
 
-	mapCgroupPaths := make(map[string]bool)
+	cookies := make(cookie)
```
sboeuf: `make(cookie)` ? Are you implementing a bakery process here :smiley:

grpc.go
```diff
@@ -155,6 +157,42 @@ func onlineMemResources() error {
 	return err
 }
 
+// updates container's cpuset cgroups visiting each sub-directory in cgroupPath and writing
+// newCpuset in the cpuset.cpus file, cookies are used for performance reasons in order to
+// don't update a cgroup twice.
+func updateContainerCpuset(cgroupPath string, newCpuset string, cookies cookie) error {
```
sboeuf: A test would be nice for this function.

grpc.go
devimc: `c.container.Config()` returns a copy of the non-pointer elements in `configs.Config`, unfortunately [Cgroup](https://github.com/opencontainers/runc/blob/master/libcontainer/configs/config.go#L128) and [Resources](https://github.com/opencontainers/runc/blob/master/libcontainer/configs/cgroup_linux.go#L31) are pointers, if one of them is modified (like currently we do) , if `c.container.Set` fails it won't be able to rollback because of the container cgroup was modified, see https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L205

I'll add more comments in the code

devimc:
@jodh-intel @sboeuf changes applied

jodh-intel:
Coo-kie! Om nom nom nom...

lgtm

PR #266, opened 2018-06-14
network: Setup localhost when running as init

### LINKED ISSUE #265: Bring up localhost when running as init
If the agent is running as PID 1 it needs to bring up the localhost network interface.

See: https://github.com/kata-containers/runtime/issues/375

jodh-intel:
When the agent is running as PID 1 it needs to setup the localhost
network interface. If not done, not only will "lo" not be available, but
the netlink library will cause the agent to segfault as it implicitly
assumes `lo` is "up".

Also added in extra parameter validation and error handling improvements.

Fixes #265.

### DIFF
--- a/network.go
+++ b/network.go
@@ -60,8 +60,16 @@ func linkByHwAddr(netHandle *netlink.Handle, hwAddr string) (netlink.Link, error
 	}
 
 	for _, link := range links {
+		if link == nil {
+			continue
+		}
+
 		lAttrs := link.Attrs()
 		if lAttrs == nil {
+			continue
+		}
+
+		if lAttrs.HardwareAddr == nil {
 			continue
 		}
 
--- a/network.go
+++ b/network.go
@@ -9,6 +9,7 @@ package main
 import (
 	"fmt"
 	"net"
+	"os"
 	"reflect"
 	"sync"
 
@@ -564,4 +565,20 @@ func (s *sandbox) removeNetwork() error {
 	}
 
 	return nil
+}
+
+// Bring up localhost network interface.
+func (s *sandbox) handleLocalhost() error {
+	// If not running as the init daemon, there is nothing to do as the
+	// localhost interface will already exist.
+	if os.Getpid() != 1 {
+		return nil
+	}
+
+	lo, err := netlink.LinkByName("lo")
+	if err != nil {
+		return err
+	}
+
+	return netlink.LinkSetUp(lo)
 }
--- a/network.go
+++ b/network.go
@@ -278,6 +278,34 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) (*pb.Interface, 
 // Routes //
 ////////////
 
+func (s *sandbox) deleteRoutes(netHandle *netlink.Handle) error {
+	if netHandle == nil {
+		return errNoHandle
+	}
+
+	initRouteList, err := netHandle.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return err
+	}
+	for _, initRoute := range initRouteList {
+		// don't delete routes associated with lo:
+		link, _ := netHandle.LinkByIndex(initRoute.LinkIndex)
+		if link.Attrs().Name == "lo" || link.Attrs().Name == "::1" {
+			continue
+		}
+
+		err = netHandle.RouteDel(&initRoute)
+		if err != nil {
+			//If there was an error deleting some of the initial routes,
+			//return the error and the current routes on the system via
+			//the defer function
+			return err
+		}
+	}
+
+	return nil
+}
+
 //updateRoutes will take requestedRoutes and create netlink routes, with a goal of creating a final
 // state which matches the requested routes.  In doing this, preesxisting non-loopback routes will be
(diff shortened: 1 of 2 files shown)


jodh-intel:
Ping @liwei, @bergwolf.

grahamwhaley:
CI gocyclo error:
```
/home/travis/gopath/src/github.com/kata-containers/agent/protocols/mockserver
network.go:333::warning: cyclomatic complexity 16 of function (*sandbox).updateRoutes() is high (> 15) (gocyclo)
```

jodh-intel:
Branch updated.

Given the unlikelihood of `lo` not being available, I don't think it's worth pursing an upstream patch at this point. I suspect it might be viewed as "user error" anyhow since it's not unreasonable to assume `lo` is available on any system.

network.go
```diff
@@ -38,17 +47,33 @@ type network struct {
...
+
 	links, err := netHandle.LinkList()
 	if err != nil {
 		return nil, err
 	}
 
 	for _, link := range links {
+		if link == nil {
+			continue
+		}
```
WeiZhang555: links is fetched from `netHandle.LinkList()`, is there any situation that link can be nil? 
Or this is just for assurance?

network.go
```diff
@@ -99,6 +135,10 @@ func updateLink(netHandle *netlink.Handle, link netlink.Link, iface *pb.Interfac
 }
 
 func (s *sandbox) addInterface(netHandle *netlink.Handle, iface *pb.Interface) (resultingIfc *pb.Interface, err error) {
+	if iface == nil {
+		return nil, errNoIF
+	}
+
```
WeiZhang555: I like this change :+1:

network.go
```diff
@@ -185,7 +225,7 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *pb.Interface
 	defer s.network.ifacesLock.Unlock()
 
 	if iface == nil {
-		return nil, grpcStatus.Errorf(codes.InvalidArgument, "Provided interface is nil")
+		return nil, errNoIF
```
WeiZhang555: Maybe this is also better to move outside above the lock? So we can avoid an unnecessary lock

WeiZhang555:
crio test failed randomly, I think it's unralated

```
not ok 7 ctr logging [tty=true]
# (in test file ctr.bats, line 243)
#   `grep --binary -P "^[^\n]+ stdout F here is some output\x0d$" "$logpath"' failed
# 0
```

network.go
jodh-intel: Just for assurance. I could remove this check but it benign I think.

network.go
jodh-intel: Good catch! Branch updated.

bergwolf:
Jenkins ubuntu 17-10 failed on exit code checking but that is not related to this PR.
```
• Failure [60.114 seconds]
run
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-17-10-PR/go/src/github.com/kata-containers/tests/integration/docker/run_test.go:25
  container with docker
  /tmp/jenkins/workspace/kata-containers-agent-ubuntu-17-10-PR/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    with 'exit 123' as workload [It]
    /tmp/jenkins/workspace/kata-containers-agent-ubuntu-17-10-PR/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected
        <int>: 123
    to equal
        <int>: -1
```

Given that ci-centos and ci-fedora both passed. Let's merge this!

grahamwhaley:
Just a passing note - the metrics CI is not live on github yet (still have some stability problems being ironed out), but whilst checking failures, I did spot that this PR failed the metrics memory footprint size as it bumped the KSM memory footprint from ~46.5M to ~53M (which went pass my ~51M metrics check high water mark ;-) ). Interestingly, the non-KSM footprint looks to have stayed about the same.

bergwolf:
@grahamwhaley The extra lo setup must have dirtied more memory and thus bumped the KSM memory footprint. Is the ~51 high water mark an experience value or a target that we aim to achieve at?

jodh-intel:
I think we'll need more info on that increase as we can't really do without `lo` :)

grahamwhaley:
@jodh-intel Indeed, I think the change is unavoidable, and we probably have no alternatives - so, yes, this is mostly for information - I'm not expecting anything to get changed :-)
@bergwolf The high watermark is currently set as roughly 'max+10%', where max is the largest value I've seen in 10 or so runs. The low watermark is set similarly (min-10%).
Sadly the metrics CI system (at least the build slaves) do not have access to historical data, so for the 'quick PR CI' check, we hard code a +/- range we expect the results to fall within. The longer term plan is to have a separate system that does collect the historical data and can then do some smarter trend analysis on the results.

The values are also build machine dependent to some extent. The memory footprint is now less machine dependant I think since we no long use vCPUs==physCPUs as the default for the VMs (we now use vCPUs==1 as default I believe, so apprx. the same on all machines).

PR #271, opened 2018-06-22
logging: Fix handling of non-string fields when PID 1

### LINKED ISSUE #270: Fix logging of non-string fields when running as PID 1
Currently, the agent assumes (incorrectly) that all `logrus.Fields` are in fact `string` values when logging the startup announcement. This results in fields like the PID being displayed as:

```
pid='\\x01'
```
... and this upsets the [log-parser](https://github.com/kata-containers/tests/tree/master/cmd/log-parser).

jodh-intel:
Convert all non-string types into strings prior to logging to avoid
attempting to display non-string types with the `%q` format verb.

Fixes #270.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -521,6 +521,23 @@ func getAnnounceFields() (logrus.Fields, error) {
 	}, nil
 }
 
+// formatFields converts logrus Fields (containing arbitrary types) into a string slice.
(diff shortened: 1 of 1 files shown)


jodh-intel:
Another victim of the crio failures. This is unrelated to this PR so I vote we force-merge. Opinions?

Issue #278, opened 2018-06-26
agent hangs using kernel 4.14.36+


jcvenegas:
# Description of problem

When we try to use kata with the latest kata 4.14.x. The container creation hangs at `create` command.

# Expected result

The  agent can be created with a new kernel

# Actual result

The kata containers master fails to create kata. This does not happend to kata 1.0.0

---

Looking to the cause of the issue grpc request `CreateSandbox` never finish. This is due to the agent in early versions create a nspath using `rand` golang module https://github.com/kata-containers/agent/blob/master/namespace.go#L52.

`rand`  is failing not returning due to the  low-entropy. This is related to https://github.com/kata-containers/agent/blob/master/namespace.go#L52 as pointed in https://github.com/linuxkit/linuxkit/issues/3032. 

For this issue I think we should split in two:

- Remove random dependency at container creation. The ns-paths are only created once so we can change it for a known path.
- Add a test to make sure rand (/dev/random) works in a container. 
- Improve rand inside a container (use VirtIORNG or rngd if needed).

``` 
level=info msg="[\x1b[0;32m  OK  \x1b[0m] Started Kata Containers Agent.\n" name=kata-proxy pid=27048 source=agent
 level=info msg="[\x1b[0;32m  OK  \x1b[0m] Reached target Kata Containers Agent Target.\n" name=kata-proxy pid=27048 source=agent
 level=info msg="time=\"2018-06-26T15:43:42.964796872Z\" level=info msg=announce device-handlers=\"blk,scsi\" name=kata-agent pid=101 source=agent storage-handlers=\"ephemeral,9p,blk,scsi\" system-memory=\"2045296 kB\" version=1.0.0-f06eb20527f01097f9185a65a6f77e1ed5d2d360\n" name=kata-proxy pid=27048 source=agent
 level=info msg="time=\"2018-06-26T15:43:42.986904769Z\" level=info msg=\"Enable grpc tracing\" name=kata-agent pid=101 source=agent\n" name=kata-proxy pid=27048 source=agent
 level=info msg="time=\"2018-06-26T15:43:43.007601883Z\" level=info msg=\"agent grpc server starts\" name=kata-agent pid=101 source=agent\n" name=kata-proxy pid=27048 source=agent
 level=info msg="time=\"2018-06-26T15:43:43.012611116Z\" level=debug msg=\"New serial channel event\" events=4 name=kata-agent pid=101 source=agent\n" name=kata-proxy pid=27048 source=agent
[...]

PR #279, opened 2018-06-26
namespace: do not create path using rand.

jcvenegas:
Using random char generation at early boot may slow the process due to
low entropy.

Also from kernel 4.14.36 rand call hangs due to low-entropy. Lets
remove dependency on rand at early sandbox creation.

Fixes: #278

### DIFF
--- a/namespace.go
+++ b/namespace.go
@@ -7,7 +7,6 @@
 package main
 
 import (
-	"crypto/rand"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


nitkon:
@jcvenegas: Thanks for reporting this issue. I have faced this issue multiple times in the past few days but thought it was some local config issue.

jcvenegas:
hit error: https://github.com/kata-containers/tests/issues/361

jodh-intel:
Good catch @jcvenegas!

lgtm

I am curious to know why we originally created a randomly-named ns though (which seems perfectly reasonable).

sboeuf:
I have retriggered the build again, I really want to merge this asap so that we can move forward with the latest kernel 4.14.51. In the meantime, we have to figure out how to give more entropy to the VM.

jodh-intel:
Couple of thoughts:

- http://www.issihosts.com/haveged/
- https://github.com/dustinkirkland/pollen

PR #284, opened 2018-06-29
grpc: Don't close the terminal on CloseStdin() call

### LINKED ISSUE #283: grpc: Don't close the terminal when CloseStdin is called
`CloseStdin()` is supposed to close only the standard input, which helps the process terminating in some cases. But for the case where we're using a terminal, this terminal will be closed by the agent once the `WaitProcess()` will complete, otherwise it could prevent some output from the terminal to reach the shim.

sboeuf:
CloseStdin() is supposed to close only the standard input, which
helps the process to terminate in some cases. But for the case where
we are using a terminal, it should be closed by the agent once the
WaitProcess() will complete. Otherwise, closing it before the end of
the process could prevent some output from the terminal to reach the
shim.

Fixes #283

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -1089,14 +1089,13 @@ func (a *agentGRPC) CloseStdin(ctx context.Context, req *pb.CloseStdinRequest) (
 		return emptyResp, err
 	}
 
-	var file *os.File
-	if proc.termMaster != nil {
-		file = proc.termMaster
-	} else {
-		file = proc.stdin
+	// If stdin is nil, which can be the case when using a terminal,
+	// there is nothing to do.
+	if proc.stdin == nil {
+		return emptyResp, nil
 	}
 
-	if err := file.Close(); err != nil {
+	if err := proc.stdin.Close(); err != nil {
 		return emptyResp, err
 	}
 
--- a/grpc.go
+++ b/grpc.go
@@ -1044,6 +1044,16 @@ func (a *agentGRPC) WriteStdin(ctx context.Context, req *pb.WriteStreamRequest) 
 		return &pb.WriteStreamResponse{}, err
 	}
 
+	proc.RLock()
+	defer proc.RUnlock()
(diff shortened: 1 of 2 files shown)


sboeuf:
@jodh-intel @amshinde @bergwolf @grahamwhaley PTAL

jodh-intel:
lgtm

bergwolf:
@sboeuf I think we should deny `WriteStdin` as well in such case.

sboeuf:
@bergwolf what do you mean ? Do you mean that we should ignore any `WriteStdin()` after `CloseStdin()` got invoked ?

bergwolf:
@sboeuf yes

sboeuf:
@bergwolf yes I can definitely add a flag to ignore WriteStdin() if stdin has been closed previously !

sboeuf:
@bergwolf PR updated, PTAL !

devimc:
lgtm

grpc.go
```diff
@@ -1069,14 +1079,18 @@ func (a *agentGRPC) CloseStdin(ctx context.Context, req *pb.CloseStdinRequest) (
...
+		return emptyResp, nil
 	}
 
-	if err := file.Close(); err != nil {
+	proc.Lock()
+	defer proc.Unlock()
+
+	proc.stdinClosed = true
+
+	if err := proc.stdin.Close(); err != nil {
```
bergwolf: Only set stdinClosed to true when proc.stdin.Close() succeeds? Otherwise we might end up failing `CloseStdin()` and also start to ignore `WriteStdin()`. It's protected by the mutex so it's safe to just move `proc.stdinClosed = true` to after `stdin.Close()`.

sboeuf: I have thought about it and I was not sure, but that's fair, let's set the flag only is this succeeded.

sboeuf:
@bergwolf PR updated!

PR #286, opened 2018-06-29
namespace: Add check to make sure PID namespace is not received

### LINKED ISSUE #285: Pid namespace should be cleared by the runtime. Add a check to verify this in the agent.

amshinde:
The runtime should clear the PID namespace from the config
before sending it to the agent. Add a check in the agent to make
sure PID namespace is not recieved in the configuration.
Information about PID namespace is conveyed by means of a flag instead.

Fixes #285

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -521,16 +521,8 @@ func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container) {
 }
 
 func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (resp *gpb.Empty, err error) {
-	if a.sandbox.running == false {
-		return emptyResp, grpcStatus.Error(codes.FailedPrecondition, "Sandbox not started, impossible to run a new container")
-	}
-
-	if _, err = a.sandbox.getContainer(req.ContainerId); err == nil {
-		return emptyResp, grpcStatus.Errorf(codes.AlreadyExists, "Container %s already exists, impossible to create", req.ContainerId)
-	}
-
-	if a.pidNsExists(req.OCI) {
-		return emptyResp, grpcStatus.Errorf(codes.FailedPrecondition, "Unexpected PID namespace received, should have been cleared out", req.ContainerId)
+	if err := a.createContainerChecks(req); err != nil {
+		return emptyResp, err
 	}
 
 	// re-scan PCI bus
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

grpc.go
```diff
@@ -434,12 +434,13 @@ func (a *agentGRPC) postExecProcess(ctr *container, proc *process) error {
...
 // sense to rely on the namespace path provided by the host since namespaces
 // are different inside the guest.
-func (a *agentGRPC) updateContainerConfigNamespaces(config *configs.Config, ctr *container) {
+func (a *agentGRPC) updateContainerConfigNamespaces(config *configs.Config, ctr *container) error {
 	var ipcNs, utsNs bool
 
 	for idx, ns := range config.Namespaces {
-		// TODO: NEW_PID should be cleared out by the runtime. Add a strict check
-		// for this to return an error if NEWPID is found once the runtime PR lands.
+		if ns.Type == configs.NEWPID {
```
grahamwhaley: I'm slightly in two minds here. We know we are not going to use the PID namespace in the agent don't we, so is this really an error - does it actually affect any functionality? I wonder if:
- We could make it a warning, not an error
- We could move the check to the runtime?, as adding it here grows the agent (sure, only a fraction), inside the container - and really we'd like to try and keep our footprint overhead as small as we can.

So, nothing wrong with the patch, and lgtm - I just thought I'd ask if it's wholly necessary and in the best place to do the check?

jodh-intel: Ah - your ack hasn't applied @grahamwhaley as it's not at start of line - was that intended?

grahamwhaley: It was :-) - I thought we'd discuss a touch first.

bergwolf: I would prefer to do the checking at the beginning of the grpc request handler so that we do not have to revert anything if the sanity check fails.

amshinde: @bergwolf Fair enough, I'll add this check early on.

amshinde:
@bergwolf @grahamwhaley PTAL

grahamwhaley:
From 16.04 CI, I think you may have hit the gocyclo limit:
```
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-16-04-PR/go/src/github.com/kata-containers/agent/protocols/mockserver
grpc.go:533::error: no formatting directive in Errorf call (vet)
grpc.go:523::warning: cyclomatic complexity 16 of function (*agentGRPC).CreateContainer() is high (> 15) (gocyclo)
```

devimc:
lgtm

bergwolf:
lgtm

bergwolf:
> Stderr: E: Unable to determine file size for fd 11 - fstat (2: No such file or directory)

The CI failure does look like the guest kernel is missing the 9pfs hack. Let's ignore it and merge since fedora and ubuntu CI all passed.

Issue #290, opened 2018-07-04
Add sandbox level reference counter for sandbox storages


harche:
# Description of problem

Right now there is no sandbox level reference counter for the storages. 

As @bergwolf pointed out in this [comment](https://github.com/kata-containers/runtime/pull/458#issuecomment-402195377)

Sandbox should add a [reference counter](https://github.com/kata-containers/agent/blob/master/agent.go#L83) to the storage that is used at sandbox level. 

I will send a PR to add that reference counter.

PR #293, opened 2018-07-05
[WIP] [DNM] agent: Add sandbox level reference counter for sandbox storage

harche:
This PR adds sandbox level storage reference counter. The
first consumer of this reference counter is ephemeral type
storage

Fixes: #290

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>


harche:
@bergwolf I have added a `sandboxStorage` struct in agent.go . For now, I have added only the path of the storage in the storage struct, let me know if you think we should have more fields in there. I  thought of adding storage mount but sandbox struct already has `mounts` field. 

@WeiZhang555 ping.

agent.go
```diff
@@ -200,6 +205,21 @@ func (c *container) getProcess(execID string) (*process, error) {
 	return proc, nil
 }
 
+func (s *sandbox) getSandboxStorages() []sandboxStorage {
+	return s.storages
+}
+
+func (s *sandbox) setSandboxStorages(sb sandboxStorage) {
```
bergwolf: nit: `sb` can be confusing with `sandbox` :(

agent.go
```diff
@@ -80,6 +80,10 @@ type container struct {
 	useSandboxPidNs bool
 }
 
+type sandboxStorage struct {
+	sourcePath string
```
bergwolf: What is missing is a reference counter for each sandboxStorage. And since we want to use `sourcePath` to identify a sandbox storage, it's better to make it a map. Something like
```
type sandbox struct {
        <snip>
        storages map[string]sandboxStorage
}
```
And 
```
type sandboxStorage struct {
        refcount int
        [possibly other fields]
}
```
The container setup code path need to be modified such that if `sourcePath` is already mounted, do not mount it again. And the container cleanup code path also needs to be modified to make sure we do not unmount a `sourcePath` when there is still someone using it (IOW, refcount != 0).

agent.go
amshinde: yes, better to rename this, something like `sbStorage`

amshinde: I would also rename the method itself to something like appendSandboxStorage or even setSandboxStorage to make it clear we are adding this storage to existing list.

agent.go
```diff
@@ -200,6 +206,39 @@ func (c *container) getProcess(execID string) (*process, error) {
 	return proc, nil
 }
 
+func (s *sandbox) setSandboxStorage(path string) {
```
bergwolf: It's better to return a `bool` here to indicate that if the storage is a new one or an old one. Given such information, the caller would be able to tell if it needs to mount to storage or not, instead of calling `GetRefCount()` again.

agent.go
```diff
@@ -220,6 +259,12 @@ func (s *sandbox) setContainer(id string, ctr *container) {
 
 func (s *sandbox) deleteContainer(id string) {
 	s.Lock()
+	for k := range s.storages {
+		sbs := s.storages[k]
+		sbs.Lock()
```
bergwolf: This would deadlock since `DecrementRefCount` takes the lock again. And is it intentional that sbs is not removed from the map?

And what about `container.removeContainer()`?

agent.go
```diff
@@ -80,6 +80,11 @@ type container struct {
 	useSandboxPidNs bool
 }
 
+type sandboxStorage struct {
+	sync.RWMutex
+	refcount int
```
sboeuf: s/refcount/refCount/

mount.go
```diff
@@ -197,7 +197,12 @@ var storageHandlerList = map[string]storageHandler{
 }
 
 func ephemeralStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
-	if _, err := os.Stat(storage.MountPoint); os.IsNotExist(err) {
+	s.Lock()
+	defer s.Unlock()
+	mountTmpfs := s.setSandboxStorage(storage.MountPoint)
```
sboeuf: Use a clearer name here. I would use `existStorage` instead of `mountTmpfs`, and then for the `if` statement, I would use the negative form:
```go
if !existStorage {
...
}
```

It looks clearer to me that we're trying to create the path/storage if it does not exist.

agent.go
```diff
@@ -200,6 +206,67 @@ func (c *container) getProcess(execID string) (*process, error) {
...
+}
+
+// unSetSandboxStorage will decrement the sandbox storage
+// reference counter. If there aren't any containers using
+// that sandbox storage, this method will also remove the
+// storage from the sandbox.
+//
+// It's assumed that caller is calling this method after
+// acquiring a lock on sandbox.
+func (s *sandbox) unSetSandboxStorage(path string) {
```
sboeuf: We should return an error through this function.
```go
func (s *sandbox) unSetSandboxStorage(path string) error {
```

agent.go
```diff
@@ -200,6 +206,67 @@ func (c *container) getProcess(execID string) (*process, error) {
...
+	if sbs, ok := s.storages[path]; !ok {
+		agentLog.WithField("sandbox-storage-path", path).Error("sandbox storage not found")
+	} else {
+		sbs.Lock()
```
sboeuf: You should call into `defer sbs.Unlock()` here, this way, we can add an error return without worrying about unlocking properly.

agent.go
```diff
@@ -200,6 +206,67 @@ func (c *container) getProcess(execID string) (*process, error) {
...
+		sbs.refcount--
+
+		if sbs.refcount < 1 {
+			sbs.Unlock()
+			err := removeMounts([]string{path})
+			if err != nil {
+				agentLog.WithField("sandbox-storage-path", path).Error("Unable to unmount sandbox storage")
+			}
+			err = os.RemoveAll(path)
```
sboeuf: This does not look good as it's not mirroring what `setSandboxStorage()` did. What I mean is, during `setSandboxStorage()` we simply add the refcounting, but through `unSetSandboxStorage()`, we take care of unmounting and removing the path. This does not feel right.

agent.go
```diff
@@ -200,6 +206,99 @@ func (c *container) getProcess(execID string) (*process, error) {
...
 
+// setSandboxStorage sets the sandbox level reference
+// counter for the sandbox storage.
+// This method also returns a boolean to let
+// callers know if the storage already existed or not.
+// It will return true if storage is new.
+//
+// It's assumed that caller is calling this method after
+// acquiring a lock on sandbox.
+func (s *sandbox) setSandboxStorage(path string) bool {
```
bergwolf: It seems both `setSandboxStorage` and `unSetSandboxStorage` are called under the protection of the sandbox lock. The same lock is used to protect the storage map IIUC. Then there is no need for a sbs lock just to protect the refcount.

agent.go
```diff
@@ -200,6 +206,99 @@ func (c *container) getProcess(execID string) (*process, error) {
...
+
+// unsetAndRemoveSandboxStorage unsets the storage from sandbox
+// and if there are no containers using this storage it will
+// remove it from the sandbox
+//
+// It's assumed that caller is calling this method after
+// acquiring a lock on sandbox.
+func (s *sandbox) unsetAndRemoveSandboxStorage(path string) error {
+	if sbs, ok := s.storages[path]; ok {
+		if err := s.unSetSandboxStorage(path); err != nil {
```
bergwolf: `unSetSandboxStorage` already removes the sbs from the map. So I'd prefer it to return a boolean to indicate that the sbs needs to be cleaned up. Then you do not need to check the refcount again below.

agent.go
sboeuf: There's also `GetRefCount()` being called by `unsetAndRemoveSandboxStorage()` but again, this is protected by a sandbox lock at some point.
I agree the sbs lock might be a bit too much here, but we need to be careful with this in case we rework the locking to a finer granularity in the future.

agent.go
sboeuf: Yes that's a good point, the boolean will save some code, but I would like to keep the error return too, so I think the best thing to do is:
```go
func (s *sandbox) unSetSandboxStorage(path string) (bool, error) {
```

harche:
@bergwolf @sboeuf @amshinde Thanks for your comments.

sboeuf:
@harche it'd be great if you could fix the code coverage here as there's a 0.75% drop.

agent.go
```diff
@@ -200,6 +205,100 @@ func (c *container) getProcess(execID string) (*process, error) {
...
+		return nil
+	}
+	return grpcStatus.Errorf(codes.NotFound, "Sandbox storage with path %s not found", path)
+}
+
+// GetRefCount returns the reference count
+//
+// It's assumed that caller is calling this method after
+// acquiring a lock on sandbox.
+func (sbs *sandboxStorage) GetRefCount() int {
```
bergwolf: please remove `GetRefCount()` since it is not used any more.

bergwolf:
@harche Please fix code coverage and remove `GetRefCount()`. Then this PR is ready to be merged. Thanks for your work!

bergwolf:
LGTM! Thanks @harche!

grpc.go
```diff
@@ -993,13 +993,31 @@ func (a *agentGRPC) RemoveContainer(ctx context.Context, req *pb.RemoveContainer
...
 		go func() {
 			if err := ctr.removeContainer(); err != nil {
 				done <- err
 			}
 
+			// Find the sandbox storage used by this container
+			for _, path := range ctr.mounts {
+				if _, ok := a.sandbox.storages[path]; ok {
+					if err := a.sandbox.unsetAndRemoveSandboxStorage(path); err != nil {
+						done <- err
```
WeiZhang555: Here you need to return and close(done), the `done` is only a channel without buffer which can block if there's no receiver. so the go routine could leak here.

grpc.go
```diff
@@ -993,13 +993,31 @@ func (a *agentGRPC) RemoveContainer(ctx context.Context, req *pb.RemoveContainer
...
+				if err := a.sandbox.unsetAndRemoveSandboxStorage(path); err != nil {
+					return emptyResp, err
+				}
+			}
+		}
 	} else {
 		done := make(chan error)
 		go func() {
 			if err := ctr.removeContainer(); err != nil {
 				done <- err
```
WeiZhang555: You need to return and close(done) here, or if error happens in later part, the done channel will block and go routine will leak

harche: Not part of my changes but the channel is getting closed a few lines below.

grpc.go
harche: channel is getting closed here, https://github.com/kata-containers/agent/pull/293/files#diff-727ccd4217b7f335e689be03bfb69e9bR1021

harche: Makes sense. Thanks.

grpc.go
harche: Makes sense. Thanks.

sboeuf:
@WeiZhang555 please take a look at the updated PR, and merge if you're fine with it ;)

grpc.go
```diff
@@ -993,14 +993,34 @@ func (a *agentGRPC) RemoveContainer(ctx context.Context, req *pb.RemoveContainer
...
+				close(done)
+				return
 			}
 
-			close(done)
```
WeiZhang555: This is still necessary, and should be moved to last line of this goroutine function.

Because if you don't close the done channel, the 
```
select {
 		case err := <-done:
 			if err != nil {
 				return emptyResp, err
 			}
 		case <-time.After(time.Duration(req.Timeout) * time.Second):
 			return emptyResp, grpcStatus.Errorf(codes.DeadlineExceeded, "Timeout reached after %ds", timeout)
 		}
 	}
```

`<-done` will always block if no error happens, and this function will always time out

WeiZhang555:
LGTM

The CI itself has some problem, I'm not sure if we should merge this without CI's guard.
Let's wait a moment to see if CI can resume.

WeiZhang555:
This PR is ready to merge.

Anyone please merge this once CI is back (or without CI if we're confident enough)

grahamwhaley:
Let's see if we can get the CI fixed - it is actively being looked at over on https://github.com/kata-containers/tests/pull/515.

egernst:
rebased the changes onto latest in order to restart the CI.  Fingers crossed...

PR #289, opened 2018-07-03
logging: Redirect yamux warnings/errors to logger

jodh-intel:
Capture yamux messages to the logger.

Fixes #288.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #295, opened 2018-07-06
agent: disable yamux keep alive

### LINKED ISSUE #294: yamux: keepalive failed: i/o deadline reached for proxybuiltin
# Description of problem

When kata is integrated into hyper, which uses the proxybuiltin mode. 
Once do  hyperctl pause <pod id>, which will pause the sandbox. 
After pause about 30s, builtin proxy will complain of "yamux: keepalive failed: i/o deadline reached"

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

lifupan:
We don't know how much time a sandbox can be paused,
hence connection write timeout should be disabled to
don't close the connection while the sandbox is paused.

The same issue has been fixed in kata-proxy, for katabuiltin proxy,
it also needs this fix.

fixes #294

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -147,7 +147,10 @@ func agentDialer(addr *url.URL, enableYamux bool) dialer {
 		}()
 
 		var session *yamux.Session
-		session, err = yamux.Client(conn, nil)
+		sessionConfig := yamux.DefaultConfig()
+		// Disable keepAlive since we don't know how much time a container can be paused
+		sessionConfig.EnableKeepAlive = false
+		session, err = yamux.Client(conn, sessionConfig)
 		if err != nil {
 			return nil, err
 		}


protocols/client/client.go
```diff
@@ -147,7 +147,10 @@ func agentDialer(addr *url.URL, enableYamux bool) dialer {
 		}()
 
 		var session *yamux.Session
-		session, err = yamux.Client(conn, nil)
+		sessionConfig := yamux.DefaultConfig()
+		// Disable keepAlive since we don't know how much time a container can be paused
+		sessionConfig.EnableKeepAlive = false
+		session, err = yamux.Client(conn, sessionConfig)
```
jodh-intel: As you are now passing config to the client api, you might want to set `sessionConfig.LogOutput`, similar to #289, so that any yamux warnings+errors are logged to the agent log?

lifupan: Hi, Jodh

Here the yamux client is running in the buildin proxy, instead of agent context, and there is no logger instance here.

jodh-intel: ah - ok. So do you not get any agent logs then?

bergwolf: No, this is consumed by the runtime instead.

bergwolf:
LGTM!

devimc:
lgtm

Just waiting for the CI

amshinde:
CI seems to be failing due to error downloading `alpine` docker image. Have restarted the CI jobs.

bergwolf:
Ignore `jenkins-ci-ubuntu-16-04-initrd` failure since that is expected to pass only after we merge #297

PR #299, opened 2018-07-17
channel: support vsocks hotplug

### LINKED ISSUE #298: Support vsocks
If there are no serial ports in the VM, the agent must wait for a vsock to be hot plugged

see https://github.com/kata-containers/runtime/issues/383

devimc:
If no serial ports are found in the VM, the agent must wait
for a vsock to be hot plugged.
To avoid clashes with other processes that are using vsocks
and their context IDs, the runtime will try to get a unique context
ID and then hot plug a vsock in the VM, this process is repeated
until a vsock is hot plugged correctly.

see https://github.com/kata-containers/runtime/issues/383

fixes #298


devimc:
@kata-containers/agent  please take a look

sboeuf:
@devimc will take a look today!

channel.go
```diff
@@ -29,27 +35,34 @@ type channel interface {
 }
 
 func newChannel() (channel, error) {
```
sboeuf: It'd be nice to add a verbose comment on top of this function. We need to explicitly say that the agent will try by default to connect to a serial port, and that a failure doing so will not result in an agent failure, but instead it will fallback to the mechanism of waiting for a VSock socket to be created, and this will be gated by a timeout.

channel.go
```diff
@@ -21,6 +22,11 @@ import (
 	grpcStatus "google.golang.org/grpc/status"
 )
 
+const (
+	vSockExistMaxTries = 20
+	vSockExistWaitTime = 50 * time.Millisecond
```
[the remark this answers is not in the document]
devimc: agree, I think 10 seconds should be enough

devimc:

sboeuf:
@devimc thanks I'll take another look

channel.go
```diff
@@ -199,15 +220,15 @@ func isAFVSockSupported() (bool, error) {
 }
 
 func vSockPathExist() (bool, error) {
```
sboeuf: The way you modified this function, there's no need for `vSockPathExist()` to return an error anymore. It should be a simple boolean.

devimc: you're right

devimc:
@sboeuf  change applied, thanks

bergwolf:
@devmic Instead of waiting, I'm thinking a different approach how this can be implemented:
1. always start with serial channel service
2. add a new gRPC call that tells the agent to restart the gRPC service with either serial or vsock channel
3. the runtime can hotplug vsock device and then call the new gRPC call to switch from serial connection to vsock connection

It has a few benefits:
1. no more sleep waiting on a device agent does not know if it would show up
2. vsock switch is explicitly instructed by the runtime, so more controlabe
3. works well with the vm factory model
4. when we ever add live migration support, the switch gRPC would make it possible since vsock CIDs are not maintained across machines

WDTY?

sboeuf:
@bergwolf currently @devimc is re-working this PR to get the vsock socket being coldplugged directly into the VM so that the agent won't have to wait.

I thought about your proposal, but there is also one drawback. We're going to leave one serial port accessible to anything inside the VM that might escape the container, and then access the host through this security hole.

bergwolf:
@sboeuf 
> We're going to leave one serial port accessible to anything inside the VM that might escape the container, and then access the host through this security hole.

If runtime uses serial connection, the port is already there. An escaped guest process can always access it by hacking the serial connection. If runtime does not use serial connection, there is nothing reading the serial port on the host side and thus escaped guest process cannot access anything on the host through it.

bergwolf:
@sboeuf And cold-plugged vsock device will make it impossible to migrate the guest locally or to other hosts. So it won't work with vm factory, nor live migration if we ever want that in future.

sboeuf:
@bergwolf what about having the serial port hotplugged so that it can be unplugged in case vsock is used?

bergwolf:
@sboeuf Yes, that can work as well. Still I think we will need a new gRPC call to tell the agent to switch between serial and vsock connections. I think that is how it can work with vm factory and live migration. Do you have other ideas?

sboeuf:
@bergwolf the new grpc seems like a good idea since I don't want to introduce another protocol to get the information about the communication type.
But I am wondering if this will not introduce some latency in case of vsock (need to restart the grpc server)
```
agent waits for serial port ==> agent connects to serial port ==> agent start the grpc server on this connection ==> if the connection is vsock, stop the grpc server ==> wait for the vsock socket to appear ==> connect to the vsock socket ==> start the grpc server on the new connection.
```

bergwolf:
@sboeuf Or we can make agent watch for both serial port and vsock device add events and use whatever comes first. Then the switch only happens when runtime asks agent to.

bergwolf:
And runtime should know what to use since it is the one hotpluging the device to the guest.

sboeuf:
@bergwolf oh so we would use the new grpc call only for vm templating, right ? The regular case would rely on the ability of the agent to detect which communication way has been chosen?

bergwolf:
@sboeuf Yes, I think so. And in future when we want to start migration, we should ask agent to switch to serial port before doing migration. But right now, vm factory would be the only consumer of the new gRPC call. Actually we can proceed with the hotplug idea and add the new gRPC call in a separate PR if you want.

sboeuf:
@bergwolf I think we should proceed in several steps here:
- @devimc should finish his work on coldplugging the vsock (from the runtime) since this is the easiest thing to do. This means the agent would simply have to check for serial of vsock. I think this is an important first step to validate that vsock works properly as expected with the whole chain/stack.
- As a second step, we make both serial and vsock sockets hotpluggable from the runtime codebase. This will need a corresponding PR on the agent side to detect if serial or vsock get hotplugged (could be based on uevent or simply checking some known paths).
- In the meantime, we can add the new gRPC function (something like `changeCommChannel()`), with the appropriate agent implementation which will take care of stopping the gRPC server, close the current connection and connect to the new socket/port in order to re-spin a new gRPC server. The agent should have a timeout to check for this new socket/port for a certain amount of time since the hotplug might take some time.
- As a third step, we would use leverage the `changeCommChannel()` from the runtime in case of VM templating.

And I think we should open an issue to describe the proposal as we don't want to lose this conversation once this PR will be merged.

bergwolf:
@sboeuf Yes, I created kata-containers/runtime/issues/532 to track the planning we discussed for vsock support.

sboeuf:
@bergwolf ah cool, thanks for opening this one, so that we can keep track of the steps/progress !

devimc:
@bergwolf  @sboeuf  please take a look

PR #301, opened 2018-07-18
protocols/client: close yamux session when closing the stream

bergwolf:
For yamux builtin client, we should close the yamux session when closing
the stream otherwise the session keeps open and can confuse the server.


sboeuf:
@amshinde ptal

PR #303, opened 2018-07-18
agent: do not quit on grpc serve errors

bergwolf:
If any errors happens at that time, we should just tear down the
connection and start over again.


bergwolf:
@sboeuf yes, I get the situation in case of vm factory w/ builtin proxy.

Issue #308, opened 2018-07-27
Do not rely on kernel parameters for sandbox id


bergwolf:
CreateSandbox() is the point we know which sandbox id will use this guest. If we rely on the kernel parameters for it, we have to put sandbox id in qemu args and it breaks vm factory.

PR #309, opened 2018-07-27
agent: set sandbox id log field in CreateSandbox

bergwolf:
This effectively reverts 04457e38("logging: Add sandbox field").
CreateSandbox() is the point we know which sandbox id will use
this guest. If we rely on the kernel command line for it, we have
to put sandbox id in qemu args and it breaks vm factory.

Fixes: #308

cc @jodh-intel


agent.go
```diff
@@ -87,6 +87,7 @@ type sandbox struct {
 	sync.RWMutex
 
 	id              string
+	hostname        string
```
jodh-intel: These hostname changes seem unrelated? Maybe put them on a different commit with an explanation?

protocols/grpc/agent.proto
```diff
@@ -251,12 +251,13 @@ message CreateSandboxRequest {
 	string hostname = 1;
 	repeated string dns = 2;
 	repeated Storage storages = 3;
-	
+
```
jodh-intel: Nit: Whitespace damage.

protocols/grpc/agent.proto
```diff
@@ -251,12 +251,13 @@ message CreateSandboxRequest {
...
 	// This field means that a pause process needs to be created by the
-	// agent. This pid namespace of the pause process will be treated as 
+	// agent. This pid namespace of the pause process will be treated as
 	// a shared pid namespace. All containers created will join this shared
 	// pid namespace.
 	bool sandbox_pidns = 4;
+	string sandbox_id = 5;
```
jodh-intel: It might be worth adding a comment here explaining that we don't need this member in the other sandbox and network calls (as listed in https://github.com/kata-containers/tests/issues/453#issuecomment-401339364) since we implicitly require `CreateSandbox()` to be called before any other sandbox/network calls :)

agent.go
bergwolf: It's related... The original code saves hostname to sandbox.id. I moved it to sandbox.hostname.

protocols/grpc/agent.proto
bergwolf: I deleted the extra whitespaces because it was unintentionally added by someone. I don't think we need white space there.

protocols/grpc/agent.proto
bergwolf: Sure, I'll add some text about it.

agent.go
jodh-intel: Ah - right, thanks.

protocols/grpc/agent.proto
jodh-intel: ack

jodh-intel:
lgtm!

sboeuf:
@chavafg @GabyCT Swarm CI is not happy, any idea here ?

GabyCT:
@sboeuf , the problem is not Swarm, the test is detecting that some processes are running before running Swarm
```
Found unexpected /usr/libexec/kata-containers/kata-proxy present
```

sboeuf:
@GabyCT I see, but this `checkprocess()` function is called during the teardown, right ?
So I would assume something got wrong during the swarm test that makes the shim still being around ?

GabyCT:
@sboeuf , using the following configuration
```
kata-runtime : 1.1.0
commit : cfbc974
OCI specs: 1.0.1
kata-proxy version 1.1.0-c416c9fc2ff8389c201075f0f26bfabd08a3b140
kata-shim version 1.1.0-6ddca0334d0e8dee6e1e00f0c120fde4b8c1c176
kata-agent version 1.1.0-c6ac153f0c4c03b475d2b65bb73427cd33459bf7
```

I ran `swarm` tests in a loop for 15 times and there were not issues. However, if I run first the `kubernetes` tests and then `swarm`, I am obtaining the failure that some processes were left behind.

bergwolf:
As agreed we do not trust `codecov/patch`. So merge this since all other checks have passed. I'll see how to disable `codecov/patch` globally (e.g., via [codecov team yaml](https://docs.codecov.io/v4.3.0/docs/codecov-yaml#section-team-yaml)).

PR #310, opened 2018-07-31
codecov: remove codecov.yml

bergwolf:
And let's rely on the team yaml setup in codecov.io.

I've put a default team yaml at https://codecov.io/account/gh/kata-containers/yaml and according to the docs, it is supposed to be working with all repos in the kata-containers team. Let's see how it plays out.


jodh-intel:
lgtm

bergwolf:
`codecov/patch` is now gone so the team yaml is working.

`codecov/project` reports `-0.73%` but it actually says the touched file is untracked. I've updated the commit with a real issue number fix.

Issue #311, opened 2018-07-31
Support random number generator reseeding


bergwolf:
Because vm templating would share the initial kernel memory among guests, we need to reseed CRNG when new guests are created from template. Add ReseedRandomDev rpc to do it.

PR #312, opened 2018-07-31
agent: add ReseedRandomDev rpc

bergwolf:
It is used to reseed the guest random number generator with provided
seeds.

Fixes: #311


random.go
```diff
@@ -0,0 +1,49 @@
...
+	rngDevMode = 0666
+
+	// include/uapi/linux/random.h
+	// RNDRESEEDCRNG   _IO( 'R', 0x07 )
+	_IOC_RNDRESEEDCRNG = 0x5207
+)
+
+func reseedRNG(data []byte) error {
+	if len(data) == 0 {
+		return nil
```
jodh-intel: The only legitimate reason for calling this function would be to call it with  some random data so I wonder if it's worth adding a log call for this scenario as a warning that something went wrong?

jodh-intel:
lgtm

random.go
bergwolf: hmm, good point! I'll fix it up. Thanks!

random.go
```diff
@@ -0,0 +1,49 @@
...
+	rngDev = "/dev/random"
+
+	// include/uapi/linux/random.h
+	// RNDRESEEDCRNG   _IO( 'R', 0x07 )
+	_ioc_RNDRESEEDCRNG = 0x5207
+)
+
+func reseedRNG(data []byte) error {
+	if len(data) == 0 {
+		return fmt.Errorf("missing entropy data")
```
jodh-intel: Nice - even better :)

jodh-intel:
lgtm

random.go
```diff
@@ -0,0 +1,63 @@
+//
```
devimc: @jodh-intel  I wonder if this file should be called `random_linux.go`
see https://github.com/kata-containers/runtime/issues/379

jodh-intel: Good call!

/cc @raravena80

random.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+const (
+	rngDev = "/dev/random"
```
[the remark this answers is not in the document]
bergwolf: urandom is using the same rng source, so we only need to reseed one of them.

random.go
```diff
@@ -0,0 +1,63 @@
...
+
+func reseedRNG(data []byte) error {
+	if len(data) == 0 {
+		return fmt.Errorf("missing entropy data")
+	}
+
+	// Write entropy
+	f, err := os.OpenFile(rngDev, os.O_WRONLY, 0)
+	if err != nil {
+		agentLog.WithError(err).Warn("Could not open rng device")
```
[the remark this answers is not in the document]
bergwolf: random number generator.

random.go
bergwolf: kata-containers/runtime#379 is about building `kata-runtime` on Mac. Are we planning on building `kata-agent` on other platforms as well? IMO kata-agent relies heavily on Linux specific features (sysfs, device tree, cgroups, namespaces etc.), and we have not done any thing/design to make it cross platform AFAICT.

jodh-intel: @bergwolf - I agree that it could be an uphill struggle to port the agent. But I still think it makes sense to put obviously Linux-specific code in a `_linux.go` file. It's not a lot of effort and might make life easier for @raravena80 :)

bergwolf: @jodh-intel My point is that kata-agent on other platforms may never happen, unless we are willing to rewrite kata-agent completely (in which case hyperstart might be better suited for the task), -- e.g., libcontainer does not work on platforms other than linux. I can rename this to random_linux.go but I wanted to remind @raravena80 (or anyone who wants to look at the direction) that it is a dead end.

bergwolf:
@jodh-intel @devimc random.go renamed. PTAL.

sboeuf:
@devimc @egernst PTAL

jodh-intel:
Yep - protocol buffer change so we need another ack...

bergwolf:
I'm surprised that @sboeuf is not enlisted in the protocol buffer change approvers!

jodh-intel:
@bergwolf - well, that pullapprove config was added to ensure we get *other* (non-agent) folk to "eyeball" protocol buffer changes due to their potential impact. We still need the agent team (of which @sboeuf is a maintainer member) to provide 2 acks of course:

- https://github.com/kata-containers/agent/blob/master/.pullapprove.yml#L45..L56

Feel free to raise a PR to tweak this though :)

PR #316, opened 2018-08-03
agent: add ListInterfaces and ListRoutes rpc

### LINKED ISSUE #315: network: Add list APIs to show the actual network info in a container
According to the discussion in https://github.com/kata-containers/runtime/issues/113, two APIs are missing.
```
rpc ListInterfaces(ListInterfacesRequest) returns(Interfaces);
rpc ListRoutes(ListRoutesRequest) returns (Routes);
```
Need to implement them to support network hotplug.

And how about these two?
```
rpc AddInterface(AddInterfaceRequest) returns(Interface);
rpc RemoveInterface(RemoveInterfaceRequest) returns (Interface);
```
Delete them or keep them for compatibility？

caoruidong:
These are part of network hotplug and used to show actual network
info in the sandbox.

Fixes #315

Signed-off-by: Ruidong Cao <caoruidong@huawei.com>


network.go
```diff
@@ -322,6 +322,32 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) (*pb.Interface,
...
+	}
+
+	var interfaces pb.Interfaces
+	for _, link := range links {
+		ifc, err := getInterface(netHandle, link)
+		if err != nil {
+			agentLog.WithError(err).Error("listInterfaces() failed")
+			return &interfaces, err
+		}
+		interfaces.Interfaces = append(interfaces.Interfaces, ifc)
```
sboeuf: I don't really like the syntax, but this is because of the grpc protocol...
Any way we could define a list of interfaces through grpc, so that it ends up like:
```go
type Interfaces []*Interface
```
instead of
```go
type Interfaces struct {
	Interfaces []*Interface `protobuf:"bytes,1,rep,name=Interfaces" json:"Interfaces,omitempty"`
}
```

network_test.go
```diff
@@ -185,3 +185,50 @@ func TestUpdateRoutes(t *testing.T) {
...
+			MTU:          1500,
+			TxQLen:       -1,
+			Name:         "ifc-name",
+			HardwareAddr: macAddr,
+		},
+	}
+	netHandle.LinkAdd(link)
+	netHandle.LinkSetUp(link)
+
+	resultingIfc, err := s.updateInterface(netHandle, &ifc)
```
sboeuf: This whole test might run on a dev computer, and I am wondering if this is safe to do such a thing?
I mean we're basically updating the network and then checking. We should run this test into its own network namespace to check the function easily.
Basically, we enter the netns (need to be safe about it, meaning no go routine as they will jump to a different netns), then we create a few links, and then we run `listInterfaces()` to check the list of interfaces we actually found.

network.go
caoruidong: Yes. It is from grpc. Struct is needed to serialize . I just want to align with route.
```
 message Routes {
 	repeated Route Routes = 1;
 }
```
If grpc allows return a slice instead of a struct, I can make the change.

network.go
```diff
@@ -322,6 +322,32 @@ func getInterface(netHandle *netlink.Handle, link netlink.Link) (*pb.Interface,
...
+	links, err := netHandle.LinkList()
+	if err != nil {
+		return nil, err
+	}
+
+	var interfaces pb.Interfaces
+	for _, link := range links {
+		ifc, err := getInterface(netHandle, link)
+		if err != nil {
+			agentLog.WithError(err).Error("listInterfaces() failed")
```
jodh-intel: I'd be tempted to add a `.WithField("link", link)` here.

network_test.go
```diff
@@ -185,3 +185,50 @@ func TestUpdateRoutes(t *testing.T) {
...
+	assert.Nil(t, err, "Unexpected update interface failure: %v", err)
```
jodh-intel: Nit: I find it's much cleaner to avoid having to pass `t` to all the assert function using the following idiom:

```
assert := assert.New(t)
   :

assert.Nil(err, "Unexpected update interface failure: %v", err)
```

network.go
caoruidong: Done

network_test.go
caoruidong: Yes, definitely. Fixed

jodh-intel:
Thanks @caoruidong.

lgtm

network_test.go
```diff
@@ -185,3 +185,31 @@ func TestUpdateRoutes(t *testing.T) {
...
+
+	assert := assert.New(t)
+	s := sandbox{}
+
+	netHandle, _ := netlink.NewHandle()
+	defer netHandle.Delete()
+
+	results, err := s.listRoutes(netHandle)
+	assert.Nil(err, "Expected to list interfaces successfully")
+	assert.NotZero(len(results.Routes), "Expected to list all interfaces, but found nothing")
```
sboeuf: s/`Expected to list all interfaces`/`Expected to list all routes`/

network_test.go
```diff
@@ -185,3 +185,31 @@ func TestUpdateRoutes(t *testing.T) {
 	assert.True(t, reflect.DeepEqual(results.Routes[0], testRoutes.Routes[1]),
 		"Interface created didn't match: got %+v, expecting %+v", results.Routes[0], testRoutes.Routes[1])
 }
+
+func TestListInterfaces(t *testing.T) {
```
sboeuf: Those test are just checking the underlying network. How do you know something is going to be there? We should have stronger tests as I suggested earlier, by creating a netns and by creating some fake interfaces in it. Then, we can run the function listInterfaces() and it should find exactly what has been created.

Issue #320, opened 2018-08-10
client: speed up agent dialer


bergwolf:
The grpc dialer has a default 1 second backoff on the first connection failure. And it kills kata containers creation speed because we try to connect before knowing for certain that the agent is up and serving.

grahamwhaley:
Ooh, nice find. Be great to see how that changes our timings :-)

bergwolf:
@grahamwhaley As mentioned in #321, it speeds up kata container creation by around 0.5 second (w/o vm templating) or 1 second (w/ vm templating).

jon:
IIUC the agent listens for an incoming vsock connection and the runtime connects, retrying until the agent is actually booted and listening.

I may have misunderstood the code, but to me, this feels backwards. The runtime could be "always listening" and inject (via kernel command line) an address for the agent to connect to. This would allow the agent to connect at the first possible moment to an already-listening endpoint. That would seem to minimize the time to connect while also avoiding retries. If a connection in the other direction is desired, the agent could make a single RPC over this channel "Connect to me on <address>" at which point the runtime could make a connection that it expected to succeed on the first try.

Alternatively, override the Dialer on both sides: have the runtime "Dial" by listening, and have the Agent "Listen" by dialing. It would be a strange abuse of Go's Dialer/Listener semantics, but it might accomplish the above with a minimum need for retries and trying to race to speed up connection initiation.

sboeuf:
@jon 
>The runtime could be "always listening" and inject (via kernel command line) an address for the agent to connect to. This would allow the agent to connect at the first possible moment to an already-listening endpoint.

We're gonna move soon to the hotplug of the serial/vsock communication channel into the VM, after this one got started, meaning we have no way to tell the agent when this is going to be ready. And at the time the kernel parameter is passed, the socket might not have been hotplugged yet.

PR #321, opened 2018-08-10
client: bypass grpc dialer backoff strategy

bergwolf:
This would bypass the grpc dialer backoff strategy and handle dial timeout
internally. Because we do not have a large number of concurrent
dialers, it is not reasonable to have such aggressive backoffs which
would kill kata containers boot up speed. For more information, see
https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md

Fixes: #320

This would speed up kata containers creation by about 0.5 second for non vm templating case and by about 1 second when vm templating is enabled.

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -179,9 +179,13 @@ func agentDialer(addr *url.URL, enableYamux bool) dialer {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
Good catch!

lgtm

PR #323, opened 2018-08-16
protocols: increase dial timeout

### LINKED ISSUE #324: Failed to check if grpc server is working: context deadline exceeded: unknown when using vsocks
see https://github.com/kata-containers/runtime/issues/576

### LINKED ISSUE #576: Fix capitalization in docs
Fix incorrect capitalization.

devimc:
Run containers that use vsock as communication channel inside VMs
(nested environments) randomly fails with following error:

```
Stderr: docker: Error response from daemon: OCI runtime create failed:
Failed to check if grpc server is working: context deadline exceeded: unknown.
```

Sometimes the connection with the container's vsock is slow because of
kata-runtime disables modern (don't rely on fast MMIO) on some devices
including vsocks. This issue can be fixed by increasing the dial timeout.

fixes #324

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -28,7 +28,7 @@ const (
 	vsockSocketScheme = "vsock"
 )
 
-var defaultDialTimeout = 5 * time.Second
+var defaultDialTimeout = 15 * time.Second
 
 // AgentClient is an agent gRPC client connection wrapper for agentgrpc.AgentServiceClient
 type AgentClient struct {


chavafg:
This will allow us to enable vsock testing on the CI (where the nested environment tend to be slow sometimes), so 

lgtm

jodh-intel:
lgtm.

I wonder if we need a way to modify this value at runtime though.

bergwolf:
@chavafg Have we verified that this fix the random vsock test failures in nested CI environment?

devimc:
@jodh-intel 
> I wonder if we need a way to modify this value at runtime though.

yes, we need to re-vendor kata-agent in the kata-runtime

jodh-intel:
Sorry, what I meant more was whether we need a way to set that timeout dynamically (API call, environment variable, etc).

chavafg:
@bergwolf not in a CI environment, but on a local VM where the issue was also reproduced. With this fix, the issue didn't appear anymore.

bergwolf:
@chavafg Thanks for the clarification!

LGTM

Issue #322, opened 2018-08-13
Add opentracing support


jodh-intel:
See https://github.com/kata-containers/kata-containers/issues/27.

jodh-intel:
The plan for adding opentracing support to the agent is currently not clear: nominally, a "host system" will run a Jaeger agent (which the application will send spans to via UDP). However, we really do not want to have to add another component to the [guest images](https://github.com/kata-containers/osbuilder).

Further, adding an extra component to the image when the agent is running as PID 1 will require custom code to launch the agent at VM startup... by the agent. Clearly, that isn't going to work unless the Jaeger client could "buffer" the spans in readiness for sending to the Jaeger agent when it eventually started.

The current thinking is to mandate a Jaeger agent in the host environment only (the host the runtime runs on). We could use a virtio-serial port or vsock to allow the agent to send spans out of the VM back to the agent running outside the VM. However, vsock can't really be the only solution given some users run old host kernels that do not have vsock support (and they cannot change the host kernel).

The current thinking is that we either offer multiple approaches to extracting the spans from the agent inside the VM (yuck), or we:

- Use a virtio-serial port.
- Configure Jaeger to log to shared memory and share in a block of shared memory from the host into the VM for the agent to write spans into.
  However, we need to know what happens if the block is too small / becomes "full" - does Jaeger support circular logging buffers, etc?

/cc @sameo, @mcastelino, @sboeuf, @bergwolf, @grahamwhaley.

grahamwhaley:
Just a brief (slightly devils advocate) thought.... given:
 - we have enabled vsock in the runtime now
 - pretty much everything that happens in the agent is driven from the runtime, so 'one end' of the conversation is already being traced....

then I'd not be unhappy if we said, at least initially, that we would implement agent tracing via VSOCK, and if you don't have VSOCK available on your system, then, well, sorry, but you still have 90+% of the useful information just by looking at the runtime traces.

We could always put non-VSOCK as a secondary item, and if a member really really needed it, they are free to work with us to PR it ;-)

jodh-intel:
I guess that could work, but it would be good to get input from some power users about whether this approach would work for them.

P.S. You missed an opportunity to use :imp: :smile:

jodh-intel:
We're long overdue for an update on this issue, so brace yourself as I attempt to summarise the current situation...

# Current serial behaviour

## Startup

[...]

PR #328, opened 2018-08-20
client: Add context parameter and enable tracing support

### LINKED ISSUE #327: Add gRPC tracing for clients
Add a `context.Context` parameter to the client `NewAgentClient()` API and enable gRPC tracing if the specified context contains an opentracing span.

jodh-intel:
Add a `context.Context` parameter to the client `NewAgentClient()` API and
enable gRPC tracing if the specified context contains an opentracing
span.

Fixes #327.

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -14,8 +14,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 	"github.com/hashicorp/yamux"
 	"github.com/mdlayher/vsock"
+	opentracing "github.com/opentracing/opentracing-go"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	grpcStatus "google.golang.org/grpc/status"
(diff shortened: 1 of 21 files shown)


jodh-intel:
Note that this change is somewhat "invasive" since it adds an opentracing dependency to the client. However, that's the direction we want to move in anyway (see https://github.com/kata-containers/kata-containers/issues/27).

Note also that the gRPC *server* the agent runs is not currently traced - that will be dealt with on #322.

devimc:
CI is not happy

```
protocols/client/client_test.go:96:35:warning: ...
```

jodh-intel:
Branch updated.

protocols/client/client.go
```diff
@@ -54,14 +56,28 @@ type dialer func(string, time.Duration) (net.Conn, error)
 //   - unix://<unix socket path>
 //   - vsock://<cid>:<port>
 //   - <unix socket path>
-func NewAgentClient(sock string, enableYamux bool) (*AgentClient, error) {
+func NewAgentClient(ctx context.Context, sock string, enableYamux bool) (*AgentClient, error) {
```
devimc: why this change? the context is always `Background()`

jodh-intel: We want to be able to pass the context the runtime creates to `virtcontainers/kata_agent.go` because the runtime's context will have the trace information stored inside it so with this change, when the runtime is executed, we'll get gRPC traces for the client (runtime) side.

devimc: make sense, thanks

Issue #150, opened 2018-02-21
Document all protocol elements


jodh-intel:
We should document *all* Messages and all message members for the `*.proto` files in:

- https://github.com/kata-containers/agent/tree/master/protocols/grpc

This will:

- improve comprehension, thus making it easier to understand and review the protocol
  (which might indirectly help ensure the protocol is as safe as possible).
- help avoid issues like #147.

Additional to the description, we should pay particular attention to `string` fields due to their increased flexibility (and thus increased ability to confuse):

- What is the format of each string member?
- Can the value be `""`? If so, what does it mean?
- Are there any other "special" values (like the terrible golang `net.IP` `"<nil>"` value)?

Example where we have ambiguity:

```
message StringUser {
	string uid = 1;
	string gid = 2;
	repeated string additionalGids = 3;
}
```

Are those member values numbers encoded as strings (`"3"`, `"1000"`, `"0"`)? Or can they also (?) be specified as symbolics (`"adm"`, `"james"`, `"root"`)?

egernst:
@jodh-intel - PTAL and update issue w/ current relevance.  AFAIU we have protocol documented now, but perhaps not at the detail you are hoping for.  Please clarify.

/cc @amshinde

jodh-intel:
Most grpc `message` fields are now documented - we seem to have got into the groove of explaining all new types. Some of the older `message` types are still undocumented at the field level but are generally obvious. If there is confusion for these older ones, let's updated as that arises.

Closing.

PR #331, opened 2018-08-23
agent: add support for online memory and cpu separately

### LINKED ISSUE #332: add support for online memory only
In func OnlineCPUMem, cpu is always onlined, no matter whether we want to online it , which is not expected. Therefore we should only online memory when NbCpus in OnlineCPUMemRequest is
equal to zero.

clarecch:
agent: add support for online memory and cpu separately.

In func OnlineCPUMem, cpu is always onlined,
which is not expected. So we add "CpuOnly"
field to support separating memory and cpu online.

Fixes #332

Signed-off-by: Clare Chen <clare.chenhui@huawei.com>
Signed-off-by: Zichang Lin <linzichang@huawei.com>


clarecch:
This PR is related to https://github.com/kata-containers/runtime/pull/624

grpc.go
```diff
@@ -193,17 +193,19 @@ func updateContainerCpuset(cgroupPath string, newCpuset string, cookies cookie)
 }
 
 func (a *agentGRPC) onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
-	if req.NbCpus <= 0 {
+	if req.NbCpus < 0 {
```
bergwolf: Good catch!

jodh-intel:
Hi @clarecch - thanks for raising!

It might be worth changing the commit message a little. Previously -- as you have said -- CPU onlining was always attempted (even when zero cpus were specified). But also, it looks like because of that initial test, previously memory was only ever onlined when CPUs were also onlined? With this change, we online memory "always", not only when the number of cpus is zero I think?

It might even be clearer to change the code to online the memory first - we could then keep the original `req.NbCpus` test?

lgtm

devimc:
CI is not happy, @clarecch can you fix it?

```
grpc.go:196:5:warning: unsigned values are never < 0 (SA4003) (staticcheck)
```

clarecch:
@jodh-intel Yes, there are more troubles than I think. 
1.  req.NbCpus is unsigned int , which is never below zero. This should be corrected.
2.  As you have said, now memory online become always, which is also need to be corrected. How about seprate cpu online with memory online, or add some flag to req so that we can judge whether cpu or memory online is needed.

clarecch:
@devimc I will fix it as soon as possible.

linzichang:
@jodh-intel @devimc I think there're three solutions:
1. The modify in this PR. It's very simple. But maybe not so graceful.
2. Adding member to "req *pb.OnlineCPUMemRequest" to decide online cpu, mem or both.
3. Separating onlineCPUMem to two function, onlineCPU and onlineMem.

jodh-intel:
Yep - I think option 3 is the "cleanest", but let's get some more input from others....

devimc:
yep, I think hyper guys wrote this protocol
cc @bergwolf @laijs @gnawux

bergwolf:
@devimc @jodh-intel @linzichang @clarecch We put onlinecpu and onlinememory into one call since they are mostly called together, and we save one grpc call when they actually are. Also we did want to only use a small set of grpc APIs to keep it simple.

And if we split/rename the onlinecpumem API, it will break grpc backward compatibility. So I would suggest we go with option 2 and name the new field `CpuOnly` so that it defaults to `false` and old client can still work properly.

bergwolf:
@clarecch, I don't think we need an extra field. The new grpc onlinecpumem handler can:
1. NbCpus > 0: online CPU
2. NbCpus == 0 && CpuOnly: return error
3. !CpuOnly: online memory

And all cases are covered. WDYT?

bergwolf:
@clarecch I think we can rely on `NbCpus>0` to say that we want to online CPU, and `CpuOnly=false` to say that we want to online memory, and do not specially care about the cpu/memory **only** cases. The combination of `NbCpus<=0 && CpuOnly == true` means we do not want to online cpu nor memory, in which case caller should not call the API in the first place, and we return error for it.

So to answer your question, `NbCpus == 0 && CpuOnly == false` means we only online memory. Maybe we should rename `CpuOnly` to `SkipMemory` to make it non-relevant to CPU ;). I wanted to make online memory the default behavior so I picked `CpuOnly` but it seems to cause more confusion than I expected.

jodh-intel:
Adding a `CpuOnly bool` wfm...

```diff
diff --git a/protocols/grpc/agent.proto b/protocols/grpc/agent.proto
index b930d94..865fe0f 100644
--- a/protocols/grpc/agent.proto
+++ b/protocols/grpc/agent.proto
@@ -330,11 +330,13 @@ message ListRoutesRequest {
 message OnlineCPUMemRequest {
        // Wait specifies if the caller waits for the agent to online all resources.
        // If true the agent returns once all resources have been connected, otherwise all
        // resources are connected asynchronously and the agent returns immediately.
        bool wait = 1;
 
        // NbCpus specifies the number of CPUs that were added and the agent has to online.
        uint32 nb_cpus = 2;
+
+       bool cpu_only = 3;
 }
```

... as long as we document in that file every possible scenario (as we have done on this issue) to maximise understanding.

Related: https://github.com/kata-containers/agent/issues/150.

linzichang:
I have no idea why the ubuntu-ci failed. May someone give a help?

grpc.go
```diff
@@ -193,21 +193,25 @@ func updateContainerCpuset(cgroupPath string, newCpuset string, cookies cookie)
 }
 
 func (a *agentGRPC) onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
-	if req.NbCpus <= 0 {
+	if req.NbCpus == 0 && req.CpuOnly {
```
bergwolf: What about `NbCpus < 0`? We should fail in such case.

linzichang: `NbCpus` is uint32. I think there is no need to consider `NbCpus <0`.

bergwolf: heh, good point! I thought it was int32 as we had `<=` there ;)

bergwolf:
CI failed with
```
not ok 2 Running within memory constraints
# (from function `waitForProcess' in file ../../.ci/lib.sh, line 161,
#  in test file k8s-memory.bats, line 47)
#   `waitForProcess "$wait_time" "$sleep_time" "$pod_status_cmd"' failed
```

clarecch:
@bergwolf I have no idea what happen when Ci run k8s-memory.bats and failed. Could you please help to find out the reason, if convenient?

caoruidong:
This PR https://github.com/kata-containers/runtime/pull/643 also fails on this test. But it just bumps version of golang. Kind of wired. I think there is something wrong with CI.

bergwolf:
LGTM!

bergwolf:
CI green now. It needs one more approval on the protocol change though.

caoruidong:
Hold on. @clarecch Do you generate pb.go file by proto-gen-go tool?

devimc:
good catch @caoruidong!   @clarecch  please run `hack/update-generated-agent-proto.sh` to generate the `pb.go` file

clarecch:
@devimc @caoruidong @bergwolf Commit is updated, and CI is green. PTAL！

bergwolf:
Thanks @clarecch ! Still need one more approval on the protocol change though. ping @kata-containers/agent

caoruidong:
LGTM

jodh-intel:
Thanks @clarecch!

lgtm

grpc.go
```diff
@@ -193,21 +193,25 @@ func updateContainerCpuset(cgroupPath string, newCpuset string, cookies cookie)
 }
 
 func (a *agentGRPC) onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
```
sboeuf: @linzichang @bergwolf 
For sure this patch works, but why don't we simply define two different functions for CPU and memory ?
This way, no need for specific flags, we would simply call the appropriate function:
- `onlineCPU`
- `onlineMemory`

jodh-intel: We did discuss this a bit above (https://github.com/kata-containers/agent/pull/331#issuecomment-415412330) and I agree it would be a lot clearer (but potentially slower as multiple calls might be required).

sboeuf: Yes I can see that!
@bergwolf I understand the concern about multiple calls, but I think it's definitely negligible compared to the whole stack.
And about the API compatibility breakage, we better break it now to make it right, as the project is still not completely stable.

PR #131, opened 2018-02-01
enable interruptible and multiple WaitProcess()

laijs:
In some cases, these might be multiple clients calls to WaitProcess(), this pr enables it.
and some other cases the clients might disconnect while calling WaitProcess(), we need to handle it too.

fixes #129 #130

PR #334, opened 2018-08-24
container: wait a bit before destroying non-stopped container

bergwolf:
Signal handling is async. If RemoveContainer is called right after
SignalProcess, it is possible that the container process is still
running. Let's wait a bit to see to give it some chance to exit
before calling destroy(). Otherwise destroy() might fail because
container process is still running.

cc @lifupan

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -188,11 +188,35 @@ func (c *container) deleteProcess(execID string) {
 	c.Unlock()
 }
 
+func (c *container) waitContainer() error {
+	for i := 0; i < 100; i++ {
+		if err := c.container.Signal(syscall.Signal(0), false); err != nil {
+			return nil
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	return fmt.Errorf("container is still running")
+}
+
 func (c *container) removeContainer() error {
+	s, err := c.container.Status()
+	if err != nil {
+		return err
+	}
+
+	// Wait a bit to see if container stops. This might happen if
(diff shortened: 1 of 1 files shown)


agent.go
```diff
@@ -188,11 +188,35 @@ func (c *container) deleteProcess(execID string) {
 	c.Unlock()
 }
 
+func (c *container) waitContainer() error {
+	for i := 0; i < 100; i++ {
+		if err := c.container.Signal(syscall.Signal(0), false); err != nil {
```
grahamwhaley: Nice poll ;-)

agent.go
```diff
@@ -188,11 +188,35 @@ func (c *container) deleteProcess(execID string) {
...
+	return fmt.Errorf("container is still running")
+}
+
 func (c *container) removeContainer() error {
+	s, err := c.container.Status()
+	if err != nil {
+		return err
+	}
+
+	// Wait a bit to see if container stops. This might happen if
```
devimc: well, 10 seconds is not a bit

bergwolf: Referencing [runc](https://github.com/opencontainers/runc/blob/master/delete.go#L18) :)

bergwolf:
@sboeuf After discussing with @lifupan, I think a better approach is to make `WaitProcess` reentrant so that runtime can call it multiple times, like @laijs did in #131. I'll spin sometime to revive that one.

sboeuf:
This PR can be closed as it's been superseded by another one.

Issue #337, opened 2018-08-28
wait on yamux session close


bergwolf:
Newer version of yamux library supports to wait on session close. We should use to to get a clean state otherwise yamux on serial is prone to reconnection errors.

PR #338, opened 2018-08-28
wait on yamux session close

bergwolf:
For both yamux client and server, we should wait on session close to get a clean state after close. Otherwise there might be inflight requests dirtying the serial channel and cause reconnection failures.

Fixes: #337

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -96,11 +96,11 @@
   revision = "8e809c8a86450a29b90dcc9efbf062d0fe6d9746"
 
 [[projects]]
-  digest = "1:73d3d2f8f2bcf510db08576eca6c1d2b87bcea348de26bf1386b291ad1b52296"
+  digest = "1:63e0b20cfa3fe456480edf93a7995f776afb610e49da8e3da04d8904472a44cc"
   name = "github.com/hashicorp/yamux"
   packages = ["."]
   pruneopts = "NUT"
-  revision = "f5742cb6b85602e7fa834e9d5d91a7d7fa850824"
+  revision = "cc6d2ea263b2471faabce371255777a365bf8306"
 
 [[projects]]
   digest = "1:0159dcdabe50788e5dcfb469521f8f8dcd362db3ab6b465b99a3d33a90726fc0"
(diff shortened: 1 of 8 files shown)


jodh-intel:
Hi @bergwolf - the changes look good - any possibility of adding a unit test for this?

Gopkg.lock
```diff
@@ -3,122 +3,95 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:86feeb6c372fd7984322882737b00f32d7e75c6b437f93d75f43db3195574dab"
```
devimc: seem like you are using an old version of `dep`

bergwolf:
@devimc @jodh-intel PR updated. PTAL.

jodh-intel:
Thanks @bergwolf.

lgtm

jodh-intel:
From 16.04 initrd CI run:

```
=== RUN   TestNewAgentClientWithYamux
coverage: 81.9% of statements
panic: test timed out after 10s
```

Restarting...

protocols/client/client.go
```diff
@@ -31,6 +32,7 @@ const (
 )
 
 var defaultDialTimeout = 15 * time.Second
+var defaultCloseTimeout = 5 * time.Second
```
devimc: have you tested this change with vsocks? see https://github.com/kata-containers/agent/issues/324 and https://github.com/kata-containers/agent/commit/93fe84e041b8895e68d268246a06c5a9a1395f3c

devimc: why you are reducing the timeout again?

bergwolf: This is close timeout not dial timeout. Why would a guest take more than 5 seconds to close a connection?

And it is only used by the yamux client. So only kata-runtime and kata-proxy will use it.

devimc: make sense, thanks

channel.go
```diff
@@ -200,11 +201,16 @@ func (c *serialChannel) listen() (net.Listener, error) {
...
 	}
+	c.waitCh = session.CloseChan()
 
 	return session, nil
 }
 
 func (c *serialChannel) teardown() error {
+	// wait for the session to be fully shutdown first
+	if c.waitCh != nil {
+		<-c.waitCh
```
sboeuf: You should use a go `select:` here, and timeout after some time, otherwise this might result into a deadlock.

bergwolf: hmm, good point. I'll get to it later today.

Issue #333, opened 2018-08-24
wait a bit in RemoveContainer


bergwolf:
@lifupan hit an issue that if RemoveContainer is called right after SignalProcess(SIGKILL), the container process might be still running and container `Destroy()` will fail.

sboeuf:
@bergwolf that's what the shim is here for. You should wait for the shim to return the exit code before you try to remove the container.

bergwolf:
@sboeuf This is for containerd shimv2. It uses buildin proxy and shim so there is no standalone shim process to wait on.

sboeuf:
@bergwolf sure I understand that, but there is a go routine waiting for this, right ?

bergwolf:
@sboeuf The problem is kata runtime `container.stop()`. It kills the container process and then removes the container w/o waiting. OTOH, we cannot wait there either because otherwise if there is another waiter, it cannot get the container exit status.

@lifupan I agree that we should move container cleanup codes to `container.delete()` but it won't help solving this issue, unless you require `container.wait()` being always called before `container.delete()` in which case `container.stop()` become meaningless since it does the same thing as `container.kill()`. And container state machine becomes complicated because `wait()` is required to get a container to transit to STOPPED state.

lifupan:
@bergwolf  Yes,  previously I thought we can require the container.wait() being always called before container.delete(). If you guys think this will complicated the state machine, I agree with you on the  approach to make WaitProcess reentrant you mentioned in https://github.com/kata-containers/agent/pull/334.

Issue #129, opened 2018-02-01
WaitProcess() should be able to called multi time from client(s)


laijs:
Those requests which were issued before the process exit
should be all fulfilled.

PR #339, opened 2018-08-28
agent: allow multiple waitProcess()

lifupan:
There will be multiple waitProcess() requests
issued before process exits, and all those requests
should be fulfilled.

Fixes: #129
Fixes: #333

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -768,17 +768,19 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
 		return &pb.WaitProcessResponse{}, err
 	}
 
-	defer func() {
+	defer proc.Do(func() {
 		proc.closePostExitFDs()
 		ctr.deleteProcess(proc.id)
-	}()
+	})
 
 	// Using helper function wait() to deal with the subreaper.
 	libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
-	exitCode, err := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
-	if err != nil {
-		return &pb.WaitProcessResponse{}, err
-	}
+	exitCode, _ := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
+	//refill the exitCodeCh with the exitcode which can be read out
+	//by another WaitProcess(). Since this channel isn't be closed,
+	//here the refill will always success and it will be free by GC
+	//once the process exits.
+	proc.exitCodeCh <- exitCode
 
 	return &pb.WaitProcessResponse{
 		Status: int32(exitCode),
--- a/agent.go
+++ b/agent.go
@@ -70,6 +70,7 @@ type process struct {
 	consoleSock *os.File
 	termMaster  *os.File
 	exitCodeCh  chan int
+	sync.Once
 	stdinClosed bool
 }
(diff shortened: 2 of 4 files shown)


jodh-intel:
Hi @lifupan - the changes make sense but it would be great if you could add some unit tests to assert the correct behaviour for:

- a single call to `waitProcess()`.
- multiple calls to `waitProcess()`.

lifupan:
Hi  @jodh-intel  - NP, I'll add the unit tests and resend this PR.

Thanks

Fupan

grpc.go
```diff
@@ -768,10 +770,16 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
 		return &pb.WaitProcessResponse{}, err
 	}
 
+	proc.exitCodeCh.RLock()
+
 	defer func() {
+		proc.exitCodeCh.RLock()
```
bergwolf: No need for the extra func() closure. Please just use `defer proc.exitCodeCh.RLock()`.

grpc.go
```diff
@@ -768,10 +770,16 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
+	}()
+
+	defer proc.Do(func() {
 		proc.closePostExitFDs()
```
bergwolf: Please also reset all process FDs in `closePostExitFDs` so that they are not closed multiple times.

grpc.go
```diff
@@ -780,6 +788,19 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
 		return &pb.WaitProcessResponse{}, err
 	}
 
+	select {
+	//the exitCodeCh is closed, do nothing
+	case <-proc.exitCodeCh.exitCodeCh:
```
bergwolf: One `WaitProcess` goroutine can pick up the `exitCode` filled by another `WaitProcess` goroutine and then the exit code is lost forever. All remaining waiting goroutines in `reaper.wait()` will be blocked forever.

It is a hard to test scenario because the check here happens quickly, but the race window is there.

grpc.go
lifupan: Got it.

Fupan

grpc.go
lifupan: No, the other WaitProcess will be blocked on the line of:
exitCode, err := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
 and the exitCode only will be picked up by it and once they come to the "select" block, 
the exitCode has been picked up there.

grpc.go
lifupan: Yes,  it has already been enclosed in this defer block.

grpc.go
bergwolf: Good point! Thanks for the explanation!

lifupan:
Hi @jodh-intel  
After discussed with @bergwolf , I decided to drop this complex fix which involved so many locks and syncs between goroutines and use a simple way to fix it. I'll resend this PR later.

jodh-intel:
Hi @lifupan - ok - thanks for the update! Do you want to close this PR then?

lifupan:
Hi @jodh-intel 
I can close this PR now and resend another one.

Fupan

PR #344, opened 2018-08-30
Multi wait process

lifupan:
There will be multiple waitProcess() requests
issued before process exits, and all those requests
should be fulfilled.

Fixes: #129
Fixes: #333

Signed-off-by: fupan lifupan@gmail.com

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -768,10 +768,10 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
 		return &pb.WaitProcessResponse{}, err
 	}
 
-	defer func() {
+	defer proc.Do(func() {
 		proc.closePostExitFDs()
 		ctr.deleteProcess(proc.id)
-	}()
+	})
 
 	// Using helper function wait() to deal with the subreaper.
 	libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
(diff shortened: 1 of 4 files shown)


grpc.go
```diff
@@ -768,17 +768,19 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
-	}()
+	})
 
 	// Using helper function wait() to deal with the subreaper.
 	libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
-	exitCode, err := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
-	if err != nil {
-		return &pb.WaitProcessResponse{}, err
-	}
+	exitCode, _ := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
```
bergwolf: Why is the error return ignored?

jodh-intel: Even though that might be technically correct *now*, I'd still check the result since if the wait function changes in the future, we might forget to update this part of the code.

lifupan: Yeah, I can return it back. : )

jodh-intel:
Thanks @lifupan - nice tests too! ;)

lgtm

grpc.go
```diff
@@ -768,17 +768,22 @@ func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
...
 	libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
 	exitCode, err := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
 	if err != nil {
 		return &pb.WaitProcessResponse{}, err
 	}
+	//refill the exitCodeCh with the exitcode which can be read out
+	//by another WaitProcess(). Since this channel isn't be closed,
+	//here the refill will always success and it will be free by GC
+	//once the process exits.
+	proc.exitCodeCh <- exitCode
```
sboeuf: I think that we should still close the go channel from `deleteProcess()`.

lifupan: Hi @sboeuf , we cannot close the exitCodeCh in deleteProcess(),  if it's closed in deleteProcess(), then the other waitProcess() which are issued before the process exited
will not get the exit code, and this close will also cause the write panic in the next waitProcess().

sboeuf: Ah yes, you're right.

PR #346, opened 2018-08-30
agent: write OCI spec to config.json in bundle path

eguzman3:
In order to comply with the OCI specification, the spec must be written to a file named 'config.json' located in the bundle directory.

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -524,6 +524,35 @@ func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container) {
 	}
 }
 
+func finishCreateContainer(a *agentGRPC, ctr *container, req *pb.CreateContainerRequest, config *configs.Config) (resp *gpb.Empty, err error) {
+	containerPath := filepath.Join("/tmp/libcontainer", a.sandbox.id)
+	factory, err := libcontainer.New(containerPath, libcontainer.Cgroupfs)
+	if err != nil {
+		return emptyResp, err
+	}
+
+	ctr.container, err = factory.Create(req.ContainerId, config)
+	if err != nil {
+		return emptyResp, err
+	}
+	ctr.config = *config
+
+	ctr.initProcess, err = buildProcess(req.OCI.Process, req.ExecId)
+	if err != nil {
+		return emptyResp, err
+	}
+
+	if err = a.execProcess(ctr, ctr.initProcess, true); err != nil {
+		return emptyResp, err
+	}
+
+	if err := a.updateSharedPidNs(ctr); err != nil {
+		return emptyResp, err
+	}
+
+	return emptyResp, a.postExecProcess(ctr, ctr.initProcess)
+}
+
 func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (resp *gpb.Empty, err error) {
 	if err := a.createContainerChecks(req); err != nil {
 		return emptyResp, err
@@ -579,6 +608,13 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	// Change cwd because libcontainer sets the bundle path to cwd
+	oldcwd, err := pb.ChangeToBundlePath(ociSpec)
+	if err != nil {
+		return emptyResp, err
+	}
+	defer os.Chdir(oldcwd)
+
 	// Convert the OCI specification into a libcontainer configuration.
 	config, err := specconv.CreateLibcontainerConfig(&specconv.CreateOpts{
 		CgroupName:   req.ContainerId,
@@ -590,38 +626,21 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	// In order to comply with the OCI specification, the spec must be
+	// written to a file named 'config.json' located in the bundle
+	// directory. https://github.com/opencontainers/runtime-spec
+	err = pb.WriteSpecToFile(ociSpec)
+	if err != nil {
+		return emptyResp, err
+	}
+
 	// Update libcontainer configuration for specific cases not handled
 	// by the specconv converter.
 	if err = a.updateContainerConfig(ociSpec, config, ctr); err != nil {
 		return emptyResp, err
 	}
 
-	containerPath := filepath.Join("/tmp/libcontainer", a.sandbox.id)
-	factory, err := libcontainer.New(containerPath, libcontainer.Cgroupfs)
-	if err != nil {
-		return emptyResp, err
-	}
-
-	ctr.container, err = factory.Create(req.ContainerId, config)
-	if err != nil {
-		return emptyResp, err
-	}
-	ctr.config = *config
-
-	ctr.initProcess, err = buildProcess(req.OCI.Process, req.ExecId)
-	if err != nil {
-		return emptyResp, err
-	}
-
-	if err = a.execProcess(ctr, ctr.initProcess, true); err != nil {
-		return emptyResp, err
-	}
-
-	if err := a.updateSharedPidNs(ctr); err != nil {
-		return emptyResp, err
-	}
-
-	return emptyResp, a.postExecProcess(ctr, ctr.initProcess)
+	return finishCreateContainer(a, ctr, req, config)
 }
 
 func (a *agentGRPC) createContainerChecks(req *pb.CreateContainerRequest) (err error) {
--- a/grpc_test.go
+++ b/grpc_test.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"reflect"
 	"testing"
@@ -532,4 +533,70 @@ func TestReseedRandomDev(t *testing.T) {
 	r, err := a.ReseedRandomDev(context.TODO(), req)
 	assert.NoError(err)
 	assert.Equal(r, emptyResp)
+}
+
+func TestChangeToBundlePath(t *testing.T) {
+	assert := assert.New(t)
+
+	originalCwd, err := os.Getwd()
+	assert.NoError(err)
+	defer os.Chdir(originalCwd)
+
+	bundlePath, err := ioutil.TempDir("", "bundle")
+	assert.NoError(err)
+	defer os.RemoveAll(bundlePath)
+
+	rootFsPath := path.Join(bundlePath, "rootfs")
+	err = os.Mkdir(rootFsPath, 0666)
+	assert.NoError(err)
+
+	spec := &specs.Spec{}
+	spec.Root = &specs.Root{
+		Path:     "",
+		Readonly: false,
+	}
+
+	_, err = pb.ChangeToBundlePath(spec)
+	assert.Error(err)
+
+	spec.Root.Path = rootFsPath
+	cwd, err := pb.ChangeToBundlePath(spec)
+	assert.NoError(err)
+	assert.Equal(cwd, originalCwd)
+
+	cwd, err = os.Getwd()
+	assert.NoError(err)
+	assert.Equal(bundlePath, cwd)
+}
+
+func TestWriteSpecToFile(t *testing.T) {
+	assert := assert.New(t)
+
+	bundlePath, err := ioutil.TempDir("", "bundle")
+	assert.NoError(err)
+	defer os.RemoveAll(bundlePath)
+
+	originalCwd, err := os.Getwd()
+	assert.NoError(err)
+	defer os.Chdir(originalCwd)
+
+	err = os.Chdir(bundlePath)
+	assert.NoError(err)
+
+	spec := &specs.Spec{}
+	spec.Root = &specs.Root{
+		Path:     "/this/is/a/path/",
+		Readonly: false,
+	}
+	err = pb.WriteSpecToFile(spec)
+	assert.NoError(err)
+
+	file, err := os.Open(path.Join(bundlePath, pb.OCIConfigFile))
+	assert.NoError(err)
+	defer file.Close()
+
+	stat, err := file.Stat()
+	assert.NoError(err)
+
+	assert.True(stat.Size() > 0)
 }
--- a/protocols/grpc/utils.go
+++ b/protocols/grpc/utils.go
@@ -7,12 +7,23 @@
 package grpc
 
 import (
+	"encoding/json"
+	"errors"
+	"os"
+	"path"
+	"path/filepath"
 	"reflect"
 
 	"google.golang.org/grpc/codes"
 	grpcStatus "google.golang.org/grpc/status"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// OCI config file
+const (
+	OCIConfigFile     string      = "config.json"
+	ociConfigFileMode os.FileMode = 0444
 )
 
 func copyValue(to, from reflect.Value) error {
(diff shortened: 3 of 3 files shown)


grahamwhaley:
Hi @eguzman3 Thanks for the contribution :-)
Did you have a specific use case or test that found or showed this up? That might help give us some context.
I'm wondering if this fix is best done in the agent or the runtime - but, others have a better view of that than I - so /cc @jodh-intel @bergwolf @WeiZhang555 for input.

grpc.go
```diff
@@ -524,6 +525,76 @@ func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container) {
 	}
 }
 
+func changeToBundlePath(spec *specs.Spec) (string, error) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return cwd, err
+	}
+
+	rootfsPath := spec.Root.Path
```
jodh-intel: I know that `spec` and `spec.Root` and `spec.Root.Path` MUST be specified, but I'd still opt for paranoia and check they are not `nil`, not `nil` and `!= ""` since if this resolves to `""`, the function will do nothing and the `config.json` will be written to `cwd` I think?

grpc.go
```diff
@@ -524,6 +525,76 @@ func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container) {
 	}
 }
 
+func changeToBundlePath(spec *specs.Spec) (string, error) {
```
jodh-intel: This (and `writeSpecToFile()`) are OCI-specific functions so please can you move them to `protocols/grpc/utils.go`.

grpc.go
```diff
@@ -524,6 +525,76 @@ func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container) {
...
+	return cwd, nil
+}
+
+func writeSpecToFile(spec *specs.Spec) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	configPath := path.Join(cwd, "config.json")
```
jodh-intel: Once this function is in `protocols/grpc/utils.go`, you could move `ociConfigFile` from `protocols/grpc/utils_test.go` into `protocols/grpc/utils.go` to avoid hard-coding it.

grpc.go
```diff
@@ -524,6 +525,76 @@ func (a *agentGRPC) rollbackFailingContainerCreation(ctr *container) {
...
+	f, err := os.OpenFile(configPath, os.O_RDWR|os.O_CREATE, 0666)
```
jodh-intel: Could you move the perms to a global `const` and make it an explicit `os.FileMode` value? Also, can't it be `0444`?

grpc.go
```diff
@@ -590,38 +667,28 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
+	// directory. https://github.com/opencontainers/runtime-spec
+	err = writeSpecToFile(ociSpec)
 	if err != nil {
 		return emptyResp, err
 	}
-	ctr.config = *config
 
-	ctr.initProcess, err = buildProcess(req.OCI.Process, req.ExecId)
+	// Change back to root dir
+	err = os.Chdir(oldcwd)
```
jodh-intel: I think it would be safer if this was a `defer` immediately after the error check for the `changeToBundlePath()` call.

grpc.go
```diff
@@ -590,38 +667,28 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
-
-	if err := a.updateSharedPidNs(ctr); err != nil {
+	// Update libcontainer configuration for specific cases not handled
+	// by the specconv converter.
+	if err = a.updateContainerConfig(ociSpec, config, ctr); err != nil {
 		return emptyResp, err
 	}
 
-	return emptyResp, a.postExecProcess(ctr, ctr.initProcess)
+	// reduce cyclomatic complexity of this function
```
jodh-intel: Please can you remove this comment - feel free to mention you needed to refactor to reduce cyclomatic complexity in the commit message though.

eguzman3:
@grahamwhaley @bergwolf 
This issue was discovered when trying to use an existing OCI hook within the guest. This hook reads the bundle path from the `HookState` passed over stdin, then attempts to open `/path/to/bundle/config.json` which doesn't exist because the bundle path is wrong and because `config.json` is never written. The reason the cwd needs to be changed is that `CreateLibcontainerConfig` sets the bundle path to cwd (https://github.com/kata-containers/agent/blob/5765593309eab087598c093ff226500af158df9c/vendor/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go#L157) which in this case is `/` so this is what the OCI hooks end up getting instead of `/run/kata-containers/shared/containers/<hash>/`

bergwolf:
@eguzman3 I think there are a few things to clarify:
1. The bundle is never passed to the guest. We only send the rootfs, which is a sub-directory in the bundle.
2. The `config.json` is already in the bundle when that is passed to kata from either containerd or crio. Otherwise `kata-runtime` cannot read the OCI spec in the first place.
3. If you need the hooks, they are in the OCI spec passed in as create container arguments.

Another thing to consider is do we really want to run the hooks inside the container? The hooks are mostly used by docker to setup network configuration for a container and in that sense they are only useful when run on the host side. What use case do you have that would require running hooks in side the guest?

sboeuf:
There is no reason to reproduce the whole OCI config.json thing on the guest OS and therefore the agent. The OCI spec only needs to be passed as a structure, and it's already done, which allows the agent to provide libcontainer with the appropriate data.
The hook issue that you're mentioning is separate from the `config.json` need, and is actually not needed since we don't want to run the hook from inside the guest as this makes no sense. The caller (Docker, CRI-O, or something else) will expect the hook to be executed on the host, and that's why this is handled directly by the kata-runtime CLI, and that's also the reason why the Hooks structure is set to `nil` by `kata_agent.go` before to be passed to the agent.

sboeuf:
-1

amshinde:
@sboeuf  I think @eguzman3 has a use case that makes it necessary for running hooks inside the guest. 
@eguzman3 Can you please describe your use case so that others can get a broader view of what you are trying to do.

eguzman3:
@sboeuf @amshinde 
Yes the use case requires that the hooks be run inside the guest. Specifically we are trying to leverage this existing OCI hook: https://github.com/NVIDIA/nvidia-container-runtime

sboeuf:
@eguzman3 what is the hook doing exactly ? I'm trying to understand the real need for this.
Also, one way to workaround this more properly would be to use a sidecar, instead of relying on the pre-start hooks which are complex to manage in case of Kata. And if we can avoid to make too many assumptions, that would be great.
Are you using k8s to orchestrate your containers ?

eguzman3:
This hook configures GPU access for a container using `libnvidia-container`. This is intended to be used in conjunction with GPU passthrough which is why this hook needs to run inside the guest.

flx42:
As mentioned in #347, the guest hooks will not be part of the Kata API, we want them to be part of the guest rootfs and auto-discovered by the agent.

sboeuf:
@eguzman3 @flx42, ok now I see.
That works, but I have one comment regarding the configuration of such a thing. As this is very specific to the fact of running some custom binaries on the guest directly, could you add some configuration to our CLI configuration, so that we don't end up trying to run some custom hooks on the agent everytime.
Most of the time, those hooks won't exist, and there's no need for the agent to look for them. Also, having the path being specified from the configuration, and not hardcoded inside the agent would be a more proper approach.

amshinde:
Yes it would make sense to configure the path for guest rootfs hooks in the runtime `configuration.toml` and pass this to the agent as part of `CreateSandbox` request.

eguzman3:
@sboeuf Can I get some clarification on the need for this configuration option? Is this a performance hit concern? Also would this runtime configuration option be disabled when the option is commented out?

sboeuf:
@eguzman3 yes it's about keeping the default path as simple and efficient as possible. And by default the agent would not try to load any custom hook.

jodh-intel:
I think I can see why this might be required...

As specified in https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks,

> The state of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.

But the [state](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state) of the container must include:

> **`bundle`** (string, REQUIRED) is the absolute path to the container's bundle directory. This is provided so that consumers can find the container's configuration and root filesystem on the host.

So the arbitrary hook binaries that need to be run inside the container must be passed the container state and must therefore have access to the bundle directory. And a bundle is "rootfs + `config.json`":

- https://github.com/opencontainers/runtime-spec/blob/master/bundle.md

sboeuf:
@jodh-intel thanks for looking into this. It's still not clear if this is needed from a libcontainer perspective though. I mean that libcontainer gets provided with the container status through a structure, and not through a CLI, so I don't know how far we need to stick to the OCI definition here.

sboeuf:
Doing some research in libcontainer codebase, here's my understanding:
[here](https://github.com/opencontainers/runc/blob/master/libcontainer/process_linux.go#L343-L357) is the code that gets called to handle OCI hook (case of pre-start, which is actually the one you're interested into).
If the list of `Labels` is empty, the function returning the bundle and the list of Annotations [here](https://github.com/opencontainers/runc/blob/master/libcontainer/process_linux.go#L344) will simply return an empty bundle string, and an empty list of Annotations.
So what I'm trying to say is that it's not a big deal from a libcontainer perspective, we won't get any error with that.
And then, it's simply up to your hook, not to use those fields. Is there any reason why the custom hook would need them ?

flx42:
The `config.json` file has two fields we need:
* The path to the `rootfs` (despite what the spec says, it's not always in the same directory tree as the config file).
* Environment variables/labels from the container image. We use that to tweak the behavior of our prestart hook: https://github.com/nvidia/nvidia-container-runtime#environment-variables-oci-spec

sboeuf:
@flx42 
>The path to the rootfs (despite what the spec says, it's not always in the same directory tree as the config file).

Why do you need the rootfs path here?
Also, just to mention, giving the rootfs path is orthogonal from creating `config.json` since the path can be provided simply through the lbcontainer `Config` structure.

>Environment variables/labels from the container image or set by the user. We use that to tweak the behavior of our prestart hook: https://github.com/nvidia/nvidia-container-runtime#environment-variables-oci-spec

Environment variables should not be an issue since they will be passed from the initial `config.json` on the host side, and they'll be carried up to the agent to be applied by libcontainer.

sboeuf:
@flx42 this is an interesting problem that we have here :)

Ok I think here could be a solution that would be a good tradeoff on how handle those customized OCI hooks:
- First, you detect from the agent that some custom hooks are available.
- Once detected, the agent will build the appropriate `Hook` structure, including the `Env` that he knows from the OCI `Spec` structure. This way, the hook will be started with the appropriate environment variables.
- In order to provide the rootfs path, I suggest that you modify slightly the NVIDIA hook so that it would look for the rootfs path through the enviroment variables. By default, it would rely on the bundle path (I don't want to break your current behavior), but if the path is empty, it should check if it can find an envvar `CONTAINER_ROOTFS`. This new environment variable should be added by the agent when creating the hook from scratch, so that it will provide the right amount of knowledge to the hook.

WDYT?

flx42:
@sboeuf We are flexible, we can definitely have some kind of glue layer that translates the environment variables to something else and then call into our `nvidia-container-runtime-hook` or even `nvidia-container-cli`. What you describe is closer to our LXC hook implementation than to a standard OCI hook:
https://github.com/lxc/lxc/blob/master/hooks/nvidia
Here we clearly needed a custom script since LXC does not follow the OCI runtime spec.

With that said, I'm not sure I understand why you don't want to provide support for standard OCI runtime hooks, what is the concern with this approach? We were hoping to leverage our existing OCI hook since we were missing only a small piece to make it happen.

sboeuf:
@flx42 
>With that said, I'm not sure I understand why you don't want to provide support for standard OCI runtime hooks, what is the concern with this approach? We were hoping to leverage our existing OCI hook since we were missing only a small piece to make it happen.

The main concern that I have here is that we're trying to implement the agent as an implementation of the OCI specification, but that's not what the agent is.
Now that we've sorted out the need you have for your problem, it'd be nice to get more input from a few ppl on this topic.
/cc @bergwolf @amshinde @jodh-intel WDYT ?

amshinde:
Here is my take on this:
I am fine with writing the config.json inside the guest, that way OCI compliant hooks that need to be run inside the guest can be executed without any changes. 
If we do things differently (lets say via environment variables), then an existing hook that is OCI compliant will need to be changed for running it inside the guest.

That said, we should  write the config.json inside the guest only if hooks need to be run.
We can provide this information to the agent, using a runtime configuration such as `guest_hooks_path` which will be empty by default. This configuration can then be passed to the agent as part of the CreateSandbox API. Only if this path is non-empty, will the agent create the config.json file and execute the hooks at the provided path.

@bergwolf @WeiZhang555 @sboeuf What do you think?

bergwolf:
I agree with @sboeuf that kata-agent is not an OCI compatible implementation on its own.

@eguzman3 @flx42 Can you do it with a sidecar container? AFAIU, with OCI hooks, you need to add whatever is needed in the guest rootfs image. That means if a user wants to use `nvidia-container-runtime-hook`, they need to install a customized guest rootfs image rather than the one shipped by kata containers. OTOH, if you do it with a sidecar container, it runs on all existing kata guest rootfs images.

And you need to pay attention that whatever you put in the OCI hooks, they are executed on the host as well. Do you really need the **SAME** hooks to be run both on the host and in the guest?

flx42:
> Can you do it with a sidecar container?

Driver installation is possible with a container, but at the container runtime level, not really. And isn't that a Kubernetes construct? That won't work with Docker (or another OCI frontend).

> That means if a user wants to use nvidia-container-runtime-hook, they need to install a customized guest rootfs image rather than the one shipped by kata containers. OTOH, if you do it with a sidecar container, it runs on all existing kata guest rootfs images.

Yes, we are fine with that, since the rootfs will also bake our user-space driver libraries, and precompiled objects for the device driver.

> And you need to pay attention that whatever you put in the OCI hooks, they are executed on the host as well. Do you really need the SAME hooks to be run both on the host and in the guest?

We want to only execute OCI hooks on the "translated" OCI spec, by discovering them autoamtically. There is no way to inject hooks in the spec docker generates today.

bergwolf:
> Driver installation is possible with a container, but at the container runtime level, not really. And isn't that a Kubernetes construct? That won't work with Docker (or another OCI frontend).

I mean whatever initialization you do with current hooks, do it in a sidecar container, -- not running the same hooks there. And docker does support sidecars, it is just about constructing proper namespace configurations for different containers. Maybe it's too much complexity for your users though.

> Yes, we are fine with that, since the rootfs will also bake our user-space driver libraries, and precompiled objects for the device driver.

In that case, how about having a builtin container inside the rootfs image? Then we can make kata containers create default sidecars when configured so. And such a design would solve the socat sidecar image issue @gnawux is trying to solve for k8s port forwarding.

> We want to only execute OCI hooks on the "translated" OCI spec, by discovering them autoamtically. There is no way to inject hooks in the spec docker generates today.

Sorry, I kinda miss the whole picture here. If not docker, who will translate the OCI spec and how? Do you expect kata-runtime to do the translation and only put translated hooks in the spec when passing them to kata-agent? Please explain a bit more about the architecture you are expecting to work with.

flx42:
> I mean whatever initialization you do with current hooks, do it in a sidecar container

It's not possible, it would require changing our whole implementation. Our hook performs bind mounts from the runtime's namespace to the container. For instance it looks for `libcuda.so.390.55` in the runtime's mount namespace and bind-mount it under the container's rootfs mount point.
With a sidecar container, I guess you would have to share a mount point between the two containers, and point the application container to the libraries/binaries in this directory.
But that's not all, we also need to bind-mount devices in `/dev`, that's not possible with a side-car container.

> Maybe it's too much complexity for your users though.

I agree. And the implementation is possible but would be hacky.

> In that case, how about having a builtin container inside the rootfs image?

This would be overkill for our use case, and would not solve the problem as well as what we have today, IMO.

> Sorry, I kinda miss the whole picture here. If not docker, who will translate the OCI spec and how?

Docker (or something else) gives you the OCI spec, and my understanding is that once in the guest you amend this spec before passing it to libcontainer. @sboeuf mentioned that hooks from the original OCI Spec are removed: https://github.com/kata-containers/agent/pull/346#issuecomment-418438160

Here, we want to add hooks from the guest rootfs, a possible way of doing that is presented in https://github.com/kata-containers/agent/pull/347

bergwolf:
@flx42 
>But that's not all, we also need to bind-mount devices in /dev, that's not possible with a side-car container.

As long as devices share the same major/minor pair, they are shared among different containers in the same guest, which means you do not need to bindmount them. Just pass the same major/minor pair in the OCI spec, and they will be treated as the same device by kernel.

OTOH, I agree in your use case the arbitrary rootfs hooks are much simpler than adding implicit sidecars. Thanks for the explanation. Please go ahead with the modification and I'm fine with saving the spec somewhere in the rootfs.

grpc.go
```diff
@@ -579,6 +608,13 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	// Change cwd because libcontainer sets the bundle path to cwd
+	oldcwd, err := pb.ChangeToBundlePath(ociSpec)
```
bergwolf: These are only useful when there are hooks in the rootfs. So please make it optional by:
1. scan for rootfs hooks when agent gets up
2. change cwd and save spec in container bundle only when necessary

eguzman3:
Consolidating this PR with https://github.com/kata-containers/agent/pull/347

PR #347, opened 2018-08-30
agent: add support for custom OCI hooks inside guest

eguzman3:
In order to allow OCI hooks to be executed during the lifecycle of the container inside the guest the path "/usr/share/oci/hooks/" has been designated to allow for 'drop-in' hooks.


caoruidong:
Who will put the hooks into /usr/share/oci/hooks/ ?

eguzman3:
@caoruidong @amshinde I am assuming the hooks will be added to that directory at the time of rootfs image creation. The use case for this is trying to leverage this existing OCI hook: https://github.com/NVIDIA/nvidia-container-runtime

flx42:
Just chiming in to point out that this drop-in hook mechanism was inspired by what Project Atomic's fork of Docker is doing:
https://github.com/projectatomic/docker/tree/docker-1.13.1-rhel#add-dockerhooks-exec-custom-hooks-for-prestartpoststop-containerspatch
And also `libpod` (for CRI-O):
https://github.com/containers/libpod/blob/v0.8.5/pkg/hooks/docs/oci-hooks.5.md

As @eguzman3, we want to ship one or multiple pre-start hooks in our guest rootfs.

bergwolf:
@eguzman3 Could you combine your two PRs into one? I see that they already share some common change. They are in fact adding the same functionality to the agent. There is no need to separate them. Within one PR, you can still have multiple commits.

grpc.go
```diff
@@ -579,49 +674,45 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
 
-	// Convert the OCI specification into a libcontainer configuration.
-	config, err := specconv.CreateLibcontainerConfig(&specconv.CreateOpts{
-		CgroupName:   req.ContainerId,
-		NoNewKeyring: true,
-		Spec:         ociSpec,
-		NoPivotRoot:  a.sandbox.noPivotRoot,
-	})
+	// Add any custom OCI hooks to the spec
+	err = addHooks(ociSpec)
```
bergwolf: The hooks are static afaics. So it is more efficient to just scan once during agent startup, and do the OCI spec/hook dance only when there are real hooks to add.

PR #350, opened 2018-08-31
CI: Add "make proto"

### LINKED ISSUE #340: Add "make proto" to CI build
We should have a CI test to ensure it is always possible to regenerate `protocols/grpc/agent.proto` from `protocols/grpc/agent.pb.go` (and the same for the other `*.proto` files).

This should be as simple as adding `make proto` to `.ci/run.sh`.

/cc @WeiZhang555.

caoruidong:
Allow make proto to work under the CI. So later we are able to check
pb.go files in CI.

fixes #340

Signed-off-by: Ruidong Cao <caoruidong@huawei.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ build-image:
 	docker build ${BUILDARGS} -t ${AGENT_IMAGE}:${AGENT_TAG} .
 
 proto: build-image
-	docker run -ti -v ${PWD}:/go/src/github.com/kata-containers/agent ${AGENT_IMAGE}:${AGENT_TAG} ./hack/update-generated-agent-proto.sh
+	docker run -i -v ${PWD}:/go/src/github.com/kata-containers/agent ${AGENT_IMAGE}:${AGENT_TAG} ./hack/update-generated-agent-proto.sh
 
 .PHONY: clean test
 clean:
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -11,4 +11,7 @@ source "${cidir}/lib.sh"
 
 pushd "${tests_repo_dir}"
 .ci/run.sh
+testcidir=$(dirname "$0")
+"${testcidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker configure -r runc -f
 popd
+make proto


caoruidong:
@jodh-intel Do we need to check the diff of pb.go files?

WeiZhang555:
LGTM 

But CI isn't happy.

jodh-intel:
Looking at two of the logs, it appears there is a bad package in the centos repo that is used by the agent's `Dockerfile`:

```
Rpmdb checksum is invalid: dCDPT(pkg checksums): fipscheck-lib.x86_64 0:1.4.1-6.el7 - u

The command '/bin/sh -c yum makecache && yum install -y     git automake libtool glibc-headers gcc-c++ make' returned a non-zero code: 1
make: *** [Makefile:54: build-image] Error 1
```

.. but that looks a bit odd as that package hasn't changed for over a year:

- http://www.mirrorservice.org/sites/mirror.centos.org/7/os/x86_64/Packages

caoruidong:
Yes. Seems that "make proto" is too heavy and not stable for CI. Maybe just call "hack/update-generated-agent-proto.sh" is better.

jodh-intel:
Hi @caoruidong - could you force a re-push to re-trigger the CI's? I've just run this locally and it works for me.

jodh-intel:
@caoruidong - oh wait, I can see that you have re-pushed :) Problem with my browser cache ;)

caoruidong:
After google, still can't find a way to solve the error.

jodh-intel:
I've tried again on a different system and it works there too.

Any thoughts @chavafg? Is the centos docker image cached (and corrupt?) maybe?

chavafg:
At that point, docker runs using kata-containers by default, and it seems that we are corrupting something when trying to install the packages in centos. I could reproduce it on my system.

When running using runc, the problem disappears.

jodh-intel:
Oooh - could it be a 9p issue I wonder?

/cc @grahamwhaley.

grahamwhaley:
Could be - we have seen 9p unlink/lock type issues before that have made package install/update fail.

jodh-intel:
It might be similar to https://github.com/clearcontainers/runtime/issues/868.

WeiZhang555:
> we just check PR by hand

caoruidong:
Add dnm to see if this change can work as we expect

jodh-intel:
Thanks @caoruidong!

lgtm

jodh-intel:
@caoruidong - do you want to remove the dnm label now?

Makefile
```diff
@@ -54,7 +54,7 @@ build-image:
 	docker build ${BUILDARGS} -t ${AGENT_IMAGE}:${AGENT_TAG} .
 
 proto: build-image
-	docker run -ti -v ${PWD}:/go/src/github.com/kata-containers/agent ${AGENT_IMAGE}:${AGENT_TAG} ./hack/update-generated-agent-proto.sh
+	docker run -i -v ${PWD}:/go/src/github.com/kata-containers/agent ${AGENT_IMAGE}:${AGENT_TAG} ./hack/update-generated-agent-proto.sh
```
bergwolf: Sorry I fail to see how this **checks** if pb.go regenerated right. Could you please enlighten me?

jodh-intel: I think the `-t` has probably been removed here since Jenkins doesn't provide a tty to the process, so this change is required to allow `make proto` to work under the CI.

But yes, it's not checking so much as "ensuring the regeneration target doesn't fail".

See https://github.com/kata-containers/agent/pull/350#issuecomment-417590652.

caoruidong: Yes, different version of protoc will generate quite different pb.go files. I think we all use v3.5 luckily.

sboeuf:
@caoruidong could you please update the commit message and the PR message, so that it fits the description provided by @jodh-intel [here](https://github.com/kata-containers/agent/pull/350#discussion_r215618879).
After that, I'll merge the change, thx!

caoruidong:
Update!

sboeuf:
Thanks @caoruidong !

PR #355, opened 2018-09-03
agent: get content of a file in sandbox

### LINKED ISSUE #354: agent: get content of a file in sandbox
# Description of problem

In https://github.com/kata-containers/runtime/pull/624, we have such problems:
1. Hot add memory should be aligned to memory section size, but the size of a memory section is architecture dependent. For example, power uses 16MB, ia64 uses 1GB, x86_64 uses 128MB, ppc64le uses 256MB.
2. Each memory block is described under /sys/devices/system/memory as /sys/devices/system/memory/memoryXXX (XXX is the memory block id.), which is the unit of memory online/offline. In sparse memory model, memory block size is a multiple of memory section size, [reference](http://www.infradead.org/~mchehab/kernel_docs/unsorted/memory-hotplug.html#unit-of-memory-online-offline-operation). We can online a memory block only if we hotplug a memory-block-sized dimm, see `pages_correctly_reserved` in [reference](https://elixir.bootlin.com/linux/v4.14.49/source/drivers/base/memory.c#L230).

So we should get the size of guestos memory block by reading /sys/devices/system/memoryblock_size_bytes in sandbox, and make hot add memory to be aligned to memory block size of sandbox. 

I think it's needed to add a ReadSandboxFile rpc to read file above in a common way.

clarecch:
Add a ReadSandboxFile rpc to read a file in sandbox in a common way,
which is also needed by memory hotplug.

Fixes #354

Signed-off-by: Clare Chen <clare.chenhui@huawei.com>


clarecch:
The CI is green. PTAL!

grpc.go
```diff
@@ -1304,3 +1304,18 @@ func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemReques
...
+		return nil, grpcStatus.Error(codes.FailedPrecondition, "File path should not be empty")
+	}
+
+	var fileData pb.FileData
+	data, err := ioutil.ReadFile(req.FilePath)
+	if err != nil {
+		return nil, err
+	}
+	fileData.Data = data
+
```
jshachm: no need for defining a var for `fileData`

use codes like below to make it clear

```
return &pb.FileData {
      Data : data,
}
```

linzichang: Done :)

clarecch:
Updated， PTAL

grpc.go
```diff
@@ -1304,3 +1304,18 @@ func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemReques
 func (a *agentGRPC) ReseedRandomDev(ctx context.Context, req *pb.ReseedRandomDevRequest) (*gpb.Empty, error) {
 	return emptyResp, reseedRNG(req.Data)
 }
+
+func (a *agentGRPC) ReadSandboxFile(ctx context.Context, req *pb.ReadSandboxFileRequest) (*pb.FileData, error) {
```
jodh-intel: There might be security implications of adding this call. At the very least, shouldn't we ensure the path is absolute?

Issue #359, opened 2018-09-05
Add serial channel close timeout


bergwolf:
As @sboeuf mentioned [here](https://github.com/kata-containers/agent/pull/338#discussion_r215174991), we should timeout waiting for yamux session close.

Issue #307, opened 2018-07-26
TestNewAgentClientWithYamux unit test timing out


chavafg:
I have seen this lately:

```
11:31:24 INFO: Running 'go test' as root user on package 'github.com/kata-containers/agent/protocols/client' with flags '-v -race -timeout 10s'
11:31:25 === RUN   TestNewAgentClient
11:31:26 --- PASS: TestNewAgentClient (1.01s)
11:31:26 === RUN   TestNewAgentClientWithYamux
11:31:35 coverage: 79.2% of statements
11:31:35 panic: test timed out after 10s
11:31:35 
11:31:35 goroutine 10 [running]:
11:31:35 testing.(*M).startAlarm.func1()
11:31:35 	/usr/local/go/src/testing/testing.go:1240 +0x146
11:31:35 created by time.goFunc
11:31:35 	/usr/local/go/src/time/sleep.go:172 +0x52
```

From: 
http://jenkins.katacontainers.io/job/kata-containers-agent-centos-7-4-master/17/console
http://jenkins.katacontainers.io/job/kata-containers-agent-centos-7-4-master/19/console

bergwolf:
Hit another at 
http://jenkins.katacontainers.io/job/kata-containers-agent-fedora-27-PR/205/console
```
=== RUN   TestNewAgentClientWithYamux
coverage: 81.9% of statements
panic: test timed out after 10s

goroutine 79 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:1240 +0x146
created by time.goFunc
	/usr/local/go/src/time/sleep.go:172 +0x52

goroutine 1 [chan receive]:
testing.(*T).Run(0xc4201c0000, 0xca1582, 0x1b, 0xcbd150, 0xc42006fc01)
	/usr/local/go/src/testing/testing.go:825 +0x597
testing.runTests.func1(0xc4201c0000)
	/usr/local/go/src/testing/testing.go:1063 +0xa5
testing.tRunner(0xc4201c0000, 0xc42006fd48)
	/usr/local/go/src/testing/testing.go:777 +0x16e
testing.runTests(0xc4201924a0, 0x1054260, 0x2, 0x2, 0xc4201be000)
	/usr/local/go/src/testing/testing.go:1061 +0x4e2
testing.(*M).Run(0xc4201be000, 0x0)
	/usr/local/go/src/testing/testing.go:978 +0x2ce
main.main()
	_testmain.go:90 +0x325

[...]

PR #360, opened 2018-09-05
channel: add serial yamux channel close timeout

bergwolf:
So that if serial yamux session fails to close somehow,
we can catch it.

Fixes: #359

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -25,6 +25,7 @@ import (
 var (
 	channelExistMaxTries   = 200
 	channelExistWaitTime   = 50 * time.Millisecond
+	channelCloseTimeout    = 5 * time.Second
 	isAFVSockSupportedFunc = isAFVSockSupported
 )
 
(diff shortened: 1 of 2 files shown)


channel.go
```diff
@@ -209,7 +210,13 @@ func (c *serialChannel) listen() (net.Listener, error) {
 func (c *serialChannel) teardown() error {
 	// wait for the session to be fully shutdown first
 	if c.waitCh != nil {
-		<-c.waitCh
+		t := time.NewTimer(channelCloseTimeout)
+		select {
+		case <-c.waitCh:
+			t.Stop()
```
devimc: according with the doc

```go
// To prevent a timer created with NewTimer from firing after a call to Stop,
// check the return value and drain the channel.
// For example, assuming the program has not received from t.C already:
//
// 	if !t.Stop() {
// 		<-t.C
// 	}
//
```

bergwolf: No, that's not necessary. That comment is to prevent accidental read from the timer channel but in our case the function returns and nothing is reading that channel.

devimc: ok, make sense

jodh-intel:
lgtm

bergwolf:
`jenkins-ci-fedora-27` failed on #307 and I've sent #361 to fix it. Let's get this one merged since all the other CIs have passed.

bergwolf:
And travis is somehow stuck? There is not even a hyperlink there for it.
`continuous-integration/travis-ci Expected — Waiting for status to be reported`

Issue #363, opened 2018-09-12
workaround memory hotplug online


bergwolf:
Because we do not wait for memory device to show up when online it, it is quite possible that we cannot online it properly.
```
[macbeth@runtime]$docker run --rm -it --runtime kata -m 536870912 busybox
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
/ # free -m
             total       used       free     shared    buffers     cached
Mem:           108         53         54         30          0         31
-/+ buffers/cache:         22         85
Swap:            0          0          0
/ # cd /sys/devices/system/memory/
/sys/devices/system/memory # ls
auto_online_blocks  block_size_bytes    memory0             memory32            memory33            memory34            uevent
/sys/devices/system/memory # cat */online
1
0
0
0
```

bergwolf:
btw, this only happens with vm factory right now, because in non-factory case, memory is not hotplugged yet. We'll hit this when adding memory hotplug with `kata-runtime update`. cc @clarecch

PR #364, opened 2018-09-12
agent: auto-online hotplug memory

bergwolf:
Kata agent does not wait for memory hotplug, and kata-runtime
does not wait on cpu/memory online. As a temporary fix, let's
make sure memory are onlined by the uevent listener at least.

Fixes: #363

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -486,6 +486,11 @@ func (s *sandbox) listenToUdevEvents() {
 			continue
 		}
 
+		// We only care about add event
+		if uEv.Action != "add" {
+			continue
+		}
+
 		fieldLogger = fieldLogger.WithFields(logrus.Fields{
 			"uevent-action":    uEv.Action,
 			"uevent-devpath":   uEv.DevPath,
@@ -495,7 +500,7 @@ func (s *sandbox) listenToUdevEvents() {
 		})
 
 		// Check if device hotplug event results in a device node being created.
-		if uEv.DevName != "" && uEv.Action == "add" && strings.HasPrefix(uEv.DevPath, rootBusPath) {
+		if uEv.DevName != "" && strings.HasPrefix(uEv.DevPath, rootBusPath) {
 			// Lock is needed to safey read and modify the pciDeviceMap and deviceWatchers.
 			// This makes sure that watchers do not access the map while it is being updated.
 			s.Lock()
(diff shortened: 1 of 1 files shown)


gnawux:
nice
LGTM

devimc:
cc @clarecch  @linzichang

agent.go
```diff
@@ -486,6 +486,11 @@ func (s *sandbox) listenToUdevEvents() {
 			continue
 		}
 
+		// We only care about add event
+		if uEv.Action != "add" {
+			continue
+		}
```
sboeuf: Please put this in a separate commit, as this is not related to the uevent for memory.

agent.go
```diff
@@ -495,7 +500,7 @@ func (s *sandbox) listenToUdevEvents() {
 		})
 
 		// Check if device hotplug event results in a device node being created.
-		if uEv.DevName != "" && uEv.Action == "add" && strings.HasPrefix(uEv.DevPath, rootBusPath) {
+		if uEv.DevName != "" && strings.HasPrefix(uEv.DevPath, rootBusPath) {
```
sboeuf: Same thing here, this should go to a separate commit.

linzichang:
I have a question. Will this face the same question I found at https://github.com/kata-containers/runtime/pull/624#issuecomment-419351740 ?

agent.go
bergwolf: Why ask for splitting a commit into two? The checks **MUST** be separated otherwise I have to check it again when testing for memory online path.

agent.go
bergwolf: Same here. It **IS** a single commit.

bergwolf:
@linzichang Are you referring to the systemd udev issue? For one thing, the commit does not make things worse and it actually fixes the `-m` memory hotplug issue I'm seeing with vm factory. For another thing, I think systemd udev should be disabled, otherwise the uevent listener with device map approach is broken, ditto the cpu online waiting logic.

linzichang:
@bergwolf The main issue is `onlineCPUResources` conflicting with other online method. See https://github.com/kata-containers/agent/blob/master/grpc.go#L145 . Your changes just care about memory, so it wouldn't be a problem :) But the `udev` should be disabled.

agent.go
sboeuf: Ah ok I understand.

sboeuf:
Travis is stuck as for a bunch of other PRs. Ignoring for now and merging this

PR #365, opened 2018-09-12
agent: add support for guest-hooks

### LINKED ISSUE #348: Add support for OCI hooks inside the guest
In order to leverage existing OCI hooks, there should be a way of designating hooks to execute within the life cycle of containers inside the guest.

eguzman3:
Adds support for running OCI hooks within the guest.
A 'drop-in' path (guest_hook_path) is specified in
the cli configuration file and if set, the agent will
look for OCI hooks in this directory and inject them
into the container life cycle.

Old Discussion: https://github.com/kata-containers/agent/pull/346, https://github.com/kata-containers/agent/pull/347
Runtime Change: https://github.com/kata-containers/runtime/pull/720

Fixes #348


grpc.go
```diff
@@ -578,6 +607,24 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	if a.sandbox.guestHookPath != "" {
+		// Add any custom OCI hooks to the spec
+		pb.AddGuestHooks(ociSpec, a.sandbox.guestHookPath)
```
bergwolf: Now I see you have pushed the hookpath to runtime configuration, still please only scan the hook path **one time only** and save the result in the sandbox data structure. You can do it in the `CreateSandbox` handler instead of `CreateContainer` here.

protocols/grpc/utils.go
```diff
@@ -15,6 +21,12 @@ import (
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
+// OCI config file
+const (
+	OCIConfigFile     string      = "config.json"
```
jodh-intel: This looks a bit odd. For consistency, please use the same capitalisation - I'd suggest changing this to use a `oci` prefix to keep it private.

protocols/grpc/agent.proto
```diff
@@ -266,6 +266,7 @@ message CreateSandboxRequest {
 	// one sandbox per agent and implicitly require that CreateSandbox is
 	// called before other sandbox/network calls.
 	string sandbox_id = 5;
+	string guest_hook_path = 6;
```
jodh-intel: Please can you add a comment here explaining what this field is. Also, please specify if the path is expected to be an absolute one.

Related: #150.

protocols/grpc/utils.go
eguzman3: This was done access it from the tests, but moving the tests over to `utils_tests.go` is probably better right?

amshinde: Yes, since you are testing a function under utils.go, the test needs to go under utils_test.go.

grpc.go
amshinde: Agree. Please move the logic to scan the hooks to CreateSandbox.

eguzman3: Ok now it should only do the directory lookup once.

grpc.go
```diff
@@ -1180,6 +1202,12 @@ func (a *agentGRPC) CreateSandbox(ctx context.Context, req *pb.CreateSandboxRequ
 	a.sandbox.running = true
 	a.sandbox.sandboxPidNs = req.SandboxPidns
 	a.sandbox.storages = make(map[string]*sandboxStorage)
+	a.sandbox.guestHooks = &specs.Hooks{}
+	a.sandbox.guestHooksPresent = false
+
+	if req.GuestHookPath != "" {
+		a.sandbox.scanGuestHooks(req.GuestHookPath)
```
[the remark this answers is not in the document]
eguzman3: >Another question, since you are relying on the rootfs layout, is there a matching osbuilder PR that allows you to build a rootfs with customized hooks? Or do you simply add the hooks by hand between the rootfs creation and image creation steps?
Currently I am using scripts to add the hooks and other files needed for my use case between rootfs and image creation.

protocols/grpc/utils.go
```diff
@@ -287,3 +299,86 @@ func ResourcesGRPCtoOCI(grpcResources *LinuxResources) (*specs.LinuxResources, e
 
 	return s, err
 }
+
+// ChangeToBundlePath changes the cwd to the bundle path defined in the OCI spec
+func ChangeToBundlePath(spec *specs.Spec) (string, error) {
```
WeiZhang555: These utils functions might be better to move out from grpc package.

You know, this package can be vendored later to kata-runtime, because kata-runtime will be client and communicate with kata-agent via GRPC protocol, they share same GRPC protocol definitions.

So if these functions do no help to runtime, which means it can't help kata-runtime do GRPC client operations, it might be better to move out from this package.

protocols/grpc/utils.go
```diff
@@ -287,3 +299,86 @@ func ResourcesGRPCtoOCI(grpcResources *LinuxResources) (*specs.LinuxResources, e
...
+	}
+
+	configPath := path.Join(cwd, ociConfigFile)
+	f, err := os.OpenFile(configPath, os.O_RDWR|os.O_CREATE, ociConfigFileMode)
+	if err != nil {
+		return err
+	}
+
+	defer f.Close()
+	err = json.NewEncoder(f).Encode(&spec)
```
WeiZhang555: Question: spec is already a pointer, why need `&spec` here?

protocols/grpc/utils.go
```diff
@@ -287,3 +299,86 @@ func ResourcesGRPCtoOCI(grpcResources *LinuxResources) (*specs.LinuxResources, e
...
+	}
+
+	// check that this file is executable
+	perm := mode & os.ModePerm
+	return (perm & 0111) != 0
+}
+
+// FindHooks searches guestHookPath for any OCI hooks for a given hookType
+func FindHooks(guestHookPath, hookType string) (hooksFound []specs.Hook) {
+	hooksPath := path.Join(guestHookPath, hookType)
```
WeiZhang555: So `prestart` hook will be in `guestHookPath/prestart` and `poststart` hook will be in `guestHookPath/poststart`, 
and `guestHookPath` is specified by kata-runtime cli configuration, right?

I think we need some documents to teach user, we can't expect user can find this hidden advanced feature.

WeiZhang555:
This implementation looks good, but there's one missing part is that we can't specify arguments for each hook, I can't be sure that if there will be any requirements for this regarding to some unknown scenarios.

protocols/grpc/utils.go
jodh-intel: +1 ;)

protocols/grpc/utils.go
```diff
@@ -287,3 +299,86 @@ func ResourcesGRPCtoOCI(grpcResources *LinuxResources) (*specs.LinuxResources, e
...
+	if err != nil {
+		return err
+	}
+
+	defer f.Close()
+	err = json.NewEncoder(f).Encode(&spec)
+	return err
+}
+
+func isValidHook(file os.FileInfo) bool {
```
jodh-intel: This is a nice clean function. However, I think it would be worth considering "polluting it" a bit with some log calls to help debug hook setup issues (basically a log call for every failure scenario).

grpc.go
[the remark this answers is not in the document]
jodh-intel: I agree with @bergwolf - if a user switches images, I'd say they should expect to have to update the config if one has hooks and the other doesn't.

As an absolute minimum, can you add a log call showing if the user requested hooks *or not* (if `GuestHookPath` is set or not we should log something).

I do wonder if we want to add either a second bool field in the proto file (`bool require_hooks = n`), or maybe even a kernel cmdline option for the agent to consume (the former is cleaner though):

- https://github.com/kata-containers/agent/blob/master/config.go#L67

That way, the agent can clearly establish whether the user wants hooks or not allowing the value (and validity) of `GuestHookPath` to be more clearly determined.

protocols/grpc/utils.go
jodh-intel: Good catch @WeiZhang555 :)

grpc.go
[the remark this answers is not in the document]
bergwolf: @amshinde In that case, I think the `GuestHookPath` config should also be passed by spec (e.g., through labels or annotations).

@eguzman3 I would prefer a fail-early strategy to let users know that they have pointed to a wrong rootfs image. For your mentioned use case, right now users have to remove `GuestHookPath ` if they switch to a rootfs image w/o hooks. In future we might support to specify all kata configs through labels or annotations, and then there is no dependency on a fixed cli configuration and each pod can be customized.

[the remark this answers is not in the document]
bergwolf: ref kata-containers/runtime#753

jodh-intel:
Hi @eguzman3 - do you think you'll get a chance to look over the final review comments some time this week? As mentioned, this is very close now I think :smile:

flx42:
@jodh-intel I will take over the patch since Edward finished his internship. Sorry for the delay!

jodh-intel:
Hi @flx42 - no problem! Do let us know if we can help in any way.

grahamwhaley:
This one is superseded by #393 , so I'll mark it as DNM so we don't try to review or accidentally merge it ;-)

amshinde:
Closing this PR, as it has been superseded by #393.

grahamwhaley:
Nudging the CIs that failed. One is due to the nested glitch. The other had what look like maybe real errors (but anything doing `apt` could be an upstream timeout type error). I'll respin to see if that is reproducible or transient:

```
Summarizing 2 Failures:

[Fail] package manager update test check apt-get update [It] should not fail 
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:41

[Fail] CPUs and CPU set updating cpus and cpuset of a running container [It] should have the right number of vCPUs 
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:432

Ran 202 of 214 Specs in 952.725 seconds
FAIL! -- 200 Passed | 2 Failed | 0 Pending | 12 Skipped --- FAIL: TestIntegration (979.86s)
FAIL
```

Issue #370, opened 2018-09-14
Exec a process and exit will hang there waiting for terminal close


lifupan:
# Description of problem

When exec a shell process and then run a background process in it, and
then exit this shell, it will hang there waiting the terminal close. 

How to reproduce: 
1:  id=`sudo docker run -d busybox top`
2:  sudo docker exec -ti $id sh
3: run "sleep 10000 &" in the container
4: run "exit"

The "exit" will not exit successfully and it will wait there.

The root cause is that, kata assumes once a process exits, its io/terminal will
also be closed, once the kata client received the io closed, it will then wait on
the process to clean it up. But the problem is that if a child process is run as 
background and it will also inherit its parent's terminal, thus even the parent 
process exits, it's terminal will not closed, thus the kata client will wait on the
terminal for ever.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

$ sudo kata-collect-data.sh

# Meta details

Running `kata-collect-data.sh` version `1.2.0 (commit a39a3f15a3c996648248de28be8719f17604fded)` at `2018-09-14.15:38:58.657470931+0800`.

---

Runtime is `/usr/local/bin/kata-runtime`.

# `kata-env`

Output of "`/usr/local/bin/kata-runtime kata-env`":
```
[Meta]
  Version = "1.0.15"

[Runtime]
  Debug = false
  Path = "/usr/local/bin/kata-runtime"
  [Runtime.Version]
    Semver = "1.2.0"
    Commit = "26e3443d7a6f698ee8557bfb66c454eb7c855d11-dirty"
    OCI = "1.0.1"
  [Runtime.Config]
    Path = "/usr/share/defaults/kata-containers/configuration.toml"

[Hypervisor]
  MachineType = "pc"
  Version = "QEMU emulator version 2.11.0\nCopyright (c) 2003-2017 Fabrice Bellard and the QEMU Project developers"
  Path = "/usr/bin/qemu-lite-system-x86_64"
  BlockDeviceDriver = "virtio-scsi"
  Msize9p = 8192
  Debug = false
  UseVSock = false

[Image]
  Path = ""

[Kernel]
  Path = "/usr/share/kata-containers/kata-vmlinuz-4.14.49.container"
  Parameters = "agent.log=debug"

[Initrd]
  Path = "/usr/share/kata-containers/kata-containers-initrd-2018-09-14-15:11:46.341680236+0800-a126188"

[Proxy]
[...]

PR #371, opened 2018-09-14
agent: fix the issue of missing close process terminal

lifupan:
Kata assumes once a process exits, it's IO will also
close automatically, thus the client will wait on the
process's IO closed and then do some cleanup for this
process. But the issue is that if the process forked
some background children processes running as daemon
process, thus those children process will intherit its
parent terminal, and even its parent exited, its terminal
will not close as the children running.

So this commit will try to fix this issue by detecting
the process's exit and then notify the client that the
process has exited and there's no need to wait on its IO.

Fixes: #370

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -69,6 +69,7 @@ type process struct {
 	stderr      *os.File
 	consoleSock *os.File
 	termMaster  *os.File
+	epoller     *epoller
 	exitCodeCh  chan int
 	sync.Once
 	stdinClosed bool
@@ -174,6 +175,10 @@ func (p *process) closePostExitFDs() {
 
 	if p.stderr != nil {
 		p.stderr.Close()
+	}
+
+	if p.epoller != nil {
+		p.epoller.sockR.Close()
 	}
 }
 
@@ -368,7 +373,17 @@ func (s *sandbox) readStdio(cid, execID string, length int, stdout bool) ([]byte
 
 	var file *os.File
 	if proc.termMaster != nil {
-		file = proc.termMaster
+		// The process's epoller's run() will return a file descriptor of the process's
+		// terminal or one end of its exited pipe. If it returns its terminal, it means
+		// there is data needed to be read out or it has been closed; if it returns the
+		// process's exited pipe, it means the process has exited and there is no data
+		// needed to be read out in its terminal, thus following read on it will read out
+		// "EOF" to terminate this process's io since the other end of this pipe has been
+		// closed in reap().
+		file, err = proc.epoller.run()
+		if err != nil {
+			return nil, err
+		}
 	} else {
 		if stdout {
 			file = proc.stdout
--- a/epoll.go
+++ b/epoll.go
@@ -1,0 +1,113 @@
+// Copyright (c) 2018 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+	"google.golang.org/grpc/codes"
+	grpcStatus "google.golang.org/grpc/status"
+)
+
+const maxEvents = 2
+
+type epoller struct {
+	fd int
+	// sockR and sockW are a pipe's files two ends, this pipe is
+	// used to sync between the readStdio and the process exits.
+	// once the process exits, it will close one end to notify
+	// the readStdio that the process has exited and it should not
+	// wait on the process's terminal which has been inherited
+	// by it's children and hasn't exited.
+	sockR   *os.File
+	sockW   *os.File
+	sockMap map[int32]*os.File
+}
+
+func newEpoller() (*epoller, error) {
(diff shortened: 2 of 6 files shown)


agent.go
```diff
@@ -368,7 +373,10 @@ func (s *sandbox) readStdio(cid, execID string, length int, stdout bool) ([]byte
 
 	var file *os.File
 	if proc.termMaster != nil {
-		file = proc.termMaster
+		file, err = proc.epoller.run()
```
bergwolf: IMO it worth a comment here to state that when `run()` returns, either the pipe is closed or the terminal is readable. And if `file` returned here is `sockR` (a.k.a process termination), `readStdio` returns EOF.

epoll.go
```diff
@@ -0,0 +1,108 @@
...
+			// any of the requested events occurred or the timeout expired
+			if err == unix.EINTR {
+				continue
+			}
+			return nil, err
+		}
+
+		for i := 0; i < n; i++ {
+			ev := &events[i]
+			if ev.Fd != fd {
```
bergwolf: Please add a comment here to explain that we do this check to make sure terminal fd is picked over pipe fd, so that client can still read terminated process' stdio.

epoll.go
```diff
@@ -0,0 +1,108 @@
...
+	// notification. There will be three cases on the epoller once it run:
+	// a: only pty's master get an event;
+	// b: only the pipe get an event;
+	// c: both of pty and pipe have event occur;
+	// for case a, it means there is output in process's terminal and what needed to do is
+	// just read the terminal and send them out; for case b, it means the process has exited
+	// and there is no data in the terminal, thus just return the "EOF" to end the io;
+	// for case c, it means the process has exited but there is some data in the terminal which
+	// hasn't been send out, thus it should send those data out first and then send "EOF" last to
+	// end the io.
```
bergwolf: Please move the comments for the three cases to `epoller.run()`. The add function does not differentiate any of the cases.

epoll.go
```diff
@@ -0,0 +1,108 @@
...
+
+const maxEvents = 2
+
+type epoller struct {
+	fd int
+	// sockR and sockW are a pipe's files two ends, this pipe is
+	// used to sync between the readStdio and the process exits.
+	// once the process exits, it will close one end to notify
+	// the readStdio that the process has exited and it should not
+	// wait on the process's terminal which hasn't been inherited
```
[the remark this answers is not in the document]
lifupan: Has fixed.

agent.go
lifupan: Has added some comments here.

epoll.go
lifupan: Done.

epoll.go
lifupan: Done.

reaper.go
```diff
@@ -62,6 +67,32 @@ func (r *agentReaper) unlock() {
 	r.RUnlock()
 }
 
+func (r *agentReaper) getEpoller(pid int) (*epoller, error) {
```
jodh-intel: Nit: you could drop the `get` prefix - it's descriptive but not necessary according to:

- https://golang.org/doc/effective_go.html#Getters

lifupan: Hi @jodh-intel , yes, this "get" prefix is redundant, but in order to keep align with the reaper's
other methods such as getExitCodeCh, Can I  keep it here?

lifupan:
Hi @jodh-intel,  I can add a real test in https://github.com/kata-containers/tests  later.

reaper.go
jodh-intel: Hi @lifupan - Fine with me - any "nits" are just tiny issues that shouldn't block the PR landing imho ;)

jodh-intel:
Thanks @lifupan ! ;)

sboeuf:
I'd like to review this properly, please don't merge until I can get a look :)

lifupan:
Hi @sboeuf, the client is kata shim and the containerd shimv2. Take kata shim as example, you can have a look at https://github.com/kata-containers/shim/blob/master/main.go from line 149 to 159, the comments say clearly that this shim has to wait  for every input/output to return before the code tries to go further and wait for the process.

egernst:
@lifupan @bergwolf  -  you guys agree this should be a candidate for stable branches as well?

lifupan:
Hi  @egernst , yes, I think the stable branch needs this patch.

lifupan:
Hi @sboeuf , I think the libcontainer doesn't needs this fix, and this issue is particular to kata. 
It's easy to close process's IO when it exits, but the problem is when to close it after it's exiting to make sure there is no IO contents lost. For runc, containerd shim  will take care of the  process's IO read and the process's exit states. In addition, the shim is the IO read's initiator, and it can make sure
to drain the process's IO before closing it. But for kata, kata's IO read isn't a stream and it depends kata's client calling to drain the IO,  thus it cannot close the IO proactively. Thus kata agent have to figure out is the IO is exhausted  after the process's exiting and then close the IO. 

BTW, I can add some more info about how this PR fix this issue in the commit log.

PR #375, opened 2018-09-19
wip: Revert "device: check for existing device PCI path before waiting"

amshinde:
We should not check for the sysfs path assciated with the PCI address
as the device node may still not be created. Only a uevent with a device
name associated indicates that the device is ready.

This change was also causing the sandbox lock to be not released causing
the uevent listening loop to be locked permanently.

This reverts commit 7caf1c8bd9ebad9ab6631e1839d62c8c98ed7d8a.

### DIFF
--- a/device.go
+++ b/device.go
@@ -112,23 +112,14 @@ func getPCIDeviceName(s *sandbox, pciID string) (string, error) {
 
 	fieldLogger := agentLog.WithField("pciID", pciID)
 
-	s.Lock()
 	// Check if the PCI identifier is in PCI device map.
+	s.Lock()
 	for key, value := range s.pciDeviceMap {
 		if strings.Contains(key, pciAddr) {
 			devName = value
 			fieldLogger.Info("Device found in pci device map")
 			break
 		}
-	}
-
-	// Check if the PCI path is present before we setup the pci device map.
-	_, err = os.Stat(filepath.Join(rootBusPath, pciAddr))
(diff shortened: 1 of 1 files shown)


sboeuf:
@amshinde is this PR solving the issue regarding the inconsistent results when hotplugging some extra network?

device.go
```diff
@@ -122,15 +122,6 @@ func getPCIDeviceName(s *sandbox, pciID string) (string, error) {
 		}
 	}
 
-	// Check if the PCI path is present before we setup the pci device map.
-	_, err = os.Stat(filepath.Join(rootBusPath, pciAddr))
-	if err == nil {
```
bergwolf: What's missing is `s.Unlock()` here. Could you please add it here instead of reverting the whole commit? It does fix a situation that is not handled in the original code.

amshinde:
@sboeuf No there are still issues @bergwolf is seeing. Wasnt able to reproduce the issue on my end yesterday.

device.go
amshinde: @bergwolf See my comment in the commit message for this. We should be taking out this check.
For eg, checking for /sys/devices/pci0000\:00/0000:00:04.0 is not sufficient, the device is fully attached when virtio# directory under that path  (/sys/devices/pci0000\:00/0000:00:04.0/virtio3/device). is created which we can only deduce with the uevent.

PR #376, opened 2018-09-19
agent: support agent as init process in rootfs images

### LINKED ISSUE #264: consider allowing agent to run as PID 1 in systemd-based rootfs images
Currently, if you build a rootfs using `AGENT_INIT=yes` using an initrd distro (such as centos), the agent will fail at startup as it cannot mount `/dev` (as it's already mounted).

We could consider adding a magic `agent.no-dev-mount=true` kernel command-line option to allow us to test the agent as PID 1 in rootfs images other than alpine.

wdyt @bergwolf?

devimc:
Don't fail if `dev` is already mounted

fixes #264

cc @jodh-intel @bergwolf 

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -847,7 +847,11 @@ func mountToRootfs(m initMount) error {
 func generalMount() error {
 	for _, m := range initRootfsMounts {
 		if err := mountToRootfs(m); err != nil {
-			return err
+			// dev is already mounted if the rootfs image is used
+			if m.src != "dev" {
+				return err
+			}
+			agentLog.WithError(err).WithField("src", m.src).Warnf("Could not mount filesystem")
 		}
 	}
 	return nil


agent.go
```diff
@@ -847,7 +847,9 @@ func mountToRootfs(m initMount) error {
 func generalMount() error {
 	for _, m := range initRootfsMounts {
 		if err := mountToRootfs(m); err != nil {
-			return err
+			// Don't fail and try to mount the rest of filesystems
+			agentLog.Warnf("Could not mount %v: %v", m.src, err)
```
bergwolf: If we are sure `/dev` is the only exception, can you filter it out and still fail for all the other cases?

grahamwhaley:
CI saw some net timeout - but not nudging, as I'm predicting there will be an update/repush which will trigger a CI re-run.

devimc:
@jodh-intel @bergwolf  changes applied, thanks

jodh-intel:
lgtm

bergwolf:
`02:33:20 ERROR: command perf not available`

metrics CI is broken today and the issue is tracked at https://github.com/kata-containers/tests/issues/763. Let's ignore it and merge the PR.

Issue #372, opened 2018-09-17
uevent listening goroutine is not (always) working


bergwolf:
As I was adding network hotplug support for vm factory and found out that the uevent listening goroutine is not working for me. After hotplugging a new nic, however long I wait in `UpdateInterface`, the uevent listening goroutine cannot get the new device sysfs uevent.

Then I tested the network monitor process (kata-containers/runtime#534), and it failed on getting the new uevent as well:
> Sep 13 09:37:20 macbeth kata-proxy[54869]: time="2018-09-13T09:37:20.532400733+08:00" level=info msg="time="2018-09-13T01:37:20.511006103Z" level=debug msg="new request" name=kata-agent pid=128 req="interface:<device:\"eth1\" name:\"eth1\" IPAddresses:<family:2 address:\"172.18.0.2\" mask:\"16\" > mtu:1500 hwAddr:\"02:42:ac:12:00:02\" pciAddr:\"02/01\" > " request=/grpc.AgentService/UpdateInterface sandbox=e2145826daa97ea538d49bffc8c25ad13e0849a1edf5546478458fa17aa63002 source=agent\n" name=kata-proxy pid=54869 sandbox=e2145826daa97ea538d49bffc8c25ad13e0849a1edf5546478458fa17aa63002 source=agent
Sep 13 09:37:20 macbeth kata-proxy[54869]: time="2018-09-13T09:37:20.533607734+08:00" level=info msg="time="2018-09-13T01:37:20.512336531Z" level=info msg="Fetched PCI address for device" completePCIAddr="0000:00:02.0/0000:01:01.0" name=kata-agent pid=128 sandbox=e2145826daa97ea538d49bffc8c25ad13e0849a1edf5546478458fa17aa63002 source=agent\n" name=kata-proxy pid=54869 sandbox=e2145826daa97ea538d49bffc8c25ad13e0849a1edf5546478458fa17aa63002 source=agent
[...]

PR #380, opened 2018-09-20
fix up remaining issues with pci hotplug and uevent listening

bergwolf:
1. rescan pci bus before waiting for a new device
2. fix redefined notify channel when waiting for a new device

Fixes: #372

I made this depends on #375 since they both modify the same function and would naturally conflict.

### DIFF
--- a/device.go
+++ b/device.go
@@ -128,7 +128,7 @@ func getPCIDeviceName(s *sandbox, pciID string) (string, error) {
 	// Note this is done inside the lock, not to miss any events from the
 	// global udev listener.
 	if devName == "" {
-		notifyChan := make(chan string, 1)
+		notifyChan = make(chan string, 1)
 		s.deviceWatchers[pciAddr] = notifyChan
 	}
 	s.Unlock()
--- a/device.go
+++ b/device.go
@@ -112,23 +112,14 @@ func getPCIDeviceName(s *sandbox, pciID string) (string, error) {
 
 	fieldLogger := agentLog.WithField("pciID", pciID)
 
-	s.Lock()
 	// Check if the PCI identifier is in PCI device map.
+	s.Lock()
 	for key, value := range s.pciDeviceMap {
 		if strings.Contains(key, pciAddr) {
 			devName = value
 			fieldLogger.Info("Device found in pci device map")
 			break
 		}
-	}
-
-	// Check if the PCI path is present before we setup the pci device map.
-	_, err = os.Stat(filepath.Join(rootBusPath, pciAddr))
-	if err == nil {
-		// Found pci path. It's there before we wait for the device map.
-		// We cannot find the name for it.
-		fieldLogger.Info("Device found on pci device path")
-		return "", nil
 	}
 
 	// If device is not found in the device map, hotplug event has not
@@ -168,9 +159,6 @@ func virtioBlkDeviceHandler(device pb.Device, spec *pb.Spec, s *sandbox) error {
 	devPath, err := getPCIDeviceName(s, device.Id)
 	if err != nil {
 		return err
-	}
-	if devPath == "" {
-		return fmt.Errorf("cannot find device name for virtio block device")
 	}
 	device.VmPath = devPath
 
--- a/device.go
+++ b/device.go
@@ -30,7 +30,11 @@ const (
 	driverEphemeralType = "ephemeral"
 )
 
-const rootBusPath = "/devices/pci0000:00"
+const (
+	rootBusPath      = "/devices/pci0000:00"
+	pciBusRescanFile = "/sys/bus/pci/rescan"
+	pciBusMode       = 0220
+)
 
 var (
 	sysBusPrefix     = "/sys/bus/pci/devices"
@@ -56,6 +60,10 @@ type deviceHandler func(device pb.Device, spec *pb.Spec, s *sandbox) error
 var deviceHandlerList = map[string]deviceHandler{
 	driverBlkType:  virtioBlkDeviceHandler,
 	driverSCSIType: virtioSCSIDeviceHandler,
+}
+
+func rescanPciBus() error {
+	return ioutil.WriteFile(pciBusRescanFile, []byte{'1'}, pciBusMode)
 }
 
 // getDevicePCIAddress fetches the complete PCI address in sysfs, based on the PCI
@@ -120,6 +128,13 @@ func getPCIDeviceName(s *sandbox, pciID string) (string, error) {
 			fieldLogger.Info("Device found in pci device map")
 			break
 		}
+	}
+
+	// Rescan pci bus if we need to wait for a new pci device
+	if err = rescanPciBus(); err != nil {
+		fieldLogger.WithError(err).Error("Failed to scan pci bus")
+		s.Unlock()
+		return "", err
 	}
 
 	// If device is not found in the device map, hotplug event has not
--- a/device_test.go
+++ b/device_test.go
@@ -443,3 +443,12 @@ func TestUpdateSpecDeviceList(t *testing.T) {
 	err = updateSpecDeviceList(device, spec)
 	assert.NoError(err)
 }
+
+func TestRescanPciBus(t *testing.T) {
+	skipUnlessRoot(t)
+
+	assert := assert.New(t)
+
+	err := rescanPciBus()
+	assert.Nil(err)
+}
--- a/grpc.go
+++ b/grpc.go
@@ -39,12 +39,6 @@ type agentGRPC struct {
 	version string
 }
 
-// PCI scanning
-const (
-	pciBusRescanFile = "/sys/bus/pci/rescan"
-	pciBusMode       = 0220
-)
-
 // CPU and Memory hotplug
 const (
 	cpuRegexpPattern = "cpu[0-9]*"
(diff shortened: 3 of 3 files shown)


jodh-intel:
lgtm

device.go
```diff
@@ -36,6 +36,7 @@ var (
 	sysBusPrefix     = "/sys/bus/pci/devices"
 	pciBusPathFormat = "%s/%s/pci_bus/"
 	systemDevPath    = "/dev"
+	pciBusRescanPath = "/sys/bus/pci/rescan"
```
devimc: This variable is in `grpc.go` `pciBusRescanFile`

device.go
```diff
@@ -122,13 +123,11 @@ func getPCIDeviceName(s *sandbox, pciID string) (string, error) {
...
 
-	// Check if the PCI path is present before we setup the pci device map.
-	_, err = os.Stat(filepath.Join(rootBusPath, pciAddr))
-	if err == nil {
-		// Found pci path. It's there before we wait for the device map.
-		// We cannot find the name for it.
-		fieldLogger.Info("Device found on pci device path")
-		return "", nil
+	// Rescan pci bus if we need to wait for a new pci device
+	if err = ioutil.WriteFile(pciBusRescanPath, []byte{'1'}, 0200); err != nil {
```
devimc: please use `pciBusMode`

devimc: or could you implement a new function here to rescan the PCI bus and use it in `grpc.go` ?

bergwolf:
@devimc updated. PTAL.

sboeuf:
I've just restarted the metrics build.

gnawux:
Ignore the non-related metrics failure.

let's merge it

grahamwhaley:
Thanks @gnawux - already aware of the perf item - we have an issue over at https://github.com/kata-containers/ci/issues/43 as well.
But, that is not an actual fail for the metrics right now - it stops us running a test (network performance), that we are not actually using as a pass/fail metric yet (so, yes, it is annoying and distracting that the error occurs, but it is not the actual failure problem). Let me go see if I can find the Jenkins logs to check (as once you have closed the PR, the links vanish from the github view ;-) ).... here we go: http://jenkins.katacontainers.io/job/kata-metrics-agent-ubuntu-16-04-PR/44/console
Wow, metrics was having a bad day :-(
The important bit of the metrics is the bounds check at the end of the logs - first thing to do with a metrics log is to check if we got to the end and did the check happen. Then you find a table like:

```
03:36:43 Report Summary:
03:36:43 +-----+----------------------+-------+--------+--------+-------+--------+--------+------+------+-----+
03:36:43 | P/F |         NAME         |  FLR  |  MEAN  |  CEIL  |  GAP  |  MIN   |  MAX   | RNG  | COV  | ITS |
03:36:43 +-----+----------------------+-------+--------+--------+-------+--------+--------+------+------+-----+
03:36:43 | *F* | boot-times           | 95.0% | 110.2% | 105.0% | 10.0% | 108.1% | 112.6% | 4.2% | 1.1% |  20 |
03:36:43 | *F* | memory-footprint     | 95.0% | 87.4%  | 105.0% | 10.0% | 87.4%  | 87.4%  | 0.0% | 0.0% |   1 |
03:36:43 | P   | memory-footprint-ksm | 95.0% | 97.2%  | 105.0% | 10.0% | 97.2%  | 97.2%  | 0.0% | 0.0% |   1 |
03:36:43 +-----+----------------------+-------+--------+--------+-------+--------+--------+------+------+-----+
```

What I've been seeing over the last week is too much variance in the metrics results, and I'm not sure why yet. I would not normally expect either memory footprint or boot times to vary so much - which is why I have the pass/fail boundaries set at 5% - but we see here boot times were up 10%, and memory footprint down 12.6%. I need to do some research into what is going on. I just didn't find time yet this week.

This is a key reason why the metrics CI's are not marked as 'required' items in Github yet...
Thanks for you patience ;-)

sboeuf:
@bergwolf @amshinde @egernst @devimc 
I have spent some time understand what was going on with `pc`, `q35` and the different ways the PCI hotplug is being performed.

__`pc` case__
- This machine type uses ACPI to hotplug PCI devices. This means the whole notification mechanism relies on ACPI methods and it works well when hotplugging on the main bus (pci host bridge).
- But why it does not work to hotplug on `pci-bridge` if `shpc=off`? The ACPI mechanism should still work, right? Yes, the ACPI mechanism notifying the guest OS about a new PCI device works properly and the PCI slot is getting rescanned. Unfortunately, the PCI spec defines that any PCI bridge that is created (coldplugged) with no device attached to it, will not be getting any memory resource allocated for the futures BARs of its devices. For this reason, an empty PCI bridge is unusable from a PCI hotplug perspective, and the guest OS generates this error:
```
[    8.424258] pci 0000:01:00.0: BAR 6: no space for [mem size 0x00040000 pref]
[    8.425179] pci 0000:01:00.0: BAR 6: failed to assign [mem size 0x00040000 pref]
[    8.425274] pci 0000:01:00.0: BAR 4: no space for [mem size 0x00004000 64bit pref]
[    8.425353] pci 0000:01:00.0: BAR 4: failed to assign [mem size 0x00004000 64bit pref]
[    8.425434] pci 0000:01:00.0: BAR 1: no space for [mem size 0x00001000]
[    8.425501] pci 0000:01:00.0: BAR 1: failed to assign [mem size 0x00001000]
[    8.425568] pci 0000:01:00.0: BAR 0: no space for [io  size 0x0020]
[    8.425635] pci 0000:01:00.0: BAR 0: failed to assign [io  size 0x0020]
[    8.472148] PCI Interrupt Link [LNKA] enabled at IRQ 11
[    8.472498] virtio-pci 0000:01:00.0: virtio_pci: leaving for legacy driver
[    8.493270] virtio-pci: probe of 0000:01:00.0 failed with error -12
```
- But then, how does it work with `shpc=on`? Well, by enabling `shpc` on the PCI bridge, we create a `shpc` controller on the slot 0 of this PCI bridge, and this will allocate some `shpc` BAR, which will force the bridge not to be empty, hence usable for later hotplug. Here is the output from the kernel when the PCI device is properly added:
```
[    8.512248] pci 0000:01:00.0: BAR 6: assigned [mem 0x80000000-0x8003ffff pref]
[    8.512417] pci 0000:01:00.0: BAR 4: assigned [mem 0x80040000-0x80043fff 64bit pref]
[    8.512597] pci 0000:01:00.0: BAR 1: assigned [mem 0x80044000-0x80044fff]
[    8.512795] pci 0000:01:00.0: BAR 0: assigned [io  0xc000-0xc01f]
[    8.535774] PCI Interrupt Link [LNKA] enabled at IRQ 11
[    8.536077] virtio-pci 0000:01:00.0: enabling device (0000 -> 0003)
[    8.558381] virtio_net virtio3 enp1s0: renamed from eth0
[    8.597475] IPv6: ADDRCONF(NETDEV_UP): enp1s0: link is not ready
[    9.564367] IPv6: ADDRCONF(NETDEV_CHANGE): enp1s0: link becomes ready
```
So that's the explanation for `pc` and why we need shpc to be enabled on any `pci-bridge` that we create.

__Note:__ coldplugging a PCI device to the PCI bridge will have the same effect, forcing the PCI bridge to have it's own BAR space, and further hotplug will work (unless you're hotplugging a PCI device with large BARs).

__`q35` case__
- There is no ACPI involved in the handling of PCI hotplug for the case of `q35`. It's either relying on `shpc` (for PCI devices) or PCIe native hotplug.
- Because we use a `pci-bridge`, `shpc` is going to be used in this case. But it will be used fully, not only for the "PCI bridge trick". It will provide the mechanism to communicate with the guest OS and that's why the guest OS needs to support it (that's why we enabled `CONFIG_HOTPLUG_PCI_SHPC`).
- Problem is about the way `shpc` works and the assumptions from the shpc driver `drivers/pci/hotplug/shpchp_ctrl.c`. When the driver detects the device has been hotplugged, it will eventually call into `handle_button_press_event()` which will delay queuing the work for 5 seconds.... Yeah 5 seconds! We can understand this from a real HW perspective, but for virtualization, it does not make sense... Unless we patch the driver itself (and I don't think we should do that), the PCI rescan looks like the best optimization we can use here. Indeed, the PCI device is already there, but we're simply waiting for the driver to resume its work.


I hope this helps clarify things :)

amshinde:
@sboeuf Thanks for summarising our findings. In conclusion:
It always makes sense to leave SHPC on for a pci-bridge.

With the current hotplug implementation for `q35` in virtcontainers with pci-bridge, we need to rescan the pci bus because of the 5 second delay in the shpc controller. 
Since this is not required for `pc`, we can do better and pass an extra parameter to the agent telling it to scan the bus or not. I am not sure how expensive scanning the bus would get, when large number of devices are attached.

Another point to consider with q35 hotplug is that by  attaching pcie devices to a pci-bridge we lose out on the pcie capabilities of the device, since the device is exposed as a pci device. We lose out on all the advantages of pcie which we would get with `q35` platform. Hence, what we really need to do is to revisit q35 hotplug in virtcontainers and implement the hotplug using pcie root ports instead of a pci-bridge.

cc @devimc

bergwolf:
@sboeuf Thanks for the detailed summary! And I agree with @amshinde that we should move to pcie ports for `q35` since `pcie` support is the main advantage of `q35` platform.

gnawux:
thanks @sboeuf , agree to move to `q35` as well.

PR #384, opened 2018-09-25
grpc: Add agent details to GetGuestDetails gRPC function

### LINKED ISSUE #381: Create GetAgentFeatures gRPC call
It would be useful if we added a `GetAgentFeatures()` gRPC call:

```protobuf
message AgentFeaturesResponse {
    bool supports_seccomp = 1;
};

rpc GetAgentFeatures(AgentFeaturesRequest) returns (AgentFeaturesResponse);
```

This call could be used by the runtime to determine the capabilities of the agent. For seccomp, that would allow the runtime to either send the seccomp profile (if the agent supports it), or potentially to remove the seccomp profile if it doesn't.

Along with showing whether the agent was built with seccomp support, we could also add the same details to `AgentFeaturesResponse` that are displayed by the [`announce()`](https://github.com/kata-containers/agent/blob/master/agent.go#L683..L702) call.

Alternative: if we want to minimise new calls, we do already have [`GetGuestDetails()`](https://github.com/kata-containers/agent/pull/358), so that could be extended to also return agent details.

Related: https://github.com/kata-containers/runtime/pull/689

jodh-intel:
Create a new `AgentDetails` message and return it inside the message returned by `GetGuestDetails()`. The agent details provided are similar to those logged by the agent at startup in it's "announce" message.

Fixes #381.


jodh-intel:
ah - need to update the mock tests...

jodh-intel:
Branch updated.

/test.

bergwolf:
Can we combine this into `GetGuestDetails` and then runtime can query once and get all the info it wants.

jodh-intel:
@bergwolf - I was waiting for someone to say that :) Yes, let's try to minimise the overall number of calls we provide and the number of calls the runtime needs to make.

I'll rework and post a comment once done...

jodh-intel:
Hi @bergwolf - branch updated and re-tested.

Note that I didn't add a `bool` to the `GuestDetailsRequest` message to conditionally return information on the agent when the caller calls `GetGuestDetails()`. I'm not really sure we even need the existing `GuestDetailsRequest.mem_block_size bool` really - it won't be that expensive as although it's opening a file, it's in `/sys/`. Maybe we could remove that and only add such `bool`'s if we know a particular check is going to be slow?

jodh-intel:
One more ack... :)

caoruidong:
lgtm!

Issue #388, opened 2018-09-27
The code online new added memory


BetaXOi:
https://github.com/kata-containers/agent/blob/042c3ebd71c2ca425aa0c70fadcbf6370116659b/agent.go#L541

`uEv.DevPath` is "/devices/system/memory/***" but `sysfsMemOnlinePath` has "/sys" in head. So the condition will never true, is it a bug?
In addition, the code listen udev event and process memoy online here looks like duplicate with `onlineMemResources`.

Issue #387, opened 2018-09-27
Write unit tests for onlineResources


jodh-intel:
The `onlineResources()` function needs unit tests.

PR #392, opened 2018-09-28
memory: Fix update memory path.

jcvenegas:
When memory is hot added the udevpath is:

 uevent-devpath=/devices/system/memory/memory81

But when the agent checks for the path it expects it  checks
 in the /sys filesystem.

Fixes: #388

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -44,7 +44,8 @@ const (
 
 var (
 	// cgroup fs is mounted at /sys/fs when systemd is the init process
-	cgroupPath                   = "/sys/fs/cgroup"
+	sysfsDir                     = "/sys"
+	cgroupPath                   = sysfsDir + "/fs/cgroup"
 	cgroupCpusetPath             = cgroupPath + "/cpuset"
 	cgroupMemoryPath             = cgroupPath + "/memory"
 	cgroupMemoryUseHierarchyPath = cgroupMemoryPath + "/memory.use_hierarchy"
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -541,9 +541,9 @@ func (s *sandbox) listenToUdevEvents() {
 			}
 
 			s.Unlock()
-		} else if strings.HasPrefix(uEv.DevPath, sysfsMemOnlinePath) {
+		} else if strings.HasPrefix("/sys"+uEv.DevPath, sysfsMemOnlinePath) {
```
sboeuf: Two comments here:
- Please define a `var` or `const` for `/sys`, something like `var sysfsDir = "/sys"` that can be reused across the whole agent code.
- Use `filepath.Join()` if you want to concatenate `/sys` with the uevent `DevPath`.

jodh-intel: Yep - I'd consider using a `var` so we can make it testable - see #387.

jcvenegas:
@bergwolf  I wonder if depends of the kernel config?  We dont have a test to check this I can add a test that does stress-ng to use something near that we requested.

agent.go
```diff
@@ -541,9 +542,9 @@ func (s *sandbox) listenToUdevEvents() {
 			}
 
 			s.Unlock()
-		} else if strings.HasPrefix(uEv.DevPath, sysfsMemOnlinePath) {
+		} else if strings.HasPrefix(sysfsDir+uEv.DevPath, sysfsMemOnlinePath) {
```
sboeuf: You forgot to address the comment where I was saying that `filepath.Join()` is safer than `+`.

jcvenegas: ouch! updating now.

PR #393, opened 2018-10-02
agent: add support for guest-hooks (v2)

flx42:
Adds support for running OCI hooks within the guest. A 'drop-in'
path (guest_hook_path) is specified in the cli configuration file
and if set, the agent will look for OCI hooks in this directory
and inject them into the container life cycle.

Fixes: #348
Replaces: #365 

Changes compared to  #365:
* Rebased on current master
* Fixed the `&spec`
* Added more logging when discovering hooks, no error is fatal today
* Moved utility functions from `grpc` package to ` main` package, with a new file `oci.go`

@jodh-intel


oci.go
```diff
@@ -0,0 +1,113 @@
...
+	files, err := ioutil.ReadDir(hooksPath)
+	if err != nil {
+		agentLog.WithError(err).Infof("Skipping hook type %s", hookType)
+		return
+	}
+
+	for _, file := range files {
+		name := file.Name()
+		if ok, err := isValidHook(file); !ok {
+			agentLog.WithError(err).Warnf("Skipping hook %s", name)
```
jodh-intel: This could become:

```
agentLog.WithError(err).WithField("oci-hook-name", name).Warn("Skipping hook")
```

oci.go
```diff
@@ -0,0 +1,113 @@
...
+	return true, nil
+}
+
+// findHooks searches guestHookPath for any OCI hooks for a given hookType
+func findHooks(guestHookPath, hookType string) (hooksFound []specs.Hook) {
+	hooksPath := path.Join(guestHookPath, hookType)
+
+	files, err := ioutil.ReadDir(hooksPath)
+	if err != nil {
+		agentLog.WithError(err).Infof("Skipping hook type %s", hookType)
```
jodh-intel: This could become:

```
agentLog.WithError(err).WithField("oci-hook-type", hookType).Info("Skipping hook type")
```

oci.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	for _, file := range files {
+		name := file.Name()
+		if ok, err := isValidHook(file); !ok {
+			agentLog.WithError(err).Warnf("Skipping hook %s", name)
+			continue
+		}
+
+		agentLog.Infof("Adding hook %s of type %s", name, hookType)
```
jodh-intel:
```
    "oci-hook-name" : name,
    "oci-hook-type" : hookType,
}).Info("Adding hook")
```

oci.go
```diff
@@ -0,0 +1,113 @@
...
+// writeSpecToFile writes the container's OCI spec to 'config.json'
+// in the bundle path as expected by the OCI specification.
+func writeSpecToFile(spec *specs.Spec) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	configPath := path.Join(cwd, ociConfigFile)
+	f, err := os.OpenFile(configPath, os.O_RDWR|os.O_CREATE, ociConfigFileMode)
```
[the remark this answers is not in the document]
flx42: And maybe `O_EXCL` too?

agent_test.go
```diff
@@ -446,3 +449,52 @@ func TestGetCgroupMountsSuccessful(t *testing.T) {
...
+
+func TestAddGuestHooks(t *testing.T) {
+	assert := assert.New(t)
+
+	hookPath, err := ioutil.TempDir("", "hooks")
+	assert.NoError(err)
+	defer os.RemoveAll(hookPath)
+
+	poststopPath := path.Join(hookPath, "poststop")
+	err = os.Mkdir(poststopPath, 0777)
```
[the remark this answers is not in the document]
flx42: I can tighten the permissions sure, but I do like the octal format since it's immediately readable for everyone.

oci.go
flx42: I think this is a nice example of why I don't like this logging model, but I will do the change :)

oci.go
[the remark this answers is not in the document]
flx42: Ok, maybe not, I'm not convinced this path can't be called multiple times.

oci_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestChangeToBundlePath(t *testing.T) {
```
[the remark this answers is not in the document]
flx42: After re-examining the code, I've decided to change the logic we have in `grpc.go`.
Instead of doing `changeToBundlePath` and then `writeSpecToFile`, I've swapped the order of these operations. The concept of bundle doesn't make sense if we don't have `config.json` serialized in a directory. And the spec struct in the Go code does not defined where the bundle should be. 

I think it makes more sense to first make `dirname(rootfs)` a valid bundle (by convention, could be any other directory), then `chdir` to this bundle.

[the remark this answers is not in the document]
flx42: So the conlusion is that we don't need to treat absolute and relative paths differently. If `spec.Root.Path` is the recommended `rootfs`, then `.` will become an OCI bundle, and that's fine.

flx42:
All of the above should have been addressed now, great feedback @jodh-intel :)

I realize I didn't ask you if amending the existing commit is actually fine, or if you prefer squashing everything at the end.

oci_test.go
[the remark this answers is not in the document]
flx42: notifying @eguzman3 in case he notices something wrong with my reasoning :)

grahamwhaley:
Hi @flx42 - we normally ammend and force push - so looks like you chose the correct flow :-)
Hmm, we used to have that in the contribution guideline/docs, but I don't see it right now. Maybe it fell off during a docs rework... I'll go look harder.

grahamwhaley:
Aha, here we go: https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#normal-pr-workflow

grahamwhaley:
The metrics CI failure here is because we have merged https://github.com/kata-containers/tests/pull/785, so I now need to go re-configure the build slave to take that into account. I'll jump on that.

grahamwhaley:
Metrics slave config updated. I re-span the build, and that has now passed.

agent_test.go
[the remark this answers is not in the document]
jodh-intel: Oh sure - octal is good :)

agent.go
```diff
@@ -240,6 +243,38 @@ func (s *sandbox) setSandboxStorage(path string) bool {
...
 }
 
+// scanGuestHooks will search the given guestHookPath
+// for any OCI hooks
+func (s *sandbox) scanGuestHooks(guestHookPath string) {
+	agentLog.Infof("Scanning guest filesystem for OCI hooks at %s", guestHookPath)
+
+	s.guestHooks.Prestart = findHooks(guestHookPath, "prestart")
+	s.guestHooks.Poststart = findHooks(guestHookPath, "poststart")
+	s.guestHooks.Poststop = findHooks(guestHookPath, "poststop")
```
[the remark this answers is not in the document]
jodh-intel: Ack ;)

jodh-intel:
Thanks @flx42!

lgtm

grpc.go
```diff
@@ -591,6 +620,25 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
...
+
+		// write the OCI spec to a file so that hooks can read it
+		err = writeSpecToFile(ociSpec)
+		if err != nil {
+			return emptyResp, err
+		}
+
+		// Change cwd because libcontainer assumes the bundle path is the cwd:
+		// https://github.com/opencontainers/runc/blob/v1.0.0-rc5/libcontainer/specconv/spec_linux.go#L157
+		oldcwd, err := changeToBundlePath(ociSpec)
```
bergwolf: It's worth noting that the code piece works because golang sets CLONE_FS when creating new M (https://github.com/golang/go/blob/2595fe7fb6f272f9204ca3ef0b0c55e66fb8d90f/src/runtime/os_linux.go#L131) so CWD is shared among all os threads.

bergwolf: It's worth noting that the code piece works because golang sets CLONE_FS when creating new M (https://github.com/golang/go/blob/2595fe7fb6f272f9204ca3ef0b0c55e66fb8d90f/src/runtime/os_linux.go#L131) so CWD is shared among all os threads. We rely on this model because the goroutine might be rescheduled to another os thread.

grpc.go
```diff
@@ -1193,6 +1216,12 @@ func (a *agentGRPC) CreateSandbox(ctx context.Context, req *pb.CreateSandboxRequ
 	a.sandbox.running = true
 	a.sandbox.sandboxPidNs = req.SandboxPidns
 	a.sandbox.storages = make(map[string]*sandboxStorage)
+	a.sandbox.guestHooks = &specs.Hooks{}
+	a.sandbox.guestHooksPresent = false
+
+	if req.GuestHookPath != "" {
+		a.sandbox.scanGuestHooks(req.GuestHookPath)
```
bergwolf: The error is still silently ignored. I wonder how a user can find it out later on if `scanGuestHooks` fails here. I think we should at least report it back in `CreateSandboxResponse` and runtime can decide if it wants to bail out or continue using the hooks-missing image.

grpc.go
flx42: That seems like an implementation detail. The runtime could very well store the cwd for each goroutine and then restore it when the goroutine is re-scheduled anywhere.

grpc.go
flx42: Through the logs:
https://github.com/kata-containers/agent/pull/393/files/980023ec62b3144ef9f333209290c6215e295538#diff-2fe34e1b6f7f9aece0af74b1635502eeR259
I thought we agreed that it was fine to just log it for now. But I'm happy to change it if we have consensus.

amshinde: Yeah, I am leaning towards just logging this for now.

bergwolf: Fair enough!

grpc.go
bergwolf: @flx42 I mentioned it because we do depend on the runtime implementation details here. If the Go runtime decides that CWD is an M local property, we are doomed because a goroutine can be rescheduled to another os thread since we do not lock os thread here, and then CWD is messed up in different threads.

bergwolf:
LGTM

jodh-intel:
Re-cranking the CI to see if we get lucky and end up running on a VT-x-enabled system...

jodh-intel:
All passed apart from the metrics CI which is showing a rather bizarre message:

```
09:34:23  > git rev-parse refs/remotes/origin/origin/pr/393/merge^{commit} # timeout=10
09:34:23  > git rev-parse origin/pr/393/merge^{commit} # timeout=10
09:34:23 ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
```

/cc @grahamwhaley

grahamwhaley:
I think that pre-existing metrics CI job is just broke from my previous mis-configuration. Let me try to kick a whole CI cycle here, and see what happens. If we end up with them all passed apart from metrics dying with some odd git merge/branch issue, then let's merge.
/test

jodh-intel:
@chavafg, @mnaser - we're still seeing failures due to jobs being scheduled on non-VT-x systems:

```
time="2018-10-09T19:24:51Z" level=error msg="open /sys/module/kvm_intel/parameters/nested: no such file or directory" arch=amd64 name=kata-runtime pid=16864 source=runtime
open /sys/module/kvm_intel/parameters/nested: no such file or directory
```

Is there any way to bump the priority of this issue please?

jodh-intel:
@grahamwhaley - the metrics CI is still failing for the same reason. I'm wondering if this implies the delta between the main CI code and the metrics code is too wide (can we share more setup code?)

jodh-intel:
The 16.04 CI failed two of its tests:

```
Summarizing 2 Failures:

[Fail] package manager update test check apt-get update [It] should not fail 
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:41

[Fail] CPUs and CPU set updating cpus and cpuset of a running container [It] should have the right number of vCPUs 
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:432

Ran 202 of 214 Specs in 952.725 seconds
FAIL! -- 200 Passed | 2 Failed | 0 Pending | 12 Skipped --- FAIL: TestIntegration (979.86s)
FAIL

Ginkgo ran 1 suite in 16m20.983774498s
Test Suite Failed
```

/cc @devimc

grahamwhaley:
@jodh-intel I'll check the metrics CI job setup against the QA setup on the Jenkins server. We share almost all of the setup/scripts already - the metrics invokes the same [jenkins build script](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh) that the QA CI does. I'll check there has not been any updates to the pre-script env setup (the stuff that lives in the Jenkins UI/xml files) that we have missed on the metrics.

grahamwhaley:
Looks like the Jenkins plugin is too smart, and remembers what it was told to do in the first place (when it was mis-configured).
I've mocked up a new job (a jenkins 'parametised build') by kicking a rebuild of a different PR (395, that has merged already), and editing the details to point at this PR. If that does not work, just skip the metrics CI for this PR. It should be configured correctly for future PRs now.

grahamwhaley:
yay, the hand crafted metrics CI job has acked' the PR ;-)

jodh-intel:
Thanks again for your contributions @flx42 and @eguzman3!

Issue #396, opened 2018-10-15
Sometimes the IO's stdout will hang on shimv2 with kata builtin proxy


lifupan:
# Description of problem

Sometimes the container's IO's stdout will hang, and at the same time the stdout
will go on outputting when you hit the keyboard to input any character.   

The same issue has been fixed in kata-proxy https://github.com/kata-containers/proxy/pull/91
but hasn't been forwarded to kata builtin proxy.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #397, opened 2018-10-15
agent: Fix the issue of stdout hang on builtin proxy

lifupan:
Since the Yamux's keepalive has been disabled both
on the server and the client side, and this brings
a weird issue where the communication between the
proxy and the agent hangs. The same issue has been
fixed in kata proxy by:
"https://github.com/kata-containers/proxy/pull/91".

This commit just cherry-pick that patch here to
fix the same issue on kata builtin proxy.

Fixes: #396


jodh-intel:
lgtm

PR #400, opened 2018-10-24
proto: Split reusable structures into their own package

sboeuf:
Some structures such as Interface and Route are not only used by the
kata-agent since they are generic enough to be reused by kata-runtime
or kata-netmon.

By splitting those structures into their separate package, we prevent
their consumers from importing the whole protocols/grpc package which
pulls a lot of unneeded content.

Depends-on: github.com/kata-containers/runtime#858

Fixes #399

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

PR #402, opened 2018-10-24
pkg: types: Add new field type

### LINKED ISSUE #401: pkg: types: Add a new field type
The Interface structure needs a new field 'type' so that we can properly determine the type of interface being described. This will be helpful for network hotplug, as it will be used to create the appropriate connection between the VM and the interface.

sboeuf:
The Interface structure needs a new field 'type' so that we can
properly determine the type of interface being described.

This will be helpful for network hotplug, as it will be used to
create the appropriate connection between the VM and the interface.

Fixes #401


sboeuf:
Depends on #400

pkg/types/types.proto
```diff
@@ -0,0 +1,44 @@
...
+	repeated IPAddress IPAddresses = 3;
+	uint64 mtu = 4;
+	string hwAddr = 5;
+
+	// pciAddr is the PCI address in the format  "bridgeAddr/deviceAddr".
+	// Here, bridgeAddr is the address at which the bridge is attached on the root bus,
+	// while deviceAddr is the address at which the network device is attached on the bridge.
+	string pciAddr = 6;
+
+	// Type defines the type of interface described by this structure.
```
jodh-intel: Can you give examples (or a complete list?) of the values this field will hold in a comment?

jodh-intel:
Travis seems to have exploded in flames on this change...?

jodh-intel:
Related: #400.

pkg/types/types.proto
sboeuf: Added!

sboeuf:
@jodh-intel @amshinde @devimc @grahamwhaley @caoruidong  This PR is ready now that #400 has been merged.

caoruidong:
lgtm

jodh-intel:
lgtm

bergwolf:
@sboeuf I'm a bit concerned about the new field. It adds an extra field to the grpc message that is not going to be used between runtime and agent. The agent does not care about what type of link is added to the VM. All it sees is a new network device. If the runtime API needs the new `linktype` field, it should do so by defining a new type in the runtime repo based on `agent/pkg/types.Interface` (e.g., via embedded types) instead of appending to `types.Interface` directly.

/cc @gnawux @WeiZhang555 @egernst

pkg/types/types.proto
```diff
@@ -30,6 +30,12 @@ message Interface {
 	// Here, bridgeAddr is the address at which the bridge is attached on the root bus,
 	// while deviceAddr is the address at which the network device is attached on the bridge.
 	string pciAddr = 6;
+
+	// Type defines the type of interface described by this structure.
+	// The expected values are the one that are defined by the netlink
+	// library, regarding each type of link. Here is a non exhaustive
+	// list: "veth", "macvtap", "vlan", "macvlan", "tap", ...
+	string type = 7;
```
bergwolf: Also please do not call it `type` since it's a reserved word in golang. `link_type` might be better here. Sorry for the late response as I was OOO last week.

sboeuf:
@bergwolf yes I see your point. I could define a new package`github.com/kata-containers/runtime/pkg/types/types.go`. Both `kata-netmon`, `kata-runtime` and `virtcontainers` would rely on this package that would create a new structure Interface:
```go
import "github.com/kata-containers/agent/pkg/types"

type Interface struct {
        types.Interface
        LinkType string
}
```

Or is it better to define it as `github.com/kata-containers/runtime/virtcontainers/pkg/types/types.go` ?

sboeuf:
@bergwolf problem with embedded struct that I forgot, is that every time a structure `Interface` is defined like this:
```go
	iface := types.Interface{
		Device:      linkAttrs.Name,
		Name:        linkAttrs.Name,
		IPAddresses: ipAddrs,
		Mtu:         uint64(linkAttrs.MTU),
		HwAddr:      linkAttrs.HardwareAddr.String(),
	}
```
`go build` will fail because it won't recognize the field from the structure...

This will imply a lot of changes for such a small thing.

bergwolf:
@sboeuf yeah, anonymous fields cannot be used directly in struct literals. You have to create explicit embedded struct:
```go
        iface := types.Interface{
                grpctypes.Interface{
                        Device:      linkAttrs.Name,
                        Name:        linkAttrs.Name,
                        IPAddresses: ipAddrs,
                        Mtu:         uint64(linkAttrs.MTU),
                        HwAddr:      linkAttrs.HardwareAddr.String(),
                },
                "macvtap",
        }
```
But they can be accessed directly through `iface.Device` etc.

Are there many places to change? I grepped a bit and only found `convertInterface()` that needs to be changed (except for *_test.go though and `virtcontainers/network.go:generateInterfacesAndRoutes` is between runtime and agent).

```shell
[macbeth@runtime]$git grep types.Interface\{|grep -v \}
netmon/netmon.go:       iface := types.Interface{
netmon/netmon_test.go:  expected := types.Interface{
netmon/netmon_test.go:  iface := types.Interface{
netmon/netmon_test.go:  expected := types.Interface{
virtcontainers/api_test.go:     inf := &types.Interface{
virtcontainers/network.go:              ifc := types.Interface{
virtcontainers/network_test.go: expectedInterfaces := []*types.Interface{
```

bergwolf:
Another option is to make grpc types internal and define virtcontainers' own types and have a converter to translate them. @WeiZhang555 also suggested it last time we had such a conversation (https://github.com/kata-containers/runtime/issues/758#issuecomment-422707263).

sboeuf:
@bergwolf the more I think about it, the more I dislike the fact that we use `types` from the agent as generic ones. The ideal solution would be to define the exact same types from `runtime/virtcontainers/pkg/types` to be reused by `kata-runtime`, `kata-netmon`, and `virtcontainers`. The `types` from the agent does not need to be modified or translated, but it should be only imported from `kata_agent.go`, as this is the Kata agent implementation. And later we could imagine having the agent relying on virtcontainers types.

But clearly, the way it is designed right now does not seem right. virtcontainers being the core of the Kata API, it should define the generic types everybody might need. On the other hand, the agent is specific to the `kata_agent.go` implementation, and should not be the reference for generic structures.

bergwolf:
@sboeuf Yes, I agree. We have two requirements here:
1. The kata runtime APIs are between kata-runtime/kata-netmon/kata-shim-v2 and virtcontainers. And virtcontainers being the core of the API, should be the one defining types used by every one.
2. The agent APIs are between virtcontainers and kata-agent. And kata-agent should be the one defining the types.

So it comes back to @WeiZhang555's initial suggestion, we define two `types` packages one in `runtime/virtcontainers/pkg/types` and another in `kata-agent/pkg/types`. 

> The types from the agent does not need to be modified or translated, but it should be only imported from kata_agent.go, as this is the Kata agent implementation.

The types from agent are used in kata-agent APIs. So runtime still need some kind of translation if it wants to talk to the agent. OTOH, `runtime/virtcontainers/pkg/types` does not need to import `agent/pkg/types` so that other components (kata-runtime/kata-netmon/kata-shimv2 etc) doesn't need to be aware of the agent types.

sboeuf:
@bergwolf 
>The types from agent are used in kata-agent APIs. So runtime still need some kind of translation if it wants to talk to the agent. OTOH, runtime/virtcontainers/pkg/types does not need to import agent/pkg/types so that other components (kata-runtime/kata-netmon/kata-shimv2 etc) doesn't need to be aware of the agent types.

Oh yes of course we'll need a tiny translation function in the `kata_agent.go` to be able to translate virtcontainers types (Interface, Route, IPAddress) into agent types.
I thought we were talking about a translation at the structure level that would run outside of run-time.

Anyway, I'm glad we agree on this one, I'll take care of it!

sboeuf:
Once https://github.com/kata-containers/runtime/pull/867 gets merged, let's revert this PR!

WeiZhang555:
> @sboeuf Yes, I agree. We have two requirements here:

> The kata runtime APIs are between kata-runtime/kata-netmon/kata-shim-v2 and virtcontainers. And virtcontainers being the core of the API, should be the one defining types used by every one.
The agent APIs are between virtcontainers and kata-agent. And kata-agent should be the one defining the types.

Yes, I agree with @bergwolf , we have to do our best to keep agent protocol unchanged, because one tiny change can also bring big damage!

Some duplicate codes with a translation layer is necessary, agent protocol stability is worth some duplicated code and extra effort.

PR #404, opened 2018-10-26
network: Handle default route where gateway is empty

### LINKED ISSUE #403: Routes are not setup properly for ipvlan plugin
While using ipvlan docker plugin in l3 mode, the route table with a runc container is setup as 
```
root@6ad93a9c8180:/# ip route
default dev eth0 scope link 
192.168.120.0/24 dev eth0 proto kernel scope link src 192.168.120.2 
```

While adding the routes with Kata Containers, you get an error:
```
docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not add route dest()/gw()/dev(eth0): one of Dst.IP, Src, or Gw must not be nil: unknown.
```

This is because source, destination and gateway all are null.
We should set the gateway to "0.0.0.0" in this case, as 
`ip route add default dev eth0 scope link` is equivalent to  `ip route add default via 0.0.0.0 dev eth0  scope link`.

amshinde:
In case of ipvlan plugin, the default route has an empty gateway.
Set this to default gateway value before adding this route with
netlink.

Fixes #403

### DIFF
--- a/network.go
+++ b/network.go
@@ -31,6 +31,18 @@ var (
 	errNoRoutes = grpcStatus.Errorf(codes.InvalidArgument, "Need network routes")
 )
 
+const (
+	// ipvlan plugin adds a route of the format "default dev eth0 scope link"
+	// Here since source, dest and gateway are empty, netlink will complain.
+	// Set the gateway address explcitly to handle this, setting this address
+	// is equivalent to the above route.
+
+	defaultV4RouteIP = "0.0.0.0"
+
+	// Use the below address for ipv6 gateway once ipv6 support is added
+	// defaultV6RouteIP = "::"
+)
+
 // Network fully describes a sandbox network with its interfaces, routes and dns
 // related information.
 type network struct {
@@ -536,6 +548,13 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *types.Route, add
 		return grpcStatus.Errorf(codes.Internal, "Could not get link's attributes for device %s", route.Device)
 	}
 
+	// We do not modify the gateway in the list of routes provided,
+	// since we loop through the routes checking for the gateway again.
+	gateway := route.Gateway
+	if route.Dest == "" && route.Gateway == "" && route.Source == "" {
+		gateway = defaultV4RouteIP
+	}
+
 	var dst *net.IPNet
 	if route.Dest == "default" || route.Dest == "" {
 		dst = nil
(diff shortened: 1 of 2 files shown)


amshinde:
cc @mcastelino @WeiZhang555

jodh-intel:
It's a simple enough fix but I still think this could do with a unit test. Since Travis is also failing...

```
network.go:523::warning: cyclomatic complexity 17 of function (*sandbox).updateRoute() is high (> 15) (gocyclo)
```

.. how about doing something like:

```go
func gatewayFromRoute(route *pb.Route) string {
    // ...
}
```

network.go
```diff
@@ -535,6 +547,13 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *pb.Route, add bo
 		return grpcStatus.Errorf(codes.Internal, "Could not get link's attributes for device %s", route.Device)
 	}
 
+	// We do not modify the gateway in the list of routes provided,
+	// since we loop through the routes checking for the gateway again.
+	gateway := route.Gateway
+	if route.Dest == "" && route.Gateway == "" && route.Source == "" {
```
sboeuf: @amshinde 
This looks weird to handle a case where everything is empty with a default behavior of having a gateway `0.0.0.0`.
Why don't we try to solve this by pushing a patch to the ipvlan plugin directly? IMO the plugin should send set the gateway properly instead of being empty. Maybe I'm missing some context though.

amshinde: When the gateway is empty the implicit value for default gateway is 0.0.0.0. We get the empty gateway while scanning the network. Unfortunately we need to provide an explicit value of "0.0.0.0" to netlink. This is what iproute2 does as well when you run the command `ip route add default dev eth0` as well. This is equivalent to calling `ip route add default via 0.0.0.0 dev eth0 scope link`

mcastelino: @sboeuf this has to be done primarily because the netlink library has an explicit check
https://github.com/vishvananda/netlink/blob/1404979ff6478eb5af45f1c77fed8e8f249dfb3f/route_linux.go#L488

But given that
`ip route add default dev enp0s3 scope link` == `ip route add default via 0.0.0.0 dev enp0s3 scope link` and that satisfies the check in the go netlink library this fix will work.

However looking at the iproute2 code https://github.com/shemminger/iproute2/blob/master/ip/iproute.c#L699
Here the gw is not filled explicitly means that the kernel may accept the gateway not being present.

i.e. https://github.com/shemminger/iproute2/blob/master/ip/iproute.c#L750 
The RTA_GATEWAY attribute does not need to be present.

If anything longer term we can push a fix to the upstream netlink library to allow leaving out RTA_GATEWAY

sboeuf: Thanks for the feedback here!
As discussed with @amshinde offline, let's move forward with this PR for now, and submit some parallel code to the upstream netlink library.
Once the netlink code gets merged, let's revendor it, and get rid of this patch.

bergwolf: Sounds good to me.

PR #407, opened 2018-11-03
agent: mount blockdevices in the guest VM

zklei:
This mount blockdevices in the guest VM instead of passing them through

Fixes: #406

Signed-off-by: l00413982 <leizhongkai@huawei.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -257,7 +257,20 @@ func mountStorage(storage pb.Storage) error {
 		return err
 	}
 
-	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
+	var fsType = storage.Fstype
+	if (storage.Driver == driverSCSIType || storage.Driver == driverBlkType) && strings.Contains(storage.Fstype, "bind") {
+		cs := strings.Split(storage.Fstype, "-")
+		if len(cs) == 2 && cs[1] != "" {
+			fsType = cs[1]
+			// here we temporarily discard the bind option,
+			// in order to be able to mount the file system of the block device.
+			// and then reset `storage.Fstype` to "bind" which pass through to the libcontainer pkg.
+			flags = flags &^ flagList["bind"]
+			storage.Fstype = "bind"
+		}
+	}
+
+	return mount(storage.Source, storage.MountPoint, fsType, flags, options)
 }
 
 // addStorages takes a list of storages passed by the caller, and perform the


mount.go
```diff
@@ -257,7 +258,19 @@ func mountStorage(storage pb.Storage) error {
 		return err
 	}
 
-	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
+	var fsType = storage.Fstype
```
bergwolf: Please just set `storage.Fstype` to the proper fstype as detected in kata-runtime. Then the agent doesn't need to specially find out the fstype.

mount.go
```diff
@@ -291,3 +304,52 @@ func addStorages(storages []*pb.Storage, s *sandbox) ([]string, error) {
 
 	return mountList, nil
 }
+
+// refer to https://github.com/kubernetes/kubernetes/blob/v1.12.2/pkg/util/mount/mount_linux.go#L512
```
caoruidong: Need to fix CI failure about comment here

mount.go
zklei: But i think that the `storage.Fstype` was set `bind` and we need to pass it to the `libcontainer` pkg, And if we want mount the blockdevice as a dir, we must specify the real filesystem such as `ext4`,`xfs`, etc.  So i temporarily discard the bind option here and find out the fstype again.

mount.go
zklei: Thank you for your opinion, I'll fix it.

sboeuf:
@zklei I don't think the issue you raised is valid. If you run docker with `runc`, and if you use a block device, here is the expected behavior:
```
$ docker run -it --rm --runtime runc -v /dev/sdb:/mnt/data busybox
/ # ls -la /mnt/
total 0
drwxr-xr-x    2 root     root            18 Nov  5 15:29 .
drwxr-xr-x   13 root     root           148 Nov  5 15:29 ..
brw-rw----    1 root     disk      253,   2 Nov  5 15:27 data
```

Now here is the result you can get with Kata:
```
$ docker run -it --rm --runtime kata-runtime -v /dev/sdb:/mnt/data busybox
/ # ls -la /mnt/
total 0
drwxr-xr-x    2 root     root            18 Nov  5 15:30 .
drwxr-xr-x   13 root     root           148 Nov  5 15:30 ..
brw-rw----    1 root     disk        8,  16 Nov  5 15:30 data
```

Kata Containers behavior has been implemented to fit what's expected with `runc`.

jshachm:
@sboeuf 

First of all, It's a very useful case in production env~  
And considering user's feeling, it's just comfortable for them to using the block device context immediately without mount device in container for another time.

Talking about fit with `runc`, it's completelyright~ 

Why don‘t we add a config flag in `toml`,(default is disable this feature) and let users to decide to use this   
feature? 

/cc @bergwolf @grahamwhaley  @WeiZhang555 @lifupan

WeiZhang555:
@sboeuf 

We discussed this before in https://github.com/kata-containers/runtime/issues/571#issuecomment-418575708 but didn't reach a conclusion, in that comment, I described the scenario in detail.

This is kind of breakage of `runc` compatibility as @bergwolf said (another ref: https://github.com/kata-containers/runtime/pull/882#issuecomment-435880738) but it's useful, I think we can treat this as a special extention from kata.

zklei:
> @sboeuf
> 
> First of all, It's a very useful case in production env~
> And considering user's feeling, it's just comfortable for them to using the block device context immediately without mount device in container for another time.
> 
> Talking about fit with `runc`, it's completely right~ And I agree with you on that point.
> 
> Why not we add a config flag in `toml`,(default is disable this feature) and let users to decide to use this
> feature?
> 
> /cc @bergwolf @grahamwhaley @WeiZhang555 @lifupan

@jshachm Really a good idea, i have updated my PR.

grahamwhaley:
I'm happy with the idea of a config option that is off by default :-)
I suspect we should also add some documentation to this PR somewhere - if we are adding a feature or config option, we need that detailed somewhere. But, I admit, we may not have an appropriate place for that right now in the docs. Let's debate...

jshachm:
@grahamwhaley  
Great that you like this idea. Talking about docs, I walked across the `documentation` repo,  can't find an appropriate place  to place this -_- sign......

Considering the fact that as  [oci-runtime-spec Implementations](https://github.com/opencontainers/runtime-spec/blob/master/implementations.md) shows, `kata-runtime` has been one part of `Runtime (Virtual Machine)`, so it's great to have some unique features not the same as `runc` which make users easier to play with `container`.And the great news is that we have a nice config toml. 

So I suggest to add a total new doc about the usage of config toml which lays in the  section [more-user-guides](https://github.com/kata-containers/documentation#more-user-guides), and talk about the config options.

Notice that , we should point out the options that will make `runtime` not compact with what `runc` did,such as here we get `disable_new_netns`  and we can't support `CNM` network. Also here we get `enable_blk_mount_in_vm`, we no longer support `runc`'s action.

What do you think about this~~ ^_^

mount.go
```diff
@@ -257,7 +257,20 @@ func mountStorage(storage pb.Storage) error {
 		return err
 	}
 
-	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
+	var fsType = storage.Fstype
+	if (storage.Driver == driverSCSIType || storage.Driver == driverBlkType) && strings.Contains(storage.Fstype, "bind") {
+		cs := strings.Split(storage.Fstype, "-")
```
sboeuf: Is the runtime already using this notation?

raravena80:
@zklei ping (from your weekly Kata herder)

jodh-intel:
This is still marked as dnm - is that still correct?

Regarding docs, @grahamwhaley recently created https://github.com/kata-containers/documentation/wiki/UserGuide which we're planning to update, then land on https://github.com/kata-containers/documentation/pull/312.

mount.go
```diff
@@ -257,7 +257,20 @@ func mountStorage(storage pb.Storage) error {
 		return err
 	}
 
-	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
+	var fsType = storage.Fstype
+	if (storage.Driver == driverSCSIType || storage.Driver == driverBlkType) && strings.Contains(storage.Fstype, "bind") {
```
jodh-intel: This new block could be handled by a function call that accepts a `storage *pb.Storage`. That would be a little cleaner, and you could then unit-test this logic.

jodh-intel:
Any update on this @zklei?

ydjainopensource:
@zklei Any progress here?? (from your weekly kata herder.)

jodh-intel:
@zklei - are you still planning to continue work on this?

jodh-intel:
@zklei - please see @sboeuf's comments above. Do you plan to continue working on this?

sboeuf:
@zklei what's the status of this PR? Did you add the equivalent PR on the runtime repo side?

jodh-intel:
Ping @zklei - are you still planning on working on this?

jodh-intel:
Final ping to @zklei. If we don't hear back this week, we'll close this PR (but feel free to re-open if you wish).

mount.go
amshinde: As discussed  earlier, we should also check for the configuration passed from the runtime, the config that decides whether the block device needs to be bind-mounted or mounted to a directory.

grahamwhaley:
@zklei @caoruidong @jshachm - is this a feature you are still interested in having/landing in Kata btw? Let us know, or eventually we will have to close this PR as stagnant. Thanks!

raravena80:
@zklei any updates? Thx!

caoruidong:
I think we can close it for now.

Issue #408, opened 2018-11-06
Build agent as position independent executable (PIE)


marcov:
This is a follow up of kata-containers/runtime#875, refer to that issues for details and motivations.

PR #409, opened 2018-11-06
agent: build as Position-Independent-Executable

marcov:
Build the agent binary as Position-Independent-Executable (PIE) for improved
security and compliancy with distros packaging guidelines.

Fixes: #408

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ ARCH := $(shell go env GOARCH)
 ifeq ($(SECCOMP),yes)
 BUILDTAGS := seccomp
 endif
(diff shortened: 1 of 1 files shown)


marcov:
@grahamwhaley good point.
The reason to go for it is that we still need to generate a binary of the agent to include inside the VM image. If we want to use OBS to generate the image, we'll need to have the agent packaged and use it as a just another package to include in the VM rootfs.

jodh-intel:
lgtm

sboeuf:
Let's merge as the metrics CI failure is not related to this PR.

PR #412, opened 2018-11-15
grpc: sandbox: add container when is fully created.

jcvenegas:
Add container to sandbox list until it is created.

This commit makes sure to add the container to the sandbox until it is
created.

This helps to avoid other go routines uses the container structure with
nil pointers (because is not fully created).

Fixes: #417

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -560,6 +560,8 @@ func (a *agentGRPC) finishCreateContainer(ctr *container, req *pb.CreateContaine
 		return emptyResp, err
 	}
 
+	// Make sure add Container to Sandbox, before call updateSharedPidNs
+	a.sandbox.setContainer(req.ContainerId, ctr)
 	if err := a.updateSharedPidNs(ctr); err != nil {
 		return emptyResp, err
 	}
@@ -605,8 +607,6 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		mounts:          mountList,
 		useSandboxPidNs: req.SandboxPidns,
 	}
-
-	a.sandbox.setContainer(req.ContainerId, ctr)
 
 	// In case the container creation failed, make sure we cleanup
 	// properly by rolling back the actions previously performed.


jodh-intel:
Hi @jcvenegas - are you seeing these `nil` values?

jcvenegas:
@jodh-intel hi yes, seems that is a race condition. When using initrd and https://github.com/kata-containers/runtime/pull/887. This is not the complete fix, I still looking how to fix the race condition.

raravena80:
@jcvenegas ping (from your weekly Kata herder)

grahamwhaley:
Let's have a peek at those CI fails then ;-)
On Centos I see:
```
• Failure [4.686 seconds]
logs
/tmp/jenkins/workspace/kata-containers-agent-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/logs_test.go:12
  logs with docker
  /tmp/jenkins/workspace/kata-containers-agent-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/logs_test.go:26
    check logs functionality
    /tmp/jenkins/workspace/kata-containers-agent-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/logs_test.go:27
      should work [It]
      /tmp/jenkins/workspace/kata-containers-agent-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/logs_test.go:28

      Expected
          <string>: 
      to contain substring
          <string>: hello

      /tmp/jenkins/workspace/kata-containers-agent-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/logs_test.go:30
```

On F27 we get something else:
```
ok 21 ctr stop idempotent
ok 22 ctr caps drop
ok 23 run ctr with image with Config.Volumes
ok 24 # skip (This is not working (Issue https://github.com/kata-containers/tests/issues/714)) ctr oom
not ok 25 ctr /etc/resolv.conf rw/ro mode
# (in test file ctr.bats, line 882)
#   `[ "$status" -eq 0 ]' failed
# 0
```

I can't help suspect they are spurious and unrelated - I'll nudge those two CIs again, and let's keep an eye on them.

grpc.go
```diff
@@ -225,30 +225,47 @@ func (a *agentGRPC) onlineCPUMem(req *pb.OnlineCPUMemRequest) error {
...
+		if err != nil {
 			return handleError(req.Wait, err)
 		}
 	}
 
 	return nil
 }
 
+func updateContainerCpuset(c *container, connectedCpus string, cookies cookie) error {
+	if c == nil {
```
devimc: I guess the race condition is because of sanbox contains nil containers, but now the question is why sandbox contains nil containers? :thinking:

jcvenegas: @devimc the line that fails is https://github.com/kata-containers/agent/blob/master/grpc.go#L228
So my conclusionis that c.container.ID() called when is nil, looking at the code when `a.sandbox.containers` gets a new element, this is not fully created that is the reason I added https://github.com/kata-containers/agent/pull/412/commits/da9e9f6cb61b7d208540f9ea65ee747548bb2e7f this help to not add the contaiener to the list until has all the values we want like c.container field.

devimc: @jcvenegas ohh great!, then these conditions are no more needed? Could you confirm?

bergwolf: Ditto @devimc. We'd better guarantee the container list doesn't have nil pointers instead of trying to catch it everywhere.

jcvenegas:
@bergwolf @jodh-intel @devimc @sboeuf please take a look again

jcvenegas:
Only metrics CI failing

devimc:
@jcvenegas  /re-test  :stuck_out_tongue_winking_eye:

amshinde:
@jcvenegas Are  there are any missing commits here ? Your PR message does not reflect what is present in the commit. 
Also, would like to see some more explanation on the commit you introduced to see how it solves the race condition.

Issue #413, opened 2018-11-16
check both interface name and mac address in updateInterface


hejianet:
# Description of problem

When passes a vf to qemu, the guest kernel driver(eg. mlx5) might
set the mac address. Thus without this patch, the updateInterface
in kata agent side will be failed and vfio-pci device can't be passed
to qemu command line

# Expected result

The sriov vf can be successfully passed to guest kernel and container can find the vf NIC by lspci

# Actual result

the vfio-pci can't be passed to guest by qemu command-line

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #414, opened 2018-11-16
network: check both interface name and mac address in updateInterface

hejianet:
When passes a vf to qemu, the guest kernel driver(eg. mlx5) might
set the mac address. Thus without this patch, the updateInterface
in kata agent side will be failed and vfio-pci device can't be passed
to qemu command line

The fix includes 2 parts:
runtime side: set the mac address to null
agent side check the link name if the mac address is null

After setting null mac address in kata runtime, kata agent needs to
check the interface name if mac address is null.

Fixes: #413

### DIFF
--- a/network.go
+++ b/network.go
@@ -272,10 +272,17 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *types.Interf
 		// Find the interface link from its hardware address.
 		link, err = linkByHwAddr(netHandle, iface.HwAddr)
 		if err != nil {
-			return nil, grpcStatus.Errorf(codes.Internal, "updateInterface: %v", err)
+			return nil, grpcStatus.Errorf(codes.Internal, "updateInterface by HwAddr: %v", err)
+		}
+	} else if iface.Device != "" {
+		fieldLogger.Info("Getting interface from name")
+		// Find the interface link from its name.
+		link, err = netHandle.LinkByName(iface.Device)
+		if err != nil {
+			return nil, grpcStatus.Errorf(codes.Internal, "updateInterface byName: %v", err)
 		}
 	} else {
-		return nil, grpcStatus.Errorf(codes.InvalidArgument, "Interface HwAddr empty")
+		return nil, grpcStatus.Errorf(codes.InvalidArgument, "Interface HwAddr and Name are both empty")
 	}
 
 	// Use defer function to create and return the interface's state in


jodh-intel:
Hi @hejianet - thanks for raising!

Looking at that function, I'm rather surprised it doesn't already fail the cyclomatic complexity tests as it's already pretty long and "hairy" :)

Related to this, how about reworking your change so that `updateInterface()` calls a function like the following which can encapsulate the logic:

```
// Deal with blank name + MAC scenarios
func getLink(iface *types.Interface) (netlink.Link, error)
```

That might also allow this function to be unit-tested although I guess using `lo` wouldn't be sufficient for that.

sboeuf:
@hejianet based on the issue https://github.com/kata-containers/runtime/issues/914 that you raised, you need to reprogram the MAC address of the interface that gets created inside the guest, right?
If that's the case, I don't see how this PR would help, and more than that, you are using the interface name, but this is not something that is known by the runtime running on the host.
/cc @amshinde

network.go
```diff
@@ -272,10 +272,17 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *types.Interf
 		// Find the interface link from its hardware address.
 		link, err = linkByHwAddr(netHandle, iface.HwAddr)
 		if err != nil {
-			return nil, grpcStatus.Errorf(codes.Internal, "updateInterface: %v", err)
+			return nil, grpcStatus.Errorf(codes.Internal, "updateInterface by HwAddr: %v", err)
```
amshinde: I suggest you could do this:
```
if iface.PciAddr != "" {
     name , err := getPCIDeviceName(s, iface.PciAddr)
      // Get link based on name
      link, err = netHandle.LinkByName(name) 
      .....
} else if if iface.HwAddr != "" {
 ....
}

Comment (hejianet) [reply network.go]: @amshinde ok, thanks for the suggestions

**Comment by hejianet** [comment]
Hi all, I digged into the root cause a bit more.
Should kata-containers allow vf NIC mac addr to be "" (null) before being passed through to qemu guest?
If the answer is Yes, may be we need to change the current design. Currently the guest will not know the host pci bdf of the vf NIC.

Even the answer is No, there is still a bug after I set the vf mac addr:
e.g. in my platform, a NIC Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
89:00.0 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
89:00.1 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
the network interfaces:
```
enp137s0f0: flags=4098<BROADCAST,MULTICAST>  mtu 1500
        ether e0:d5:5e:cf:57:fb  txqueuelen 1000  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

enp137s0f1: flags=4098<BROADCAST,MULTICAST>  mtu 1500
        ether e0:d5:5e:cf:57:fc  txqueuelen 1000  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```
after enabling one vf for pf enp137s0f0
```
89:00.0 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
89:00.1 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
89:00.2 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx Virtual Function]
```
a new interface:
```
enp137s0f2: flags=4098<BROADCAST,MULTICAST>  mtu 1500
        ether e0:d5:5e:cf:57:01  txqueuelen 1000  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

the ip link show:
```
4: enp137s0f0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether e0:d5:5e:cf:57:fb brd ff:ff:ff:ff:ff:ff
    vf 0 MAC 00:00:00:00:00:00, spoof checking off, link-state auto, trust off, query_rss off
...
24: enp137s0f2: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether e0:d5:5e:cf:57:01 brd ff:ff:ff:ff:ff:ff
```

So there are *2* mac addr for the new added vf.
After I set the mac addr by "ip l set enp137s0f0 vf 0 mac 12:34:56:12:34:56"
the mac addr of enp137s0f2 will not be changed

**But, enp137s0f2's mac addr will be passed to kata runtime, vf 0 mac addr will be passed to kata agent.** So it will not pass the linkByHwaddr check in the kata agent updateInterface() 
  The workaround is if I unbind the enp137s0f2 and bind it again. the mac addr of enp137s0f2 will be updated and same as vf 0

**Comment by devimc** [comment]
ping @amshinde, @mcastelino.

**Comment by jodh-intel** [comment]
Ping @kata-containers/agent, @amshinde, @mcastelino - any comments?

**Comment by ydjainopensource** [comment]
ping @amshinde @mcastelino

**Comment by egernst** [comment]
@hejianet - can you update the mac address of the VF link itself?

**Comment by jodh-intel** [comment]
Ping @egernst, @mcastelino, @amshinde

**Comment by liwei** [comment]
> > can you update the mac address of the VF link itself?
> 
> Thanks, the command I tried to update the mac address of the VF link is
> "ip l set enp137s0f0 vf 0 mac 12:34:56:12:34:56"
> After that:
> the ip link show result is:
> `4: enp137s0f0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
> link/ether e0:d5:5e:cf:57:fb brd ff:ff:ff:ff:ff:ff
> vf 0 MAC 00:00:00:00:00:00, spoof checking off, link-state auto, trust off, query_rss off
> 
> 24: enp137s0f2: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
> link/ether e0:d5:5e:cf:57:01 brd ff:ff:ff:ff:ff:ff`
> 

It seems vf 0 mac address not get updated after you execute the `ip link set` command?
IIRC, vf MAC should be set before guest driver(the vf driver) get loaded, so guest driver can set the correct MAC address according to it.

> Do you know other way to update the line：
> vf 0 MAC 00:00:00:00:00:00, spoof checking off, link-state auto, trust off, query_rss off

You can also update those fields under sysfs, e.g:

```
/sys/class/net/enp8s0f0/device/sriov/[VF]
```

ydjainopensource:
@hejianet any updates here?

hejianet:
sorry for the delay, will give feedback asap

hejianet:
Hi, all. 
Because we have workaround to bypass this issue i.e. hotplug vfio device twice to update the mac addr, could I close this pr temporarily? It takes too much time to restore the hardware I used before. I will keep 
an eye on this issue in the future after I finish another items.

jodh-intel:
No problem - thanks for the update @hejianet.

Issue #420, opened 2018-11-29
Secomp=yes has no effect when creating a local rootfs/initrd


nitkon:
# Description of problem

SECCOMP variable takes value "no" irrespective of the env variable set when creating rootfs/initrd. 
https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#create-a-local-rootfs-for-initrd-image

PR #421, opened 2018-11-29
Makefile: Decide if agent will be built by seccomp tag

nitkon:
Agent would be built by seccomp tag if SECCOMP is set to
yes. Else it shall be set to no.

Fixes: #420

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,6 @@ INIT := no
 UNIT_DIR := /usr/lib/systemd/system
 
 GENERATED_FILES :=
-# Define if agent will be built by seccomp tag
-SECCOMP := no
 
 ifeq ($(INIT),no)
 # Unit file to start kata agent in systemd systems
@@ -34,7 +32,9 @@ COMMIT := $(if $(shell git status --porcelain --untracked-files=no),${COMMIT_NO}
 VERSION_COMMIT := $(if $(COMMIT),$(VERSION)-$(COMMIT),$(VERSION))
 ARCH := $(shell go env GOARCH)
 ifeq ($(SECCOMP),yes)
-BUILDTAGS := seccomp
+	BUILDTAGS := seccomp
+else
+	SECCOMP=no
 endif
 # go build common flags
 BUILDFLAGS := -buildmode=pie


grahamwhaley:
I have a feeling that test kickoff doesn't work in the 'approve' dialog comments (I think it did that to me yesterday)... so...
/test

nitkon:
I guess the "Minval test failed". Should we restart just the "jenkins-metrics-ubuntu-16-04" test?

grahamwhaley:
minval on memory footprint - which is interesting, as I set that pretty wide:
```
12:03:22 +-----+----------------------+-------+--------+--------+-------+-------+--------+-------+------+-----+
12:03:22 | P/F |         NAME         |  FLR  |  MEAN  |  CEIL  |  GAP  |  MIN  |  MAX   |  RNG  | COV  | ITS |
12:03:22 +-----+----------------------+-------+--------+--------+-------+-------+--------+-------+------+-----+
12:03:22 | P   | boot-times           | 95.0% | 101.7% | 105.0% | 10.0% | 98.8% | 121.4% | 22.8% | 4.5% |  20 |
12:03:22 | *F* | memory-footprint     | 85.0% | 84.2%  | 115.0% | 30.0% | 84.2% | 84.2%  | 0.0%  | 0.0% |   1 |
12:03:22 | P   | memory-footprint-ksm | 95.0% | 98.9%  | 105.0% | 10.0% | 98.9% | 98.9%  | 0.0%  | 0.0% |   1 |
12:03:22 +-----+----------------------+-------+--------+--------+-------+-------+--------+-------+------+-----+
```

So, an interesting outcome. It could be sporadic, but I'll keep an eye on it. I'll nudge just the metrics rebuild and let's see what happens.
So, normally reducing the footprint is a good thing - but when it happens and we don't know why, then, well.... some thought is required. I doubt this was your PR - **unless** - was the previous default maybe always including some seccomp stuff in the image, and now is not - that could maybe reduce the footprint?

nitkon:
@grahamwhaley : Probably it was sporadic. Now its all green!

Issue #422, opened 2018-11-29
sync time when creating new sandbox


bergwolf:
system clock and hwclock might differ due to vm templating. We need to fix the drift by setting system clock from hardware clock since kvmclock should be trusted. Otherwise we fail on issues like kata-containers/runtime#951

PR #423, opened 2018-11-29
agent: sync system clock with hwclock when creating new sandbox

bergwolf:
The two might differ due to vm templating. We need to fix the
drift by setting system clock from hardware clock since kvmclock
should be trusted.

Depends-on: github.com/kata-containers/packaging#260
Depends-on: github.com/kata-containers/osbuilder#213
Fixes: #422

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -853,6 +853,14 @@ func (s *sandbox) stopGRPC() {
 	}
 }
 
+func (s *sandbox) syncTimeHw2Sys() error {
+	cmd := &exec.Cmd{
+		Path: "/sbin/hwclock",
+		Args: []string{"hwclock", "--hctosys"},
+	}
+	return s.subreaper.run(cmd)
+}
+
 type initMount struct {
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -853,6 +853,14 @@ func (s *sandbox) stopGRPC() {
 	}
 }
 
+func (s *sandbox) syncTimeHw2Sys() error {
+	cmd := &exec.Cmd{
+		Path: "/sbin/hwclock",
```
grahamwhaley: heh - do we know we have this in all our rootfs distro variants btw?

bergwolf: oh, yeah, you are right, I'm missing the osbuilder PR! Will send it out soon.

bergwolf: Sent github.com/kata-containers/osbuilder/pull/213 and updated the `Depends-on` tag here.

bergwolf:
replaced by #425 since it would work for all architectures and hypervisors.

PR #429, opened 2018-12-07
# Kata Containers 1.5.0-rc1

katacontainersbot:
- client: do cleanup after UT stop mock server
- Makefile: Decide if agent will be built by seccomp tag

4eca13b client: do cleanup after UT stop mock server
c25288a Makefile: Decide if agent will be built by seccomp tag


jodh-intel:
lgtm

/test

PR #433, opened 2018-12-10
protocols/grpc: implement function to copy files

### LINKED ISSUE #432: Implement a function/request to upload files from host to guest
Implement a function/request to upload files from host to guest, this function/request can be used to replace 9p only when few and small file are shared

devimc:
CopyFileRequest can be used to copy files from the host to the workload
rootfs (guest), this request is intended to copy "static" and small
files like resolv.conf and hosts, but any other file can be copied.

fixes #432


devimc:
cc @egernst

protocols/grpc/agent.proto
```diff
@@ -426,3 +427,10 @@ message StringUser {
 	string gid = 2;
 	repeated string additionalGids = 3;
 }
+
+message UploadFileRequest {
+	string path = 1;
+	string content = 2;
```
sboeuf: I think it'd make more sense to have this as an array of bytes instead of a string since it represents the content of the file.
And I'd use an extra field holding the size of this byte array.

grpc.go
```diff
@@ -1442,3 +1442,15 @@ func (a *agentGRPC) getAgentDetails(ctx context.Context) *pb.AgentDetails {
 
 	return &details
 }
+
+func (a *agentGRPC) UploadFile(ctx context.Context, req *pb.UploadFileRequest) (*gpb.Empty, error) {
+	if err := os.MkdirAll(filepath.Dir(req.Path), os.FileMode(req.DirMode)); err != nil {
+		return emptyResp, err
+	}
+
+	if err := ioutil.WriteFile(req.Path, []byte(req.Content), os.FileMode(req.FileMode)); err != nil {
```
sboeuf: Having the content as an array of bytes would prevent from the cast here, and would be more appropriate.

protocols/grpc/agent.proto
```diff
@@ -55,6 +55,7 @@ service AgentService {
 	rpc OnlineCPUMem(OnlineCPUMemRequest) returns (google.protobuf.Empty);
 	rpc ReseedRandomDev(ReseedRandomDevRequest) returns (google.protobuf.Empty);
 	rpc GetGuestDetails(GuestDetailsRequest) returns (GuestDetailsResponse);
+	rpc UploadFile(UploadFileRequest) returns (google.protobuf.Empty);
```
sboeuf: I have a global question here, it feels like having only `UploadFile()` here is only one part of the solution as it would not allow for the update of those files. I think we should, as part of this PR, make the update being supported too.
I'm not against having both initial upload and update being supported by the same function, but it should be explicit if we go that way, otherwise we can simply have two different functions.

devimc: what is going to be the difference between upload and update function? upload can be used to update a file

egernst: Sorry, wouldn’t doing a second uploadFile call overwrite the first?

egernst: I was thinking the same, the caller would call same again to update.

sboeuf: I'm not saying it's not working (the way it's implemented makes it work) but the naming does not sound explicit to me that you should reuse the same function to update an existing file.

devimc: sigh, what name sounds explicit to you?

sboeuf: We could call it CopyFile, wdyt?

egernst: I don't have a strong opinion either way.  I'll poll some other kata folks around me and provide suggestion :)

protocols/grpc/agent.proto
```diff
@@ -426,3 +427,10 @@ message StringUser {
 	string gid = 2;
 	repeated string additionalGids = 3;
 }
+
+message CopyFileRequest {
+	string path = 1;
```
bergwolf: Please add `offset` and `length` here so that it is possible to send just part of a file. Also please put `content` at last of the struct.

What happens when the file size exceeds the grpc payload limit? If we support offset and length, client can send a file by parts.

devimc: @bergwolf 
I think length is not needed, since we can get it with `len(content)`, is that correct?

bergwolf: @devimc good point!

jodh-intel: Is this the *absolute* guest root context path? In other words, is this API allowing changes to, say, `/sbin/init`? Or is it designed purely to copy files into the workloads rootfs directory?

If it's possible to overwrite the init binary, we'd better disallow that for now until the agent is able to serialise it's state and re-exec itself.

grpc.go
```diff
@@ -1449,3 +1449,26 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
 	}
 	return &gpb.Empty{}, nil
 }
+
+func (a *agentGRPC) CopyFile(ctx context.Context, req *pb.CopyFileRequest) (*gpb.Empty, error) {
+	if err := os.MkdirAll(filepath.Dir(req.Path), os.FileMode(req.DirMode)); err != nil {
+		return emptyResp, err
+	}
+
+	file, err := os.OpenFile(req.Path, os.O_WRONLY|os.O_CREATE, os.FileMode(req.FileMode))
```
jodh-intel: - What happens if the file exists as, say, a socket? Won't this overwrite it as a regular file? Maybe we should check to ensure the types are the same?

- Maybe we should enhance the `getAgentDetails()`gRPC API to include the architecture "just in case" to handle the scenario where the caller wants to send a binary file?

- I don't think we should just blindly create a file with the mode specified. What if this is a long-running container and the `CopyFile()` ends up opening ports somehow or overwriting critical files, or making critical file world-writable...

protocols/grpc/agent.proto
jodh-intel: - What about adding a `DoNotOverwrite bool` so the client can change the default behaviour of overwriting an existing file? It could be called `NoClobber bool` I guess, but "clobber" is not a common word even if English is your native tongue ;)

But maybe the default should be to not overwrite? What is the intent of this change?

grpc.go
```diff
@@ -1449,3 +1449,26 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
 	}
 	return &gpb.Empty{}, nil
 }
+
+func (a *agentGRPC) CopyFile(ctx context.Context, req *pb.CopyFileRequest) (*gpb.Empty, error) {
```
jodh-intel: We need unit tests for this - think security :)

grpc.go
```diff
@@ -1449,3 +1449,26 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
...
+		return emptyResp, err
+	}
+
+	defer file.Close()
+
+	if _, err := file.Seek(req.Offset, 0); err != nil {
+		return emptyResp, err
+	}
+
+	if _, err := file.Write(req.Content); err != nil {
```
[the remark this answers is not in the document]
jodh-intel: It would be safer to write the copied file to a temporary one, then atomically move it over the top of the existing one, overwriting the original. The downside being we'll temporarily need more disk space to retain two copies. We'll also have to take care to ensure the temp copy is written in the same directory as the original to avoid any FS-crossing performance issues.

protocols/grpc/agent.proto
devimc: :nerd_face: according with g,xA clobber means *hit (someone) hard. "if he does that I'll clobber him!"*

jodh-intel: Yeah, probably best not to use that term! No NoClobber I say! :smile:

grpc.go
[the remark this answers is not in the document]
devimc: sounds good, but then we'll need two commands `CopyFile` and `MoveFile` because `CopyFile` doesn't know when the temporarily file should be renamed/moved

devimc:
@jodh-intel 

> Is this the absolute guest root context path? In other words, is this API allowing changes to, say, /sbin/init? Or is it designed purely to copy files into the workloads rootfs directory?

it's designed to copy files into the workloads rootfs directory, I think we can validate that `Path` always  starts with `/tmp`

grpc.go
devimc: sure thing

grpc.go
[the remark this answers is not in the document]
jodh-intel: I just mean that the function could do something like the following (untested):

```go
dir := filepath.Dir(req.Path)
file := filepath.Base(req.Path)

// create hidden temp file in same directory as original file with a random suffix
tmpfile, err := ioutil.TempFile(dir, "."+file+"*")

name := tmpfile.Name()
tmpfile.WriteAt(...)
tmpfile.Close()

// move the temp file over the top of the destination path atomically
err := os.Rename(name, req.Path)
```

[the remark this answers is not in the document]
devimc: @jodh-intel  ohh I see a problem with this design, if offset is `!= 0` and `dest` exist, it must be copied to the temporary file, the problem is when `dest` is a humongous file, copying `dest` to the temporarily file will be expensive in both time and disk space.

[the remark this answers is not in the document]
jodh-intel: yep - slow, but safe. We need to capture these sorts of scenarios in a doc and decide on the best approach in each case I think.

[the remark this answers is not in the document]
egernst: @sboeuf writes:
well you need to think about this CopyFile() command as a workaround for 9p, so we're not planning on having huge files being passed through here. Thus, don't know if we should worry about performance here. I value more the security check personally.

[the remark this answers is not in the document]
devimc: I have changed the algorithm to copy files, please take a look

protocols/grpc/agent.proto
egernst: I'm even okay with the overwrite just happening, if it keeps it simple.  You can do similar things today with docker CLI, afaiu, right?  Either way - adding overwrite option seems reasonable to me.

caoruidong:
How to evaluate a file is small enough? Will this used for only /etc stuffs?

devimc:
@caoruidong , actually I need to update the commit message, since the algorithm to copy file has changed and now there is not limit in the size

egernst:
@bergwolf - we haven't tested with FC's MDS, just vsock so far.  While this may be interesting down the road, I want to enable this no-9p use case for more hypervisors than just FC (NEMU/QEMU, and simplify requirements for additional hypervisors).

bergwolf:
@egernst I see, let's move forward with copyfile then.

grpc.go
```diff
@@ -1449,3 +1449,54 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
...
+		return emptyResp, nil
+	}
+
+	// At this point temoporary file has the expected size, atomically move it overwriting
+	// the destination.
+	agentLog.WithFields(logrus.Fields{
+		"tmp-path": tmpPath,
+		"des-path": req.Path,
+	}).Debugf("Moving temporary file")
+	if err := os.Rename(tmpPath, req.Path); err != nil {
```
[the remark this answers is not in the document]
jodh-intel: Good point @bergwolf! I guess we could use this approach if:

- the destination file does not already exist, or
- the destination file does exist and its size == `len(content) && offset = 0`.

We could just [KISS](https://en.wikipedia.org/wiki/KISS_principle) and go with the basic file copy with no additional checks. If the code ends up littered with special cases, that's going to make it potentially fragile from a security perspective and difficult to maintain but might be essential to limit the scope of the feature.

Maybe simple is the best approach, but I'm not clear if that is the case yet as I don't really understand the full expected scope of this feature.

It's a superficially a simple change but does raise a lot of questions such as:

- What is the precise use-case for this feature?
- When will it be used ("boot phase only", or "at any time" (in other words potentially after the workload has started to run?)

   If it's "boot only", then maybe security is less of an issue if this feature is working in the same environment as the current 9p code (which by definition we are comfortable with using from a security perspective).

   But I think this is implicitly going to be available "post-boot" which goes beyond the scope of 9p. Is that acceptable to us as this is therefore not a direct replacement for what we currently have?
- What does 9p do currently? (Does  `CopyFileRequest` provide the same facilities or more/less?)
- Will we need this when `virtio-fs` is available? If not, can we just wait for `virtio-fs`?
- How will we test this:
  - To prove all the `CopyFileRequest` fields work as expected.
  - For metrics purposes (compared with `9p`, `virtio-fs`).
  - Will we need new security tests? (I would say yes).
- What changes to the runtime will be required to switch from 9p to this approach?

Ultimately, this looks like a useful feature. I just want to ensure we are all aware of the scope of the feature and the implications of introducing it.

/cc @grahamwhaley.

grpc.go
```diff
@@ -1449,3 +1449,54 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
...
+		return emptyResp, err
+	}
+
+	agentLog.WithFields(logrus.Fields{
+		"tmp-file-size": st.Size(),
+		"expected-size": req.FileSize,
+	}).Debugf("Checking temporary file size")
+	// if file size is not equal to the expected size means that copy file operation has not finished.
+	// CopyFile should be called again with new content and a different offset.
+	if st.Size() != req.FileSize {
```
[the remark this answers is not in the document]
devimc: this is not to check `WriteAt`, client sends file's size (expected size), temporary file is renamed to destination file once its size is equal to expected size

grpc.go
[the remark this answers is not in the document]
devimc: > This rename approach is not going to work with the sending files by parts purpose.

why? I wrote a test to check that, please take a look https://github.com/kata-containers/agent/pull/433/files#diff-2815be5959adb8605bff6e18d6f59b6fR790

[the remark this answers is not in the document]
bergwolf: @devimc If you send a file in a sequence of `copyfile` requests, each of them will do one `rename` and the previously written file is lost.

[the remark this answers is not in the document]
devimc: sorry I don't understand :worried:

grpc.go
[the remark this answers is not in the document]
sboeuf: Is this comment properly addressed? Can we resolve it?

sboeuf:
@jodh-intel @bergwolf 
I've read your comments and concerns, but I think we should really keep this method as simple as possible here for now.
First, the need for this function comes from the need to support hypervisor without a 9p or simple filesystem support. So the container rootfs is exposed as a block device, but simple files such as `resolv.conf` still needs to be passed inside the VM through a different way since they're not part of the rootfs.

The call needs to be kept simple as it will be the runtime's responsibility to update the files by monitoring any change. But for now that's a simple way to pass those files.

From a security perspective, if we keep it simple, and one-way, we're not opening any security holes, since we won't provide any mechanism to write back to the host.
By keeping it simple, I mean that we should not try to support copying the same file in several parts, at least for now.

I agree we should make sure we don't allow this call to override some files of our VM rootfs, but I think @devimc took care of this already by writing to `/tmp`.

@devimc 
You should make sure to clearly document in the code the purpose of this function, explaining the need for it and the very limited scope that is targeted here.

grpc.go
```diff
@@ -1449,3 +1449,54 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
...
+		return emptyResp, fmt.Errorf("Only is possible to copy files into the /tmp directory")
+	}
+
+	if err := os.MkdirAll(filepath.Dir(req.Path), os.FileMode(req.DirMode)); err != nil {
+		return emptyResp, err
+	}
+
+	// create a temporary file and write the content.
+	tmpPath := req.Path + ".tmp"
+	tmpFile, err := os.OpenFile(tmpPath, os.O_WRONLY|os.O_CREATE, 0755)
```
jodh-intel: Could we make this `0600` as we set the perms to the requested values later?

grpc.go
```diff
@@ -1449,3 +1449,54 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
...
+	// the destination.
+	agentLog.WithFields(logrus.Fields{
+		"tmp-path": tmpPath,
+		"des-path": req.Path,
+	}).Debugf("Moving temporary file")
+	if err := os.Rename(tmpPath, req.Path); err != nil {
+		return emptyResp, err
+	}
+
+	return emptyResp, os.Chmod(req.Path, os.FileMode(req.FileMode))
```
jodh-intel: It's safer to do the `os.Chmod()` before the `os.Rename()` as that way there is no "window" for the destination file to have higher privs (from the temp file) than the requested privs.

devimc:
@sboeuf  @jodh-intel  changes applied

grpc.go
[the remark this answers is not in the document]
bergwolf: @devimc `rename` would overwirte the entire file. E.g., if you do `write foo` -> `rename foo bar` -> `write foo w/ a different offset` -> `rename foo bar`, then contents created by the very first `write foo` is lost.

grpc.go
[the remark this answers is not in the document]
bergwolf: I don't see why we need `stat()`. In each request, we have `offset` and `len(content)`, and the sum of them should just equal to file size in the request otherwise the request is illegal on its own. If we really want to validate that, we should do it upon handling the request and before really writing to the file. The current implementation of write->stat->check size, is validating the golang fs interfaces instead.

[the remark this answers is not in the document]
bergwolf: Oh, I see your point. Sorry for the noise.

bergwolf:
LGTM! Thanks @devimc !

protocols/grpc/agent.proto
```diff
@@ -434,3 +435,20 @@ message StringUser {
...
+
+message CopyFileRequest {
+	// Path is the destination path in the guest.
+	string path = 1;
+	// FileSize is the expected file size, for security reasons write operations
+	// are made in a temporary file, once it has the expected size, it's moved
+	// to the destination path.
+	int64 file_size = 2;
+	// FileMode is the file mode.
+	uint32 file_mode = 3;
```
jodh-intel: Don't we also need to send `uid`+`gid` for both file and directory as currently they will only ever be created as the user the agent is running as (`root`)? Maybe less important but also what about `atime`, `mtime`, `ctime`?

protocols/grpc/agent.proto
```diff
@@ -434,3 +435,20 @@ message StringUser {
 	string gid = 2;
 	repeated string additionalGids = 3;
 }
+
+message CopyFileRequest {
+	// Path is the destination path in the guest.
```
jodh-intel: Thanks for adding the comment. I think it could be made clearer though that the path must be:

- absolute and canonical.
- below `/run`.

grpc.go
```diff
@@ -1450,3 +1451,68 @@ func (a *agentGRPC) SetGuestDateTime(ctx context.Context, req *pb.SetGuestDateTi
...
+// CopyFile copies files form host to container's rootfs (guest). Files can be copied by parts, for example
+// a file which size is 2MB, can be copied calling CopyFile 2 times, in the first call req.Offset is 0,
+// req.FileSize is 2MB and req.Content contains the first half of the file, in the seconds call req.Offset is 1MB,
+// req.FileSize is 2MB and req.Content contains the second half of the file. For security reason all write operations
+// are made in a temporary file, once temporary file reaches the expected size (req.FileSize), it's moved to
+// destination file (req.Path).
+func (a *agentGRPC) CopyFile(ctx context.Context, req *pb.CopyFileRequest) (*gpb.Empty, error) {
+	// container's rootfs is mounted at /run, in order to avoid overwrite guest's rootfs files, only
+	// is possible to copy files to /run
+	if !strings.HasPrefix(req.Path, containersRootfsPath) {
```
jodh-intel: This isn't strict enough - consider the following, which will be accepted:

```
req.Path = "/run/../sbin/init"
```

But since we can't guarantee the file will already exist, it can't be expanded trivially. How about just adding an extra check and denying the request if it contains "`../`" ?

devimc: good catch!, I think `realpath` can help us here

protocols/grpc/agent.proto
devimc: yes files are created as `root`, I'll add `uid` and `gid`, the other options can be added later on, what do you think?

jodh-intel: sounds good.

devimc:
@jodh-intel changes applied, thanks

protocols/grpc/agent.proto
```diff
@@ -434,3 +435,25 @@ message StringUser {
...
+	// FileSize is the expected file size, for security reasons write operations
+	// are made in a temporary file, once it has the expected size, it's moved
+	// to the destination path.
+	int64 file_size = 2;
+	// FileMode is the file mode.
+	uint32 file_mode = 3;
+	// DirMode is the mode for the parent directories of destination path.
+	uint32 dir_mode = 4;
+	// Uid is the numeric user id.
+	int32 uid = 5;
```
jodh-intel: There is the question of how the client is going to know the `uid` and `gid` values of course. They can prolly safely assume the standard users exist, but if they want to get elaborate, we may need to consider expanding `GetGuestDetails()` to return a list of uids+gids (optionally).... I know! This appears to be such a simple feature, but... :smile:

devimc: yes, I think so :smiley:

jodh-intel:
Given the recent discussion and changes, @sboeuf and @egernst may want to re-ack this PR?

grpc.go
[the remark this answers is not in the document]
devimc: @bergwolf temporary file (in your example `foo`) is renamed once its size is equal to the expected size (host file's size)

Issue #435, opened 2018-12-13
Container factory root folder is not writable on aarch64


marcov:
### Description of problem
Starting a kata container on aarch64 returns this:
```
# docker run -it --rm --runtime kata-runtime alpine
docker: Error response from daemon: OCI runtime create failed: rpc error: code = Unknown desc = mkdir /tmp/libcontainer: read-only file system: unknown.
```
This is using the standard config file and a rootfs image based on suse.

`/tmp/libcontainer` is the root used by libcontainer to store container data. 

I found that this error is triggered on ARM because the root filesystem is mounted in read-only mode (by virtcontainers) and `/tmp` is not mounted as `tmpfs` (it's like this on many distros).

This issue is not present on x86_64, as the file system is mounted as R/W (and despite `/tmp` not being mounted as `tmpfs`, you can still write in it). And using nvdimm  without `share=on` prevents writes to `/tmp` to be written back in the image file.

ARM is not yet using nvdimm, and so mounting filesystem R/W is not an option, so changes to filesystem would be propagated back to the image file.

I'd propose to use `/run/libcontainer` instead of `/tmp/libcontainer` to store container data.
I checked on {SUSE,centos,clearlinux,ubuntu,Fedora} and on all of them `/run` is mounted as `tmpfs`.

Pennyzct:
Hi~ @marcov Supporting nvdimm is on progress. you can see detailed info on issue [runtime#843](https://github.com/kata-containers/runtime/issues/843). It needs to refactoring multiple modules (host kernel, guest kernel, qemu, etc) to be fully working well, so it may need a while😭.

PR #436, opened 2018-12-17
libcontainer: use /run as root containers path

marcov:
Use /run/libcontainer as root path for container factory, as /tmp may
be read only for some configuration / architectures.

Fixes: #435

Signed-off-by: Marco Vedovati <mvedovati@suse.com>


jodh-intel:
lgtm

/test

PR #443, opened 2019-01-18
agent: support memory hotplug via probe interface on aarch64

Pennyzct:
# Description of problem
There are 2 phases in Memory Hotplug. 1) Physical Memory Hotplug phase. 2)  Logical Memory Hotplug phase.
The First phase is to communicate hardware/firmware and make/erase environment for hotplugged memory.
If firmware supports notification of connection of new memory to OS, this phase is triggered automatically. ACPI can notify this event. This also what kata supports, memory hotplug via acpi-driven.
But if not, there is another option, `probe` operation by hand. 
And since memory hotplug via acpi is missing on qemu-system-aarch64, we hope to support the other `probe` solution.
the whole detailed info about `probe` interface could be found [here](https://github.com/spotify/linux/blob/master/Documentation/memory-hotplug.txt#L192).
This whole implementation of memory hotplug via probe interface is divided into two phases, the first is to check if guest kernel is capable of memory hot-add via probe interface, which is located at `/sys/devices/system/memory/probe`.
The second is let kata-agent tell guest kernel the physical address of new memory block.
`echo start_address_of_new_memory > /sys/devices/system/memory/probe`

# Expected result
```
root@entos-thunderx2-desktop:~# docker run --runtime kata-runtime -it --rm -m 3G ubuntu
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
root@9ebdb553ba9b:/# lsmem
RANGE                                 SIZE  STATE REMOVABLE     BLOCK
0x0000000040000000-0x00000000bfffffff   2G online        no       1-2
0x0000020000000000-0x00000200bfffffff   3G online        no 2048-2050

Memory block size:         1G
Total online memory:       5G
Total offline memory:      0B
root@9ebdb553ba9b:/# ls /sys/devices/system/memory/
auto_online_blocks  block_size_bytes  memory1  memory2  memory2048  memory2049  memory2050  power  probe  uevent

```


protocols/grpc/agent.proto
```diff
@@ -348,13 +349,26 @@ message GuestDetailsRequest {
...
 	// MemBlockSizeBytes returns the system memory block size in bytes.
 	uint64 mem_block_size_bytes = 1;
 
 	AgentDetails agent_details = 2;
+
+	bool support_mem_hotplug_probe = 3;
+}
+
+message MemHotplugByProbeRequest {
+	// server needs to send the value of mem_hotplug_probe_addr into file /sys/devices/system/memory/probe,
```
devimc: `MemHotplugProbeAddr ....`

Pennyzct: thanks! updated. ;)

Pennyzct:
oops, I will add another commit to fix mockserver warning, update asap.

jodh-intel:
lgtm

Pennyzct:
Hi~ all~ @devimc @jodh-intel  I added another commit to refine the unit test for func` GetGuestDetails` and fixed the `mockserver` warning, ptal. ;)

jodh-intel:
Thanks @Pennyzct. Still...

lgtm

ping @kata-containers/agent.

sboeuf:
@Pennyzct I'm trying to understand why you need this new method to hotplug some memory, and why the current way of doing (calling into `OnlineMemory()`) does not work for `aarch64`.

Pennyzct:
Hi~ @sboeuf  FWIT, `OnlineMemory` tries to bring memory online, when receiving `udev memory add` events. That's the second phase of memory hotplug. 
What's missing on aarch64 is the first phase, which is building communication between  hardware/firmware and guest OS and then making environment for hotplugged memory. amd64 uses ACPI to notify hot-add event, which isn't applied for `qemu-system-aarch64`.  So instead we choose to use `probe` interface. 
(https://github.com/spotify/linux/blob/master/Documentation/memory-hotplug.txt#L192).
The new method tell guest kernel the physical address of new memory block.
`echo start_address_of_new_memory > /sys/devices/system/memory/probe`
And after echoing right address, we finally could get `udev memory add` event on aarch64. ;)
I hope that i clarified this whole scheme to you clearly, otherwise, the whole detailed info about `probe` interface locates [here].

sboeuf:
@Pennyzct thanks for the explanation, and I clearly get that if `qemu-aarch64` is not relying on ACPI to hotplug memory, then we clearly need another mechanism to notify the kernel about this new memory block. The solution looks good to me, please make sure you provide this same clear explanation as part of your commit message and as part of your PR description. It's very important that we can identify quickly the need for this patch when reading through the git log.

One more thing, reading from the link you provided about probing, you will certainly need to update the Kata Containers kernel config with `CONFIG_ARCH_MEMORY_PROBE`.

sboeuf:
Perfect! Thanks @Pennyzct

jodh-intel:
Ping @kata-containers/agent

devimc:
lgtm

Issue #446, opened 2019-01-28
cpuset fails when there are missing vcpus on guest.


jcvenegas:
Kata agent fails when a container creation or update request set
cpuset for cpus non exting on VM.

# Expected result

Ignore cpuset if is not possible to apply in guest.

# Actual result

This is not a bad behvior but to increase compatibilty, it could try to apply
in case of  possible.

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #447, opened 2019-01-28
cgroup: ignore cpuset if can not be applied.

jcvenegas:
The way that kata works is hotplugging cpus to the VM
based in cpu and qouta, cpuset request may not match
with the cpus requested from the host. Apply
only if possible.
# Cpuset cgroup.

From Kernel documentation:

_" Cpusets provide a mechanism for assigning a set of CPUs and Memory Nodes to
a set of tasks."_

The Kata agent brings compatibility the cgroup cpuset CPU on the guest side.

The cpuset CPU cgroup will be applied on two events:

- containers creation.

- container update.

When the runtime request to apply cpuset cgroup to the agent, the amout of
vcpus available may be not the same to the required in the request.

This is becasue the request from the agent client (the Kata runtime) pass
cpusets that were requested to be placed on the host. This is to isolate the
container workload on some specific host CPUs. The the runtime pass the
requested cpuset to the agent, that will try to apply the cgroup cpuset on the
guest.

Because the runtime will only hot-plug and amount the CPUs calculated based in
the container period and quota, the VM wont have the same amount CPUs than the host.
guest.

Example:

docker run -ti --cpus 2 --cpuset 0,1 busybox
 
The expected result is that the container is limited to the time of 2 CPUs, but
is only allowed to be scheduled on CPUs 0 and 1 

In a similar case and a valid traditional container is the following:

docker run -ti --cpus 2 --cpuset 2,3,4 busybox

Where the container is limited to 2 CPUs and can be scheduled on CPU 2,3,4.

In the case of the Kata. The Kata runtime will only hotplug 2 cpus, making
imposible to request to the guest kernel to schedule the workload on vcpu
3,4.

## cpuset best effort application.

Kata will try agent will evaluate this request and see if is possible to 
apply the cpuset request on the guest.

- If the cpuset request includes CPUs that are not available  in the guest it
  will ignore the request.
- If the request has the CPUs requested in are available  the request be
  applied in the guest.

Fixes: #446


cgroup.go
```diff
@@ -0,0 +1,50 @@
...
+func getAvailableCpusetList(cpusetReq string) (string, error) {
+
+	cpusetGuest := getCpusetGuest()
+
+	cpusetListReq, err := parsers.ParseUintList(cpusetReq)
+	if err != nil {
+		return "", err
+	}
+
+	cpusetGuestList, err := parsers.ParseUintList(strings.TrimSpace(string(cpusetGuest)))
```
devimc: cpusetGuest is a string, why is `string()` needed?

cgroup.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	for k := range cpusetListReq {
+		if !cpusetGuestList[k] {
+			agentLog.Warnf("cpu %d is not in guest cpuset, using %s cpuset", k, cpusetGuest)
+			return cpusetGuest, nil
```
devimc: `getAvailableCpusetList` returns all the cpus when a cpu in the list of cpuset is not available in the guest, for example if cpuset is 7, but the available cpus in the guest are 0-6, this function will return 0-6 even when only one cpu is requested. I wonder if this implementation is correct

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+- containers creation.
+
+- container update.
+
+When the runtime request to apply cpuset cgroup to the agent, the amout of
+vcpus available may be not the same to the required in the request.
+
+This is becasue the request from the agent client (the Kata runtime) pass
+cpusets that were requested to be placed on the host. This is to isolate the
+container workload on some specific host CPUs. The the runtime pass the
```
devimc: `The the runtime`

cgroup.go
jcvenegas: Good catch, I did not removed when factored out `getCpusetGuest()`

cgroup.go
jcvenegas: For me it is difficult to decide if correct or not. Because it is traying to make more compatible a broken use case in Kata that is a current limitation because we run the workload in VM with a different set of CPUs 

I prefer to not apply the cpuset at all instead of apply only where just a subset of them, why? 
- the workload is still limited to use limited cpu time based on cpu cgroup (cpu and qouta) so still be safe not consuming cpu time of other process. 

> for example if cpuset is 7, but the available cpus in the guest are 0-6, this function will return 0-6 even when only one cpu is requested

Lets translate to a docker example.
```
docker run -ti --runtime kata-runtime  --cpuset-cpus 7 --cpus 6  --net none --rm busybox
/ # cat /proc/self/status  | grep Cpus_allowed_list
Cpus_allowed_list:      0-5
```
1. On the host the VM will run only on this cpu, so the workload as well with  https://github.com/kata-containers/runtime/pull/1189
2. On the guest workload, the cpuset will allow to run the workload in all the vcpus (think cpuset on the host is still isolating our workload to jump to other process. 
3. From user perspective on the host this is still correct because the VM is only using only the CPU 7

Another  case:

```
docker run -tid --cpuset-cpu 1,5-9  --cpus 4 mycontainer
```
1. The  cpuset list is valid on the host and thanks to https://github.com/kata-containers/runtime/pull/1189 this is going to be limited on the host and the VM is allow to have its 4cpu time on any of the 1,5-9 host cpus (we isolate the workload)
2. In the guest we will have 4 cpus, and cpuset list of 1,5-9, where only cpu 1 is  available.

There was a few options. 
- Create a mapping for each cpu and vcpu, in this case is not even possible because host has at least 10 cpus and the guest hast 4. 
- Apply only the ones that are available, and ignore the rest, in this case the only cpu available is `1` so the workload would be able to run only on vcpu 1 but there rest of them wont be used by it, this mean waste of vcpus and the workload limited from the cpuset we partially apply to not get the cpu time of 4 cpus.
- If the request has cpus that are not part of it, just ignore it. The cpuset is not applied on the guest and workload will be jumping around the guest vcpus but the host cpuset will prevent the VM going to not allowed host cpus, and the workload will  be still limited to use 4 cpus because its cpu cgroup ( period and quota). 

So I went for the last option where it if is possible the cpuset is applied. But if not there is still isolation at host level. The solution is not perfect but solve a very basic docker use case and works with cpu cgroup and cpuset host cgroup  to work in a more compatible way.
```
docker run -tid --cpuset-cpu 2-4  --cpus 2 mycontainer
```
And should be the same for kubernetes when it isolate workloads to use a set of cpus. I still need to check this use case.

egernst: The cpuset itself needs to be translated before passing into the guest, imo.  As it is provided, it is only valid for physical CPUs.  Once cpuset is done on the host (has this landed already?), then this PR becomes relevant.  IMO, after the runtime pins on the host, the cpuset information needs to be translated to map from physical CPUs to virtual CPUs in the guest.  We shouldn't just pass the original request in.

jcvenegas: @egernst agree. I work on follow up PRs to do this, this is only allowing to run Kata on K8s static policy and docker.

documentation/features/cpuset.md
jcvenegas: done

jcvenegas:
Updated PR to fix issue found testing commit with Kuberentes (multi containers in a pod). This adds fix the agent when try to update cgroup parents with with the request. The parents now are updated with the max set of cpus in the guest.

grahamwhaley:
Just going to try and tickle pullapprove on this, as we are testing it... and I need something not on the proxy repo. Going temporarily push a DNM label on it..

jodh-intel:
:fish:

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
+# Cpuset cgroup.
+
+From Kernel documentation:
+
+_" Cpusets provide a mechanism for assigning a set of CPUs and Memory Nodes to
+a set of tasks."_
+
+The Kata agent brings compatibility the cgroup cpuset CPU on the guest side.
```
klynnrif: Suggest adding a preposition: The Kata agent brings compatibility to the cgroup...

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+
+The cpuset CPU cgroup will be applied on two events:
+
+- containers creation.
```
klynnrif: Suggest removing the periods after the bulleted items because they are not complete sentences

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+
+- container update.
+
+When the runtime request to apply cpuset cgroup to the agent, the amout of
```
klynnrif: Lines 16-17 suggested rewrite: 
When the runtime requests to apply cpuset cgroup to the agent, the amount of vcpus available might not be the same to the required vcpus in the request.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+vcpus available may be not the same to the required in the request.
+
+This is becasue the request from the agent client (the Kata runtime) pass
```
klynnrif: Lines 19-23 suggested rewrite: 
This is because the request from the agent client (i.e. the Kata runtime) passes cpusets that are requested to be placed on the host. This isolates the container workload on some specific host CPUs. The runtime passes the requested cpuset to the agent, which tries to apply the cgroup cpuset on the guest.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+cpusets that were requested to be placed on the host. This is to isolate the
+container workload on some specific host CPUs. The runtime pass the
+requested cpuset to the agent, that will try to apply the cgroup cpuset on the
+guest.
+
+Because the runtime will only hot-plug and amount the CPUs calculated based in
```
klynnrif: Lines 25-27 suggested rewrite (not sure if you want "host" or "guest" at the end of the sentence): 
The runtime only hot-plugs and calculates the CPUSs based on the container period and quota. This is why the VM will not have the same amount of CPUs as the host.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+the container period and quota, the VM wont have the same amount CPUs than the host.
+guest.
+
+Example:
+
+docker run -ti --cpus 2 --cpuset 0,1 busybox
+ 
+The expected result is that the container is limited to the time of 2 CPUs, but
```
klynnrif: Lines 33-34 suggested rewrite: 
This should result with the container limited to the time of 2 CPUs, but is only allowed to be scheduled on CPUs 0 and 1.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+is only allowed to be scheduled on CPUs 0 and 1 
+
+In a similar case and a valid traditional container is the following:
```
klynnrif: Suggested rewrite: 
The following is an example of a similar case with a valid traditional container:

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+
+docker run -ti --cpus 2 --cpuset 2,3,4 busybox
+
+Where the container is limited to 2 CPUs and can be scheduled on CPU 2,3,4.
```
klynnrif: Suggested rewrite: 
Here, the container is limited to 2 CPUs and can be scheduled on CPU 2, 3, and 4.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+
+In the case of the Kata. The Kata runtime will only hotplug 2 cpus, making
```
klynnrif: Lines 42-44 suggested rewrite: 
The Kata runtime only hotplugs 2 CPUs, making it impossible to request that the guest kernel schedules the workload on vcpu 3 and 4.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+imposible to request to the guest kernel to schedule the workload on vcpu
+3,4.
+
+## cpuset best effort application.
+
+Kata will try agent will evaluate this request and see if is possible to
```
klynnrif: Lines 48-49 suggested rewrite (unsure if I've changed the intended meaning here): 
The Kata agent evaluates the request to see if it is possible to apply the cpuset request onto the guest.

documentation/features/cpuset.md
```diff
@@ -0,0 +1,55 @@
...
+In the case of the Kata. The Kata runtime will only hotplug 2 cpus, making
+imposible to request to the guest kernel to schedule the workload on vcpu
+3,4.
+
+## cpuset best effort application.
+
+Kata will try agent will evaluate this request and see if is possible to 
+apply the cpuset request on the guest.
+
+- If the cpuset request includes CPUs that are not available  in the guest it
```
klynnrif: Lines 51-54 suggested rewrite: 
- If the CPUSs requested are not available in the guest, the request is ignored. 
- If the CPUs requested are available, the request is applied to the agent.

jcvenegas:
@klynnrif thank you! now looks better and easier to understand  :)

cgroup.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	cpusetGuestList, err := parsers.ParseUintList(cpusetGuest)
+	if err != nil {
+		return "", err
+	}
+
+	for k := range cpusetListReq {
+		if !cpusetGuestList[k] {
+			agentLog.Warnf("cpuset '%s': cpu %d is not in guest cpu list, using guest cpus %s", cpusetReq, k, cpusetGuest)
```
devimc: `WithFields`..  please :smile:

cgroup.go
```diff
@@ -0,0 +1,58 @@
...
+			return cpusetGuest, nil
+		}
+	}
+
+	// All the cpus are valid keep the same cpuset string
+	agentLog.Debugf("the requested cpuset '%s' is valid using it", cpusetReq)
```
devimc: `WithFields`

jcvenegas: done, running again CI

documentation/features/cpuset.md
```diff
@@ -5,51 +5,48 @@ From Kernel documentation:
...
-3,4.
+The Kata runtime only hotplugs 2 CPUs, making it impossible to request that the
+guest kernel schedules the workload on vcpu 3 and 4.
 
 ## cpuset best effort application.
 
-Kata will try agent will evaluate this request and see if is possible to 
-apply the cpuset request on the guest.
+The Kata agent evaluates the request to see if it is possible to apply the
+cpuset request onto the guest.
```
sboeuf: Can't you fix this directly on the first commit introducing this new documentation?

cgroup_test.go
```diff
@@ -31,8 +31,8 @@ func TestGetAvailableCpusetList(t *testing.T) {
 		{"0-3", "0-3"},
 		{"0-3,4", fakeGuestCpuset},
 		{"0-4", fakeGuestCpuset},
-		{"1", fakeGuestCpuset},
-		{"1,3", "2,13"},
+		{"1", "1"},
+		{"1,3", "1,3"},
```
sboeuf: Can't you fix this directly on the first commit introducing those tests?

grpc.go
```diff
@@ -179,7 +179,7 @@ func updateContainerCpuset(cgroupPath string, newCpuset string, cookies cookie)
 		// Don't use c.container.Set because of it will modify container's config.
 		// c.container.Set MUST BE used only on update.
 		cpusetCpusPath := filepath.Join(cpusetPath, "cpuset.cpus")
-		agentLog.WithField("path", cpusetPath).Debug("updating cpuset cgroup")
```
sboeuf: why did you remove this log?

jcvenegas: Probably miss to re-add when I move the code, I added  back

cgroup_test.go
jcvenegas: I reorganized the commits

Issue #453, opened 2019-02-06
client: Dont try connect to non existing socket.


jcvenegas:
# Description of problem

When a new Client is created, it tries to connect to
a socket no matter if exist or not. This is not
and issue because the grpc api will time out after
some time. But some simple task like kata-runtime list can
take a lot of the socket does not exist because the
VM and proxy top working.

# Expected result

If a Client tries to connect to a socket return and error 
immediately.

# Actual result

The new client take a long time before fail.

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #454, opened 2019-02-06
client: Dont try connect to non existing socket.

jcvenegas:
When a new Client is created, it tries to connect to
a socket no matter if exist or not. This is not
and issue because the grpc api will time out after
some time. But some simple task like kata-runtime list can
take a lot of the socket does not exist because the
VM and proxy top working.

Fixes: #453

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -82,6 +83,7 @@ func NewAgentClient(ctx context.Context, sock string, enableYamux bool) (*AgentC
 	if err != nil {
 		return nil, err
 	}
+
 	dialOpts := []grpc.DialOption{grpc.WithInsecure(), grpc.WithBlock()}
 	dialOpts = append(dialOpts, grpc.WithDialer(agentDialer(parsedAddr, enableYamux)))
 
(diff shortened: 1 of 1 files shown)


sboeuf:
ping @jcvenegas @bergwolf @mcastelino

jcvenegas:
@sboeuf  so what about add and timeout option and from proxy client increase to wait for the VM socket?

jodh-intel:
Ping @jcvenegas, @sboeuf.

grahamwhaley:
Marking this as DNM until @jcvenegas @sboeuf re-visit and update it.

bergwolf:
@sboeuf Yes, I share the same concern. Since we spawn the proxy process and try to connect to it without any synchronization to make sure the proxy is ready, such race can happen IMOH.

jodh-intel:
@sboeuf, @bergwolf - It might be just me, but reading through the comments, it's not clear whether you are saying this PR is helping or not?

jcvenegas:
@jodh-intel I talk with Sebastien offline about it, I will extend the function cal arguments l to provide a timeout, so runtime can use timeout 0 call and the proxy add  a timeout just  for the initial proxy setup.

grahamwhaley:
@jcvenegas @sboeuf - any update on this - did you get to discuss and come to a conclusion:
- can we fix this with timeouts
- should we close this PR?

sboeuf:
@jcvenegas did you get a chance to update the PR?

PR #457, opened 2019-02-07
network: Remove grpc AddInterface and RemoveInterface api

### LINKED ISSUE #456: Remove unused grpc network related APIs
We only call the UpdateInterface api from the runtime. While looking at the network code, I realised that the others are not really being called/used.

amshinde:
Taking a look at the network code, I saw that the AddInterface and RemoveInterface APIs are not really used/called by the runtime.
Removing unused code.

Fixes #456


sboeuf:
@amshinde 

Since I assume `kata-runtime` is the only consumer of this API, and since this API has never been used, this won't break anything, even if we're talking about upgrades.
But I'm also wondering if we should allow such thing right now. This should be on a list of deprecated functions that we will remove when releasing the next major (hence 2.0).
@egernst @jodh-intel wdyt?

amshinde:
I need to fix this PR by removing unused mock code as well.

> But I'm also wondering if we should allow such thing right now. This should be on a list of deprecated functions that we will remove when releasing the next major (hence 2.0).

I dont really think we need to wait for 2.0 for this, as this change really does not break anything. The functions although present in the API, they are not used anywhere.

jodh-intel:
@jon may have input on this (see https://github.com/kata-containers/documentation/pull/196).

jcvenegas:
@amshinde ping any update on this.?  I like the PR but, still see some jobs failing.

amshinde:
I see an unrelated java.io.EOFException for the metrics job, I'll restart that. Similarly, I see unrelated failure for the firecracker one. Is the fircracker CI supposed to pass?
I'll restart both of them nonetheless.

GabyCT:
@amshinde no the Firecracker CI is being affected by the kernel issue https://github.com/kata-containers/runtime/issues/1203 so I will disable it

amshinde:
thanks @GabyCT

GabyCT:
@amshinde now the CI of firecracker is working as I change it from using Ubuntu to Centos (which is using a kernel that is not broken with vsocks)

grahamwhaley:
ooi @GabyCT - is that because the Centos kernel is old (did not get the vsock breaking patch), or is very new and updated (has the patch of a patch that fixes the broken vsocks again) ?
We should keep track of which distro has which kernel version and is broken/fixed...

GabyCT:
@grahamwhaley the kernel is very old

sboeuf:
@amshinde @GabyCT @grahamwhaley what's the status on this? Is it okay to move forward and merge this PR without having FC CI passing?

@amshinde As you mentioned, this won't break anything, so let's move forward with it.

GabyCT:
@sboeuf , I enable the FC with centos and it is working if we want the whole CI running this PR needs to be pushed again

sboeuf:
@GabyCT I could manually retrigger the FC job, right? Any reason that would prevent me from doing this?

GabyCT:
@sboeuf , I do not think that is able just to re-trigger the job as it does not find `jenkins-ci-ubuntu...`

sboeuf:
Oh sorry I missed the fact that you recently moved FC testing from ubuntu to centos because of the vsock regression in recent ubuntu kernels...
Yes a repush should do it.

@amshinde could you please repush so that it will trigger proper CI?

grahamwhaley:
/test
yeah @sboeuf something 'moved' on the Jenkins side, so the link to the old job no longer exists. If one really needed to then one could probably hand-craft a job, but even then it might not update the fc item above, as the name has now probably changed as well. A whole /test, like above, should re-run all current CIs...

grahamwhaley:
just ignore the old dead ubuntu fc one, and watch the new shiny centos fc one ;-)

sboeuf:
cool! let's see how this goes and I'll merge if all green :)

sboeuf:
@jodh-intel @grahamwhaley PTAL

grahamwhaley:
oh, I'm probably not the person to review grpc proto changes... let's try @gnawux @sameo

PR #459, opened 2019-02-08
wip: network: Wait till link is operationally up.

### LINKED ISSUE #458: expect ip a to show link in up state
Refer to : https://github.com/kata-containers/runtime/issues/1201

amshinde:
Although we set the link up with LinkSetup, this function
returns before the link is actually up.
Additionally, besides the admin state of a network link, we
also need to check if the link is operationally up.
Namely we need to check if the netlink attribute IFLA_OPERSTATE is set
to IF_OPER_UP.

Fixes #458

### DIFF
--- a/network.go
+++ b/network.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"reflect"
 	"sync"
+	"time"
 
 	"golang.org/x/sys/unix"
 
@@ -230,6 +231,49 @@ func (s *sandbox) removeInterface(netHandle *netlink.Handle, iface *types.Interf
 	return nil, nil
 }
 
+const linkUpdateTimeout = 5
+
+var linkOperational = true
+
+func expectLinkUpdate(ch <-chan netlink.LinkUpdate, ifaceName string) bool {
+	for {
+		timeout := time.After(time.Second * linkUpdateTimeout)
+		select {
+		case update := <-ch:
+			// Reference: https://www.kernel.org/doc/Documentation/networking/operstates.txt
+
+			if ifaceName == update.Link.Attrs().Name && (update.IfInfomsg.Flags&unix.IFF_UP != 0) {
+				if linkOperational == (update.Link.Attrs().OperState == netlink.OperUp) {
+					return true
+				}
+			}
+		case <-timeout:
+			return false
+		}
+	}
+}
+
+func waitForLinkUp(netHandle *netlink.Handle, link netlink.Link, ifaceName string) (err error) {
+	ch := make(chan netlink.LinkUpdate)
+	done := make(chan struct{})
+	defer close(done)
+
+	if err = netlink.LinkSubscribe(ch, done); err != nil {
+		return
+	}
+
+	//put the link back into the up state
+	if err = netHandle.LinkSetUp(link); err != nil {
+		return
+	}
+
+	if !expectLinkUpdate(ch, ifaceName) {
+		err = fmt.Errorf("Link Up update not received as expected for interface %s", ifaceName)
+	}
+
+	return err
+}
+
 // updateInterface will update an existing interface with the values provided in the types.Interface.  It will identify the
 // existing interface via MAC address and will return the state of the interface once the function completes as well an any
 // errors observed.
@@ -286,8 +330,8 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *types.Interf
 		} else {
 			resultingIfc = iface
 		}
-		//put the link back into the up state
-		retErr := netHandle.LinkSetUp(link)
+
+		retErr := waitForLinkUp(netHandle, link, iface.Name)
 
 		//if we failed to setup the link but already are returning
 		//with an error, return the original error
--- a/network_test.go
+++ b/network_test.go
@@ -22,6 +22,11 @@ import (
 func TestUpdateRemoveInterface(t *testing.T) {
 	skipUnlessRoot(t)
 
+	linkOperational = false
+	defer func() {
+		linkOperational = true
+	}()
+
 	s := sandbox{}
 
 	ifc := types.Interface{


network.go
```diff
@@ -230,6 +231,21 @@ func (s *sandbox) removeInterface(netHandle *netlink.Handle, iface *types.Interf
 	return nil, nil
 }
 
+func expectLinkUpdate(ch <-chan netlink.LinkUpdate, ifaceName string) bool {
+	for {
+		timeout := time.After(time.Second * 5)
+		select {
+		case update := <-ch:
+			// Reference: https://www.kernel.org/doc/Documentation/networking/operstates.txt
+			if ifaceName == update.Link.Attrs().Name && (update.IfInfomsg.Flags&unix.IFF_UP != 0) && update.Link.Attrs().OperState == netlink.OperUp {
```
jodh-intel: Could you reformat this line as it makes my eyes ache? :smile:

amshinde: haha :) I'll try to see how I can simplify it.

amshinde:
@mcastelino @egernst Before I go fix the unit tests, I would like your opinion on this PR.
So this fixes the issue that we were seeing with the interface being down when you execute 
`docker run --runtime=kata-runtime debian ip a`

Adding trace logs, I saw that the netlink call to LinkSetup() does return true, but looks like this call does not really wait till the link is up. Also I realised that for the link to show as "up" in the output of `ip a` command, not only does `IFF_UP` be set in the link flags, but the also the link's `IFLA_OPERSTATE` state needs to be set as up.

We were actually running into the same scenario when running an actual terminal as well, its just that the delay in starting a shell and we actually typing command to show network interfaces masked the problem.

I have fixed the issue with adding some kind of mechanism to watch for the link state, so that we wait till the link is fully up(admin up and operational up). While just adding sleeps initially to guage the time it takes for the link to me up, I realised that it easily takes more that 500ms on my laptop for the link to be up. We would need to measure this, but yes this commit adds additional delays to our container start time.
What do you think?

network.go
```diff
@@ -230,6 +231,21 @@ func (s *sandbox) removeInterface(netHandle *netlink.Handle, iface *types.Interf
 	return nil, nil
 }
 
+func expectLinkUpdate(ch <-chan netlink.LinkUpdate, ifaceName string) bool {
+	for {
+		timeout := time.After(time.Second * 5)
```
sboeuf: You could define this outside of the `for` loop. Also, assign a constant for the 5 seconds, that would be something like `LinkUpdateTimeout` for instance.
```go
...
const LinkUpdateTimeout = 5
...
func expectLinkUpdate(ch <-chan netlink.LinkUpdate, ifaceName string) bool {
        timeout := time.After(time.Second * LinkUpdateTimeout)
        for {
....
```

network.go
```diff
@@ -286,6 +302,15 @@ func (s *sandbox) updateInterface(netHandle *netlink.Handle, iface *types.Interf
 		} else {
 			resultingIfc = iface
 		}
+
+		ch := make(chan netlink.LinkUpdate)
+		done := make(chan struct{})
+		defer close(done)
+
+		if err := netlink.LinkSubscribe(ch, done); err != nil {
```
sboeuf: This is making the defer function too complex to read IMO, instead I would create a function called `setupLink()` that would internally take care of:
```go
LinkSubscribe()
LinkSetUp()
expectLinkUpdate()

**Comment by egernst** [comment]
@amshinde - can you run the metrics tests on your branch to get a measurement on boot time impact?

**Comment by amshinde** [comment]
@egernst With just using time :)
With my changes :
```
$  time sudo docker run  --runtime kata-runtime  debian ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq state UP group default qlen 1000
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.2/16 brd 172.17.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe11:2/64 scope link tentative 
       valid_lft forever preferred_lft forever

real    0m4.036s
user    0m0.036s
sys     0m0.038s
```
Without my changes :
```
$ time sudo docker run  --runtime kata-runtime  debian ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq state DOWN group default qlen 1000
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.2/16 brd 172.17.255.255 scope global eth0
       valid_lft forever preferred_lft forever

real    0m3.444s
user    0m0.055s
sys     0m0.023s
```

sboeuf:
wow that looks bad... half a second for the interface to be up and ready to be used. But I guess that's the way to go anyway since we want the container to be fully functional when the user can start interacting with it.

egernst:
good lord, what system are you using?  3.5 seconds for the "good" result? :)

mcastelino:
> @egernst With just using time :)
> With my changes :

@amshinde which network interworking mode are you using. tcfilter or macvtap

amshinde:
> @amshinde which network interworking mode are you using. tcfilter or macvtap

@mcastelino I was using macvtap, but see similar behaviour with tcfilter as well.

grahamwhaley:
@amshinde did you take the average over multiple runs? A single run can have a 25% difference, and the 'first run' may have significant difference due to warming things up etc.
I'd strongly recommend using the [report generator](https://github.com/kata-containers/tests/tree/master/metrics/report), which will take 20 measurements, and show you some graphs that can help assess variance as well as compare the two runs. You can tell the `grabdata.sh` to only do a timing test by using its `-t` option, which will reduce the time to run the test a lot, and the report generation is all wrapped into a dockerfile so you don't have to install any extra tooling to process the results.

grahamwhaley:
/me wonders if we can tell how long it took to get the network up... if so, we could add that to the report/tests...

grahamwhaley:
Oh, and @amshinde - you might want to make sure you have a good entropy source as well - I just found one of my tests was pretty 'spiky', with a pretty often 1s leap in time to workload, and I think that was running out/waiting for entropy. Adding `haveged` to my system I think has cleared that. Watch out for that one ;-) (and, I think that is maybe why I have seen a number of spiky results graphs from users in the past... tricky that one.. maybe I can get the metrics to check and wait/timeout if entropy is not present...)

sboeuf:
@amshinde have you been able to address the comments? Did you manage to get better results with proper entropy?

sboeuf:
Just added `do-not-merge` label until we get better numbers, or until we decide those numbers are acceptable.

jodh-intel:
Ping @amshinde.

amshinde:
@jodh-intel @egernst  I did try doing a network access immediately after the container is up, it does not error out, just waits for the nework to be ready first. 
So inclined to close this.

Issue #460, opened 2019-02-20
CI: add proto review group into CODEOWNERS file


grahamwhaley:
During the move from pullapprove, I think we lost the special group of folks who need to review any grpc proto changes. Let's add them back in by making the appropriate CODEOWNERS file update.

grahamwhaley:
for ref, this is what the old pullapprove config looked like:

```yaml
protocol-changes:
    required: 2
    teams:
      - architecture-committee
      - builder
      - packaging
    conditions:
      files:
        include:
          - "*.proto"
        exclude:
          - "vendor/*"
```

PR #461, opened 2019-02-20
CODEOWNERS: add group review trigger for proto file changes

grahamwhaley:
If a proto file is changed then we want a larger number of folks
to know about and review it. Add them to the CODEOWNERS file.

Fixes: #460

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -11,3 +11,8 @@
 
 *.md    @kata-containers/documentation
 
+# All protocol changes need to get some review from these groups.
+# Note, we include all subdirs, including the vendor dir, as at present there are no .proto files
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@sboeuf @jodh-intel ptal
Could you also tell me what you think about the 'ordering rules' wrt this file - and what would happen if somebody changed both a .md and a .proto file maybe? See [the CODEOWNERS doc](https://help.github.com/articles/about-code-owners/) for instance.

jodh-intel:
Let's just merge this - it's a config change that doesn't affect the tests so the centos failure is unrelated...

PR #475, opened 2019-03-12
agent: increase the number of tries to online vcpus

### LINKED ISSUE #474: tests: cpu hotplug test fail randomly
seems like cpu hotplug tests are failing because 1 second is not enough to online vcpus

devimc:
cpu hotplug tests fail randomly in nested environments because 1 second is
not enough for the kernel and qemu to allocate and connect the vcpus.

fixes #474

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -68,7 +68,7 @@ var emptyResp = &gpb.Empty{}
 
 const onlineCPUMemWaitTime = 100 * time.Millisecond
 
-const onlineCPUMaxTries = 10
+var onlineCPUMaxTries = uint32(100)
 
 const cpusetMode = 0644
 
--- a/grpc_test.go
+++ b/grpc_test.go
@@ -227,6 +227,12 @@ func TestOnlineCPUMem(t *testing.T) {
 	sysfsCPUOnlinePath = "/xyz/123/rgb/abc"
 	sysfsMemOnlinePath = "/xyz/123/rgb/abc"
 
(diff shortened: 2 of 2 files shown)


grpc.go
```diff
@@ -68,7 +68,7 @@ var emptyResp = &gpb.Empty{}
 
 const onlineCPUMemWaitTime = 100 * time.Millisecond
 
-const onlineCPUMaxTries = 10
+const onlineCPUMaxTries = 30
```
sboeuf: Why not extend it to some round and large value? Something like `100` :)

grahamwhaley: I probably agree - unless we have some hard deadline we know we should meet, then presumably we are looking for vcpus 'never coming online', in which case a large timeout costs us nothing apart from in the interesting rare failure case that we are looking for.
To note, I have seen 'slow' machines take >3s to get to the workload - now I don't know where that time was spent (or if it was related to entropy or not, but I think not)... so I like the idea of `100` more than `30` :-)

devimc: 10 seconds? :thinking:  ...  sounds good

devimc:
@sboeuf @grahamwhaley  change applied, thanks

amshinde:
@devimc CI is failing. ptal.

devimc:
@amshinde fixed, thanks

jodh-intel:
Thanks @devimc!

Issue #471, opened 2019-03-05
cpu hotplug support on arm64


hejianet:
# Description of problem

see https://github.com/kata-containers/runtime/issues/1262

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

justin-he:
ping  @jodh-intel @grahamwhaley

PR #478, opened 2019-03-14
agent: add offline cpu interface support

justin-he:
There is no acpi-driven cpu hotplug on arm till now. But we can provide
a cpu pool implementation instead.
```
+---------------+               +----------------+         +-----------------+
|   kata qmp    |               |     qemu       |         |     guest       |
+-------+-------+               +----------------+         +--------+--------+
        |                      start guest with -smp $nbr           |
        |                       -append "maxcpus=1"                 |
        |                                |                          |
        |                                +------------------------->|
        |                                |                    start cpu with
        |                                |            1 online, others offline
        |                                |                          |
        |  query_hotpluggable_cpus       |                          |
        +-------------------------------->                          |
        |                                |                          |
        | return cpu node info           |                          |
        +<-------------------------------+                          |
        |                                |                          |
        |      device_add/del cpu        |                          |
        +------------------------------->+                          |
        |                                |                          |
        |                 online/offline cpus by gRPC               |
        +---------------------------------------------------------->+
        |                                |                          |
        +                                +                          +
```

online cpu interface has been already available, but offline is not.

Fixes: #471


grpc.go
```diff
@@ -123,6 +167,27 @@ func onlineResources(resource onlineResource, nbResources int32) (uint32, error)
...
 }
 
+func offlineCPUResources(nbCpus uint32) error {
+	resource := onlineResource{
+		sysfsOnlinePath: sysfsCPUOnlinePath,
+		regexpPattern:   cpuRegexpPattern,
+	}
+
+	var count uint32
+	for i := uint32(0); i < onlineCPUMaxTries; i++ {
```
jodh-intel: This is slightly confusing - the function is *offlining* resources but you are iterating using an *online* variable. It might be worth renaming `onlineCPUMaxTries` to something like `cpuStateChangeMaxTries` to make it clearer?

grpc.go
```diff
@@ -123,6 +167,27 @@ func onlineResources(resource onlineResource, nbResources int32) (uint32, error)
...
+		r, err := offlineResources(resource, int32(nbCpus-count))
+		if err != nil {
+			return err
+		}
+		count += r
+		if count == nbCpus {
+			return nil
+		}
+		time.Sleep(onlineCPUMemWaitTime)
```
jodh-intel: Again, the name is slightly confusing to me. But also, doesn't this assume that offlining takes the same amount of time as onlining?

grpc.go
```diff
@@ -81,9 +81,53 @@ func handleError(wait bool, err error) error {
...
+func offlineResources(resource onlineResource, nbResources int32) (uint32, error) {
+	// Get the bitmap of the online cpus
+	onlineCpuBitmapPath := filepath.Join(resource.sysfsOnlinePath, "online")
+	status, err := ioutil.ReadFile(onlineCpuBitmapPath)
+	if err != nil {
+		return 0, err
+	}
+
+	// Get from the last online cpu
+	strArr := strings.Split(strings.Replace(string(status), ",", "-", -1), "-")
```
jodh-intel: These two lines are a bit "dense" - could you split them up for more clarity please? Also, are these guaranteed not to fail if a particular field/substring is not found?

grpc.go
```diff
@@ -81,9 +81,53 @@ func handleError(wait bool, err error) error {
...
+	strLastOnlineCpu := strings.Trim((strArr[len(strArr)-1]), "\n\t ")
+	lastOnlineCpu, err := strconv.Atoi(strLastOnlineCpu)
+	if err != nil {
+		return 0, err
+	}
+	if lastOnlineCpu == 0 {
+		agentLog.Warnf("Can not offline cpu 0")
```
jodh-intel: - I think this would be clearer as:
  ```go
  agentLog.Warnf("offlining last available CPU disallowed")
  ```

- Paranoia: handle `lastOnlineCpu < 0` case?

  This is partially handled by the `for` loop below, but if `lastOnlineCpu == -1`, the function will return `(-1, nil)` rather than an error.

grpc.go
```diff
@@ -1465,6 +1590,14 @@ func (a *agentGRPC) OnlineCPUMem(ctx context.Context, req *pb.OnlineCPUMemReques
 	return emptyResp, a.onlineCPUMem(req)
 }
 
+func (a *agentGRPC) OfflineCPUMem(ctx context.Context, req *pb.OfflineCPUMemRequest) (*gpb.Empty, error) {
```
jodh-intel: Please could you add a test to `grpc_test.go` similar to `TestOnlineCPUMem()`.

grpc.go
justin-he: ok

justin-he:
> Hi @justin-he - thanks for raising!
> 
> A few comments added. Also, the Travis CI is failing on some quick-to-fix static analysis issues:
> 
> ```
> grpc.go:89:2:warning: var onlineCpuBitmapPath should be onlineCPUBitmapPath (golint)
> grpc.go:97:2:warning: var strLastOnlineCpu should be strLastOnlineCPU (golint)
> grpc.go:98:2:warning: var lastOnlineCpu should be lastOnlineCPU (golint)
> ```

ok, I will fix it in newer version

grpc.go
justin-he: Yes, I assume that.
But I can replace it with changeCPUMemStateWaitTime if you have concern

grpc.go
justin-he: My original idea is to get the last online cpu. There might be several cpu list cases:
1. 0-8
2. 0,3,4-8
3. 0,2,4,6,8

step one is to replace all "," with "-". e.g. "0,3,4-8" -> "0-3-4-8"
step two is to get the last field which is splitted by "8", that is "8"

IIUC, this parsing process should not fail if the status contains the correct online cpu list.

grpc.go
justin-he: ok

grpc.go
justin-he: ok

grpc.go
jodh-intel: >  "0,3,4-8" -> "0-3-4-8"

`0,3` means cpu `0` and cpu `3`. It doesn't mean cpus `0`, `1`, `2` and `3` though right?

grpc.go
```diff
@@ -81,9 +81,53 @@ func handleError(wait bool, err error) error {
 	return err
 }
 
-// Online resources, nbResources specifies the maximum number of resources to online.
-// If nbResources is <= 0 then there is no limit and all resources are connected.
+// Offline resources, nbResources specifies the maximum number of resources to online.
```
devimc: Offline resources, ... maximum number of resources to **online** ?

sboeuf:
@justin-he 
Thanks for the PR. I have a few questions as I'm trying to understand the need for it.
>There is no acpi-driven cpu hotplug on arm till now.

So do you mean that it is supported by QEMU now? Which version supports it?

>But we can provide a cpu pool implementation instead.

This is a cool idea, but this means you will allocate the VM with the max amount of CPUs and then you will offline them from inside the VM with the agent. Is there any performance drawback doing this?

And more globally, this should be a virtio feature to provide a paravirtualized way of adding/removing CPUs based on a pool of available ones.

justin-he:
> @justin-he
> Thanks for the PR. I have a few questions as I'm trying to understand the need for it.
> 
> > There is no acpi-driven cpu hotplug on arm till now.
> 
> So do you mean that it is supported by QEMU now? Which version supports it?
> 
Do you mean the cpu hotplug non-acpi way on arm64?
Unfortunately, till now, there is no such support.
In my local test, I implement a hacked way to mock cpu hotplug adding/removing at qemu side.

> > But we can provide a cpu pool implementation instead.
> 
> This is a cool idea, but this means you will allocate the VM with the max amount of CPUs and then you will offline them from inside the VM with the agent. Is there any performance drawback doing this?
After this patch set, the kata on arm64 will not always allocate the VM with the max amount of CPUs.
It will be limited by the "default_maxvcpus" in toml file.
If default_maxvcpus = 0, currently I will only allocate 8 vcpus by passing "-smp 8" to qemu parameter.

> 
> And more globally, this should be a virtio feature to provide a paravirtualized way of adding/removing CPUs based on a pool of available ones.
Could you please elaborate it more? IIUC, virtio feature is just to avoid context switch to speed up the data transferring btw guest and host. Please correct me :)

grpc.go
justin-he: yes, thanks for the pointing

grpc.go
justin-he: yes,

grpc.go
justin-he: > * I think this would be clearer as:
>   ```go
>   agentLog.Warnf("offlining last available CPU disallowed")
>   ```
> * Paranoia: handle `lastOnlineCpu < 0` case?
>   This is partially handled by the `for` loop below, but if `lastOnlineCpu == -1`, the function will return `(-1, nil)` rather than an error.

func offlineResources(resource onlineResource, nbResources int32) (uint32, error)
The first return value should be uint32.  And in offlineCPUResources
`        var count uint32
        for i := uint32(0); i < cpuStateChangeMaxTries; i++ {
                r, err := offlineResources(resource, int32(nbCpus-count))
                if err != nil {
                        return err
                }
                count += r`
So if return (-1,nil), count might be decreased exceptionally?

protocols/grpc/agent.proto
```diff
@@ -309,6 +310,18 @@ message OnlineCPUMemRequest {
 	// CpuOnly specifies whether only online CPU or not.
 	bool cpu_only = 3;
 }
+message OfflineCPUMemRequest {
+	// Wait specifies if the caller waits for the agent to online all resources.
```
amshinde: /to online/to offline/

protocols/grpc/agent.proto
```diff
@@ -309,6 +310,18 @@ message OnlineCPUMemRequest {
...
+	// If true the agent returns once all resources have been connected, otherwise all
+	// resources are connected asynchronously and the agent returns immediately.
+	bool wait = 1;
+
+	// NbCpus specifies the number of CPUs that were added and the agent has to online.
```
amshinde: /online/offline/

protocols/grpc/agent.proto
```diff
@@ -309,6 +310,18 @@ message OnlineCPUMemRequest {
...
+	uint32 nb_cpus = 2;
+
+	// CpuOnly specifies whether only online CPU or not.
```
amshinde: /online/offline/

amshinde:
@justin-he I have similar questions as @sboeuf.
I would like to understand the approach you are taking towards handling CPUs on arm, mainly interaction between runtime and the agent for CPU requests and how this differs from x86_64.
Can you add more details about the design?

protocols/grpc/agent.proto
justin-he: > @justin-he I have similar questions as @sboeuf.
> I would like to understand the approach you are taking towards handling CPUs on arm, mainly interaction between runtime and the agent for CPU requests and how this differs from x86_64.
> Can you add more details about the design?

Sure.
The target is to support cpu limitation for “docker run” cmd in Kata-containers project
e.g.
$ docker run --runtime=kata-runtime -ti --cpus 2 debian bash
$ docker update --cpus 4 foo
$ docker update --cpus 2 foo
Difficulties:
The best way to enable cpu hotplug for kata-containers is supporting acpi driven cpu hotplug in qemu guest. But there are some obstacles which can’t be overcome.
- The apci driven cpu hotplug is not supported in qemu guest till now, and even no clear plan for the future as far as I know.
- limitation at kvm side for gic to handle hot-adding cpu after booting. 
Mentioned by Mark Rutland:
>We'd need something along those lines. Each CPU has a notional point to
>point link to the interrupt controller (to the redistributor, to be
>precise), and this entity must pre-exist.

Thus we need to find some workaround.

My proposals:
qemu cpu hotplug(adding) process, i.e. create a cpus pool for the qemu guest
 ```
+---------------+               +----------------+         +-----------------+
 |   kata qmp    |               |     qemu       |         |     guest       |
 +-------+-------+               +----------------+         +--------+--------+
         |                      start guest with -smp 4              |
         |                       -append “maxcpus=1”                 |
         |                                |                          |
         |                                +------------------------->|
         |                                |                    start cpu with
         |                                |                   1 online,3 offline
         |                                |                          |
         |  query_hotpluggable_cpus       |                          |
         +-------------------------------->                          |
         |                                |                          |
         | return cpu node info           |                          |
         +<-------------------------------+                          |
         |                                |                          |
         |      device_add cpu            |                          |
         +------------------------------->+                          |
         |                                |                          |
         |                 online/offline cpus by gRPC               |
         +---------------------------------------------------------->+
         |                                |                          |
         +                                +                          +
```
1. When changed qemu run -smp 4 -append “maxcpus=1”, It can create 4 cpus in 
   device-tree, and set 3 as pluggable. Actually, now there are 4 cpu nodes in 
   the device-tree. This work around the vcpu re-init obstacle at kvm side.
2. Then the guest kernel will be boot with 4 cpus, but let 3 be offline
3. use qmp cmd "query_hotpluggable_cpus" to get the pluggable cpus info in the
   guest.
4. When qmp cmd "device_add" is triggered, guest will be notified to set one 
   pluggable cpus as online.
5. Use gRPC req “onlineCPU” to set the wanted cpus to online at kata agent side.
6. Use gRPC req “offlineCPU” to set the wanted cpus to offline at kata agent side.

**Comment by sboeuf** [comment]
@justin-he thanks for the summary, I think this makes things very clear.

I think this should be more widely discussed at the Kata Architecture level since we could apply the same thing for any architecture, which would provide CPU hotplug for all architectures.
This would be a great opportunity to simplify the Kata architecture specific code, and would even allow us to support CPU hotplug for other hypervisors like Firecracker.

/cc @mcastelino

**Comment by sameo** [comment]
I feel this is an orchestration workaround that neither the hypervisor nor the architecture support. I'd prefer to get that feature implemented from the hypervisor.

**Comment by sboeuf** [comment]
@bergwolf @WeiZhang555 please give us your feedback.

**Comment by WeiZhang555** [comment]
As mentioned in https://github.com/kata-containers/runtime/issues/1384#issuecomment-474329194, I don't think it's good to security since we don't trust guest OS, this breaks one isolation layer though we still have host cgroup's protection.

A better way is adding feature to Hypervisor just as @sameo said.

**Comment by sboeuf** [comment]
@WeiZhang555 Thanks for chiming in, that's very important feedback here, as we're talking about security design of Kata, based on the features we want to provide.

**Comment by bergwolf** [comment]
I agree with @WeiZhang555 and @sameo here. Let's see if it can be fixed by hypervisor first. @justin-he any chance arm can support it? Is it a hardware/spec limitation or is it just lack of software support?

File: grpc.go [hunk 10]
```diff
@@ -81,9 +81,59 @@ func handleError(wait bool, err error) error {
...
+// Returns the number of resources disconnected.
+func offlineResources(resource onlineResource, nbResources int32) (uint32, error) {
+	// Get the bitmap of the online cpus
+	onlineCPUBitmapPath := filepath.Join(resource.sysfsOnlinePath, "online")
+	status, err := ioutil.ReadFile(onlineCPUBitmapPath)
+	if err != nil {
+		return 0, err
+	}
+
+	// Get the last online cpu, e.g. 0,3,4-7  cpu7 is the last online cpu
```
jodh-intel: I still find this block confusing - if you just need the last cpu -- assuming the `online` file is guaranteed to be ordered -- why do you need to do the split?

I think it would help if you split `offlineResources()` into a number of functions like:

- `getCPUsToOffline(...) []int`
- `offlineCPU(cpuNum int32) error`
- ...

Then, you can add a new unit tests to assert which cpus should be offlined and which should be left online. A table-based approach like the following would be ideal for testing `getCPUsToOffline()` for example:

- https://github.com/kata-containers/agent/blob/master/config_test.go#L106..L137

justin-he:
> I agree with @WeiZhang555 and @sameo here. Let's see if it can be fixed by hypervisor first. @justin-he any chance arm can support it? Is it a hardware/spec limitation or is it just lack of software support?

Hi @bergwolf Unfortunately，acpi cpu hotplug support on arm64 in qemu/fw(edk2 aavmf) is far away and not within the foreseeable future. And it is even harder than acpi memory hotplug. 
I thought it is not a hardware limitation, but the implementation needs a long time to go.

grpc.go
justin-he: hi @jodh-intel Thanks for the comments, but seems my proposal hasn't been accepted yet :(

jcvenegas:
@justin-he  @bergwolf  is this  the way we want to go ? or the PR should be closed?

PR #480, opened 2019-03-14
service: Fix user initiated shutdown

### LINKED ISSUE #479: Fix user-initiated shutdown
Revert 272f273961e70096938ac7a3fe6477023c8f715a to unbreak container shutdown when user types `exit` / control-d.

jodh-intel:
Partially revert commit 272f273961e70096938ac7a3fe6477023c8f715a to stop a user-initiated shutdown (typing `exit` or control-d for example at a shall prompt) from hanging and leaving the hypervisor running.

Fixes #479.

### DIFF
--- a/TRACING.md
+++ b/TRACING.md
@@ -7,6 +7,7 @@
 * [Agent tracing terminology](#agent-tracing-terminology)
     * [Trace modes](#trace-modes)
     * [Trace types](#trace-types)
+* [Agent shutdown behaviour](#agent-shutdown-behaviour)
 * [Enabling tracing](#enabling-tracing)
     * [Enabling dynamic tracing](#enabling-dynamic-tracing)
     * [Enabling static tracing](#enabling-static-tracing)
@@ -36,9 +37,7 @@ the following attributes:
 - A reference to the span's parent span
 
 All spans need to be *finished*, or *completed*, to allow the OpenTracing
-framework to generate the final trace information. This implies that the
-agent must be shutdown to obtain a full trace. This will occur after the
-workload has ended.
+framework to generate the final trace information.
 
 # Jaeger tracing architecture
 
@@ -104,7 +103,7 @@ The agent supports two different tracing modes:
 
 | Trace mode | Description | Use-case | VM shutdown controller | Limitations |
 |-|-|-|-|-|
-| Static | Traces from agent start to agent shutdown. | Obtain holistic view of agent activities. | Agent | None |
+| Static | Traces from agent start to agent shutdown. | Obtain holistic view of agent activities. | Agent | Requires systemd - see section [Agent shutdown behaviour](#agent-shutdown-behaviour). |
 | Dynamic | Traces begin when the gRPC `StartTracing()` API is called and end when the corresponding `StopTracing()` API is called. | on-demand (partial) tracing. | Runtime | Only "isolated" type supported. |
 
 ## Trace types
@@ -115,6 +114,21 @@ Each trace mode is sub-divided into two different types:
 |-|-|-|-|
 | isolated | The traces only apply to the agent; after the container has been destroyed, the first span will start at agent startup and the last at agent shutdown | Observing agent lifespan. | |
 | collated | In this mode, spans are associated with their `kata-runtime` initiated counterparts. | Understanding how the runtime calls the agent. | Requires runtime tracing to be enabled in `configuration.toml` (`enable_tracing=true`). |
+
+# Agent shutdown behaviour
+ 
+As shown in the [trace modes](#trace-modes) section, the different trace modes
+change how the VM is shutdown. In normal operation, the `kata-runtime` manages
+the VM shutdown. The **only** scenario the agent itself is responsible for
+VM shutdown is if static tracing is enabled. In this scenario, the
+agent finishes the trace spans and actually exits. The systemd unit that
+launches the `kata-agent` detects this and then runs the `ExecStop=` commands,
+which shutdown the VM.
+
+> **Note:** This implies that currently systemd is required for static
+> tracing. However, it is not a hard requirement. The agent can be modified to
+> invoke the `ExecStop=` commands itself to avoid relying on systemd to handle
+> VM shutdown.
 
 # Enabling tracing
 
--- a/TRACING.md
+++ b/TRACING.md
@@ -51,7 +51,7 @@ forwards the information on to other components of a Jaeger system.
 
 The Jaeger client package bound into the application sends the trace spans to
 the Jaeger agent process using the UDP protocol. This is problematic since it
-is not desirable to run a `jaeger-agent` process inside the VM:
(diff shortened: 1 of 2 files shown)


amshinde:
@sboeuf PTAL

sboeuf:
@jodh-intel I'm fine with the patch, but it'd be nice to understand why this is needed. We thought we could remove it and now we're saying that we need it because something is broken. It seems very important to me that we should narrow down the root cause for this.

devimc:
@jodh-intel do we have the same issue with initrd images?

jodh-intel:
Wait a sec - this is only needed when [static tracing](https://github.com/kata-containers/agent/blob/master/TRACING.md#primary-trace-types) is enabled.

@devimc - going further, it only affects static tracing with an image; static tracing with an initrd is fine! ;)

Still investigating...

jodh-intel:
Added dnm as this isn't required for non-trace scenarios.

jodh-intel:
@sboeuf - the problem afaics is that with static tracing enabled, *one* of the two following scenarios can happen when the user kills the container:

- Normal scenario:
  - The runtime calls `DestroySandbox()`.
  - `DestroySandbox()` sends back the response to the runtime.
  - The runtime receives the response then sends the QMP `QUIT` to kill the VM.
  - The VM shuts down and the runtime is happy.

- Alternative scenario:
  - The runtime calls `DestroySandbox()`.
  - The agent closes the `stopServer` channel.
  - The `waitForStopServer()` thread detects this and stops the gRPC server.
  - The agent exits.
  - Meanwhile, the runtime is still waiting for the response to its `DestroySandbox()` call, but that will never arrive now that the `gRPC` server has gone.
  - The runtime's vsock socket connection blocks forever waiting in vain for the server to respond.

Note that it is only the second scenario that requires this PR: the runtime will then get "a" response from `DestroySandbox()` (an error):

```
time="2019-03-15T14:49:39.631975139Z" level=debug msg="sending request" arch=amd64 command=delete container=26946cd0296341551926aa060ffafc2fd0135833432d772fc4099268a4f3f56d name=grpc.DestroySandboxRequest pid=21660 req= sandbox=26946cd0296341551926aa060ffafc2fd0135833432d772fc4099268a4f3f56d source=virtcontainers subsystem=kata_agent
time="2019-03-15T14:49:39.766877474Z" level=warning msg="Agent did not stop sandbox" arch=amd64 command=delete container=26946cd0296341551926aa060ffafc2fd0135833432d772fc4099268a4f3f56d error="rpc error: code = Unavailable desc = transport is closing" name=kata-runtime pid=21660 sandbox=26946cd0296341551926aa060ffafc2fd0135833432d772fc4099268a4f3f56d sandboxid=26946cd0296341551926aa060ffafc2fd0135833432d772fc4099268a4f3f56d source=virtcontainers subsystem=sandbox
```

That error allows the runtime to unblock and of course by that stage the VM has been shutdown by systemd thanks to `ExecStop=`.

I've updated the service file to add an explanatory comment.

jodh-intel:
> We should write down somewhere that static tracing is not going to be supported without systemd.

I don't quite follow? The current agent tracing requires systemd for the Jaeger vsock->udp proxy but that applies to both static and dynamic tracing. Once we switch to opencensus, we won't need that systemd proxy job in the guest but that's true for dynamic and static tracing I think?

> Also, another comment is that we assume the agent is not supposed to terminate, except for the specific case of static tracing.

Agreed - I'll update the docs on this PR once #489 lands so marking as dnm for now.

sboeuf:
@jodh-intel 
>I don't quite follow? The current agent tracing requires systemd for the Jaeger vsock->udp proxy but that applies to both static and dynamic tracing. Once we switch to opencensus, we won't need that systemd proxy job in the guest but that's true for dynamic and static tracing I think?

My point is that once we move to Opencensus, we won't need systemd at all to get the dynamic tracing. But in case of the static one, because you need the agent to terminate, you will need the systemd service to shutdown the VM (unless you were coding the system shutdown directly part of the agent).

jodh-intel:
Ah - I see your point. Yes, once this PR lands we shall be using systemd to actually cause the agent to exit and shutdown the VM but only in the static tracing scenario. Now the agent code has been reworked, we could just make the agent exec the shutdown command itself when we move to opencensus + static tracing + agent-as-init case since we stop the tracing and report the spans before agent shutdown. That means we get to a point where we no longer care about the agent process when it's about to be killed so either systemd can do it or the agent can do it itself.

But until that happens, like the other point, I'll update `TRACING.md` on this PR once #489 lands...

jodh-intel:
@sboeuf - branch updated with new doc changes.

TRACING.md
```diff
@@ -116,6 +115,21 @@ Each trace mode is sub-divided into two different types:
 | isolated | The traces only apply to the agent; after the container has been destroyed, the first span will start at agent startup and the last at agent shutdown | Observing agent lifespan. | |
 | collated | In this mode, spans are associated with their `kata-runtime` initiated counterparts. | Understanding how the runtime calls the agent. | Requires runtime tracing to be enabled in `configuration.toml` (`enable_tracing=true`). |
 
+# Agent shutdown behaviour
+ 
+As shown in [Trace modes](#trace-modes) section, the different trace modes
```
klynnrif: Add "the": 
As shown in the [Trace modes](#trace-modes) section, the different trace modes

TRACING.md
```diff
@@ -116,6 +115,21 @@ Each trace mode is sub-divided into two different types:
...
+change how the VM is shutdown. In normal operation, the `kata-runtime` manages
+the VM shutdown. The **only** scenario when the agent itself is responsible for
```
klynnrif: Suggested rewrite: 
.... The **only** scenario the agent itself is responsible for VM shutdown is if static tracing is enabled.

TRACING.md
```diff
@@ -116,6 +115,21 @@ Each trace mode is sub-divided into two different types:
...
+shutting down the VM is when static tracing is enabled. In this scenario, the
+agent finishes the trace spans and actually exits. The systemd unit that
+launches the `kata-agent` detects this and then runs the `ExecStop=` commands,
+which shutdown the VM.
+
+> **Note:** that this implies that currently systemd is required for static
```
klynnrif: Lines 129-131 suggested rewrite: 
**Note:** This implies that currently systemd is required for static tracing. However, it is not a hard requirement. The agent can be modified to invoke the `ExecStop=` commands itself to avoid relying on systemd to handle VM shutdown.

jodh-intel:
Thanks @klynnrif - branch updated.

jodh-intel:
ppc64le CI still appears to be borked as the job is never run seemingly?:

```
Building remotely on Power8_slave01 (power8-ubuntu-1604 power8_node) in workspace /home/jenkins/workspace/kata-containers-agent-Power8-ubuntu-16-04-PR-initrd
[WS-CLEANUP] Deleting project workspace...
```

/cc @nitkon.

PR #483, opened 2019-03-15
tracing: Hide trace messages unless tracing enabled

### LINKED ISSUE #482: Hide trace message unless tracing is really enabled
Since #415 landed, the agent logs show trace-related messages like `created span...` even when tracing is not enabled.

This is technically correct, but confusing to users: when tracing is disabled, trace spans are still created - but the tracer used is a NOP. Therefore, the agent should only display the message when tracing is explicitly enabled.

jodh-intel:
Don't log trace-related messages unless tracing is actually enabled. If tracing is not enabled, trace spans are still created, but the tracer is a NOP. However, this is an implementation detail the user does not need
to be aware of so hide trace logs unless the user has enabled tracing explicitly.

Fixes #482.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -95,7 +95,10 @@ func setupTracing(rootSpanName string) (opentracing.Span, context.Context, error
 	span.SetTag("source", "agent")
 	span.SetTag("root-span", "true")
 
-	agentLog.Debugf("created root span %v", span)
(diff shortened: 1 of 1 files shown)


amshinde:
Restarting firecracker and metrics CI.

PR #493, opened 2019-03-20
network: guest network interface support NOARP

zhabinecho:
According to the host link rawflags, setting link NOARP when needed

Fixes: kata-containers#492

Signed-off-by: Zha Bin <zhabin@linux.alibaba.com>


grahamwhaley:
@jodh-intel - like the parallel PR over in the runtime, this has failed the 'Fixes' check, even though I can see one - I wonder if maybe it could be a character encoding issue?
```
Found 1 commit between commit HEAD and branch master
ERROR: No "Fixes" found
ERROR: checkcommits failed. See the document below for help on formatting
commits for the project.
```

Any ideas?

jodh-intel:
Ah - I have see it now! The "raw" commit shows:

```
Fixes: kata-containers#492
```

... but github unhelpfully in this case renders it as:

```
Fixes: #492
```

The problem is that our [`checkcommits`](https://github.com/kata-containers/tests/tree/master/cmd/checkcommits) tool doesn't understand that syntax.

@zhabinecho - please could you change the "Fixes" line of you commit to be *exactly* as follows:

```
Fixes: #492
```

zhabinecho:
> Ah - I have see it now! The "raw" commit shows:
> 
> ```
> Fixes: kata-containers#492
> ```
> ... but github unhelpfully in this case renders it as:
> 
> ```
> Fixes: #492
> ```
> The problem is that our [`checkcommits`](https://github.com/kata-containers/tests/tree/master/cmd/checkcommits) tool doesn't understand that syntax.
> 
> @zhabinecho - please could you change the "Fixes" line of you commit to be _exactly_ as follows:
> 
> ```
> Fixes: #492
> ```

OK, I will fixed it. Thanks

pkg/types/types.proto
```diff
@@ -36,6 +36,7 @@ message Interface {
 	// library, regarding each type of link. Here is a non exhaustive
 	// list: "veth", "macvtap", "vlan", "macvlan", "tap", ...
 	string type = 7;
+	uint32 rawFlags = 15;
```
sboeuf: 15? What happened to 8, 9, ....?
This is a protobuf message definition here, so I would expect:
```go
uint32 flags = 15;
```

Also, is it necessary to call this `rawFlags`? A simple `flags` should be enough, right?

zhabinecho: My first idea is that reserved other tag numbers, also numbers 1-15 require one less byte to encode than higher numbers, So I use it. If you don't mind, I would like use the number 8.

The LinkAttrs struct use the RawFlags member in kata-runtime, so maybe the rawflags has some link meanging, not just a flags

sboeuf: >If you don't mind, I would like use the number 8. :-)

Yes please use 8.

>The LinkAttrs struct use the RawFlags member in kata-runtime, so maybe the rawflags has some link meaning, not just a flags?

Ok let's keep the name `RawFlags` then.

zhabinecho: Fixed it done.

pkg/types/types.proto
```diff
@@ -36,6 +36,7 @@ message Interface {
 	// library, regarding each type of link. Here is a non exhaustive
 	// list: "veth", "macvtap", "vlan", "macvlan", "tap", ...
 	string type = 7;
+	uint32 RawFlags = 8;
```
sboeuf: Sorry to be picky but we should follow the protobuf syntax defined [here](https://developers.google.com/protocol-buffers/docs/proto#simple). Therefore, `RawFlags` should be `raw_flags` here.
But I noticed we forgot to enforce this for other fields of this structure, which is confusing...

zhabinecho: When use raw_flags, it can generate  different gzipped FileDescriptorProto  with rawFlags or RawFlags. Maybe we use rawFlags in order to be consistent with other fields. Next, I will submit another PR to follow the protobuf?

zhabinecho: I don't understand why the  continuous-integration failed ?

gnawux: The travis tests on ppc64le looks not stable today

teawater:
```
INFO: Forcing gometalinter version 
Installing:
  deadcode
  dupl
  errcheck
  gochecknoglobals
  gochecknoinits
  goconst
  gocyclo
  goimports
  golint
  gosec
  gotype
  gotypex
  ineffassign
  interfacer
  lll
  maligned
  misspell
  nakedret
  safesql
  staticcheck
  structcheck
  unconvert
  unparam
  varcheck
INFO: gometalinter args: '--exclude=vendor/.*" --tests --exclude=".*\.pb\.go" --disable-all --enable=gofmt --concurrency=2 --enable=misspell --enable=vet --enable=ineffassign --enable=gocyclo --cyclo-over=15 --enable=golint --deadline=600s --enable=structcheck --enable=unused --enable=staticcheck --enable=maligned --enable=varcheck --enable=unconvert'
```

```
gometalinter: error: unknown linters: unused
The command ".ci/static-checks.sh" failed and exited with 1 during .
```
I just met the issue in my host.  Maybe gometalinter removed unused.  I will take a look on it.

teawater: I think we can just wait for https://github.com/kata-containers/tests/issues/1323 to fix this issue.

zhabinecho: Can we merge it ?  :-)

jodh-intel:
@sboeuf - please can you re-review?

sboeuf:
LGTM

PR #496, opened 2019-03-20
agent: support read-only rootfs

### LINKED ISSUE #495: agent: unsupported read-only rootfs
Could not start a container if the agent is init process and the rootfs is a read-only filesystem

devimc:
mount /run as a tmpfs to support read-only rootfs.

fixes #495

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -62,6 +62,7 @@ var initRootfsMounts = []initMount{
 	{"devtmpfs", "dev", "/dev", []string{"nosuid"}},
 	{"tmpfs", "tmpfs", "/dev/shm", []string{"nosuid", "nodev"}},
 	{"devpts", "devpts", "/dev/pts", []string{"nosuid", "noexec"}},
+	{"tmpfs", "tmpfs", "/run", []string{"nosuid", "nodev"}},
 }
 
 type process struct {


agent.go
```diff
@@ -62,6 +62,7 @@ var initRootfsMounts = []initMount{
 	{"devtmpfs", "dev", "/dev", []string{"nosuid"}},
 	{"tmpfs", "tmpfs", "/dev/shm", []string{"nosuid", "nodev"}},
 	{"devpts", "devpts", "/dev/pts", []string{"nosuid", "noexec"}},
+	{"tmpfs", "tmpfs", "/run", []string{"nosuid", "nodev"}},
```
sboeuf: This is handling the case where the agent runs as init. What about the systemd case? Is it directly handled by systemd?

amshinde: @devimc Why is this required in case of initrd. I suppose we dont mount the /run as tmpfs for systemd based rootfs.
I would be wary of mounting anything as tmpfs when it is not specified in the config.json as this adds to our overall memory consumption in case /run is written to.(This may mess up resource accounting)

cc @mcastelino

devimc: @sboeuf 
> What about the systemd case? Is it directly handled by systemd?

yes, systemd mount it as a `tmpfs` (check in your host)


@amshinde 
> I would be wary of mounting anything as tmpfs when it is not specified in the config.json as this adds to our overall memory consumption in case /run is written to.(This may mess up resource accounting)

sorry I don't understand your concern, this patch is to mount the guest `/run` as a tmpfs, instead of using the disk (NVDIMM),  the config.json is only used for containers, isn't it?

amshinde: @devimc my bad, I mistook it as container run.

devimc:
can we merge it?

PR #499, opened 2019-03-21
make: Install systemd targets in systemd unit dir

### LINKED ISSUE #498: Install systemd units after evaluating the systemd unit-dir configured for the package.
Install the agent systemd targets after evaluating the systemd
unitdir configured for the systemd package.
We were initially assuming this as "/usr/lib/systemd/system/"
but this is not the case for all distros such as debian and ubuntu.

amshinde:
Install the agent systemd targets after evaluating the systemd
unitdir configured for the systemd package.
We were initially assuming this as "/usr/lib/systemd/system/"
but this is not the case for all distros such as debian and ubuntu.

Fixes #498

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,12 @@ TRACE_DEV_MODE := no
 # Path to systemd unit directory if installed as not init.
 UNIT_DIR := /usr/lib/systemd/system
 
+HAVE_SYSTEMD := $(shell pkg-config --exists systemd 2>/dev/null && echo 'yes')
+
+ifeq ($(HAVE_SYSTEMD),yes)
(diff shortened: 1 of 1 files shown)


amshinde:
@grahamwhaley Is the metrics still flaky?

grahamwhaley:
@amshinde yeah, afraid so: http://jenkins.katacontainers.io/computer/x86_packet02/builds
I may not have time to look at it until the middle of next week I'm afraid though.
I'm not sure it is a quite trivial fix - I think we are still getting 'one of two results' per run, and I don't know why - it seemed to start happening when we did some optimisations, like systemd trimming etc.

amshinde:
@sboeuf @devimc @jcvenegas Can this be merged?

PR #507, opened 2019-03-29
vendor: use latest github.com/mdlayher/vsock

### LINKED ISSUE #485: vendor: upgrade github.com/mdlayher/vsock
Hey folks! I just wanted to let you know that I've landed Go 1.12 runtime network poller support in https://github.com/mdlayher/vsock, meaning that the vsock package should now comply completely with `net.Listener` and `net.Conn`! I'd recommend upgrading the package when possible.

Go 1.11 also still works, but you have to put a loop around the Accept method in order to wait for incoming connections. This is no longer necessary as of Go 1.12.

Please do file an issue if you run into anything unusual!

mdlayher:
This version of vsock fully supports Go 1.12 runtime network poller
integration, and is compliant with net.Listener and net.Conn as
checked by golang.org/x/net/nettest.

Fixes #485.

Signed-off-by: Matt Layher <mdlayher@gmail.com>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -118,11 +118,11 @@
   revision = "cc6d2ea263b2471faabce371255777a365bf8306"
 
 [[projects]]
-  digest = "1:f3e11773cc8a305f97b42c276f62d92899c9b0c1c3c584a667149b18f18a530a"
+  digest = "1:79062b1e6d284b7a41cf04ba20f5dbc371a4fdd5edaca748ddb827ab07846dbf"
   name = "github.com/mdlayher/vsock"
   packages = ["."]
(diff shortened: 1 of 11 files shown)


bergwolf:
Although we have quite a few CI jobs for the agent, I don't see `jenkins-ci-fedora-vsocks` listed there, or any one having a `-vsock` suffix. Are we sure vsock is tested here?

sboeuf:
Good point @bergwolf 
@chavafg any input about our lack of `fedora-vsock` job here?

chavafg:
Ups, seems like I forgot to add it. It should now report.

/test

sboeuf:
Thanks @chavafg :)

PR #508, opened 2019-04-02
channel: Check for channel type in kernel cmdline options

### LINKED ISSUE #506: agent cannot rely on existence of /dev/vsock for determining vsock support
see https://github.com/kata-containers/runtime/issues/1378#issuecomment-477884710

amshinde:
With a recent kernel change, the agent can no longer rely on
/dev/vsock and AF_VSOCK in socket(), to detect if vhost-vsock
channel has been passed by the runtime.
These are not created when the vhost-vsock driver is initialised.

The runtime should now pass this information explicilty.
Based on the channel type passed, the agent now checks for that
partcular channel type.
In case it is not passed, check for both serial and vsock channel.
This also introduces a change in the way vsock channel is detected
by checking for devices under the vhost-vsock driver path.

Fixes #506

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -156,6 +156,22 @@ var collatedTrace = false
 // if true, coredump when an internal error occurs or a fatal signal is received
 var crashOnError = false
 
+// commType is used to denote the communication channel type used.
+type commType int
+
+const (
+	// virtio-serial channel
+	serialCh commType = iota
+
+	// vsock channel
+	vsockCh
+
+	// channel type not passed explicitly
+	unknownCh
+)
+
+var commCh = unknownCh
+
 // This is the list of file descriptors we can properly close after the process
 // has been started. When the new process is exec(), those file descriptors are
 // duplicated and it is our responsibility to close them since we have opened
--- a/channel.go
+++ b/channel.go
@@ -50,24 +50,29 @@ func newChannel(ctx context.Context) (channel, error) {
 	defer span.Finish()
 
 	var serialErr error
-	var serialPath string
 	var vsockErr error
-	var vSockSupported bool
+	var ch channel
 
 	for i := 0; i < channelExistMaxTries; i++ {
-		// check vsock path
-		if _, err := os.Stat(vSockDevPath); err == nil {
-			if vSockSupported, vsockErr = isAFVSockSupportedFunc(); vSockSupported && vsockErr == nil {
-				span.SetTag("channel-type", "vsock")
-				return &vSockChannel{}, nil
+		switch commCh {
+		case serialCh:
+			if ch, serialErr = checkForSerialChannel(ctx); serialErr == nil && ch.(*serialChannel) != nil {
+				return ch, nil
 			}
-		}
+		case vsockCh:
+			if ch, vsockErr = checkForVsockChannel(ctx); vsockErr == nil && ch.(*vSockChannel) != nil {
+				return ch, nil
+			}
 
-		// Check serial port path
-		if serialPath, serialErr = findVirtualSerialPath(serialChannelName); serialErr == nil {
-			span.SetTag("channel-type", "serial")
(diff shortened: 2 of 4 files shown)


agent.go
```diff
@@ -156,6 +156,22 @@ var collatedTrace = false
...
 
+// commType is used to denote the communication channel type used.
+type commType int
+
+const (
+	// virtio-serial channel
+	serialCh commType = iota
+
+	// vsock channel
+	vsockCh
```
sboeuf: Because there's a new line and comments between `serialCh` and `vsockCh`, and `unknownCh`, I think both `vsockCh` and `unknownCh` are going to be typed as `int` instead of `commType`.

agent.go
```diff
@@ -156,6 +156,22 @@ var collatedTrace = false
 // if true, coredump when an internal error occurs or a fatal signal is received
 var crashOnError = false
 
+// commType is used to denote the communication channel type used.
+type commType int
+
```
sboeuf: In the commit message, you wrote:
```
The runtime should not pass this information explicilty
```
Did you mean:
```
The runtime should now pass this information explicitly
```
?

amshinde: @sboeuf Ahh bad typo!! Fixed this.

agent.go
amshinde: @sboeuf I was able to assign between those values, I think compile would have complained in that case.

sboeuf: You can still compile since it's a cast later in the code, but unless the Go compiler got smarter, this was a problem I had encountered before.

agent.go
sboeuf: Cool!

channel.go
```diff
@@ -84,6 +89,44 @@ func newChannel(ctx context.Context) (channel, error) {
...
+
+	return nil, serialErr
+}
+
+func checkForVsockChannel(ctx context.Context) (*vSockChannel, error) {
+	span, _ := trace(ctx, "channel", "checkForVsockChannel")
+	defer span.Finish()
+
+	// check vsock path
+	var err error
```
devimc: useless variable, please use
```go
if _, err := os.Stat(vSockDevPath); err != nil {
```

config.go
```diff
@@ -20,6 +21,7 @@ const (
 	logLevelFlag       = optionPrefix + "log"
 	devModeFlag        = optionPrefix + "devmode"
 	traceModeFlag      = optionPrefix + "trace"
+	useVsockFlag       = optionPrefix + "use_vsock"
```
jodh-intel: We are now starting to document agent options in the top-level `README.md`. This isn't a general options like debug/trace but I wonder if we should add it anyway as it could be useful to someone?

channel.go
amshinde: @devimc Fixed.

agent.go
amshinde: No golang will never cast automatically.
Looks like newline are allowed according to this wiki: https://github.com/golang/go/wiki/Iota

config.go
amshinde: @jodh-intel Tests added. 
This option is something that should be added by the runtime, I dont think we should expose it to the user.

channel.go
```diff
@@ -228,23 +268,51 @@ func (c *serialChannel) teardown() error {
 	return c.serialConn.Close()
 }
 
+// isAFVSockSupported checks if vsock channel is used by the runtime
+// by checking for devices under the vhost-vsock driver path.
+// It returns true if more a device is found for the driver.
```
egernst: nit: can you clarify this last sentence?

config.go
egernst: makes sense @amshinde

channel.go
amshinde: done

Issue #511, opened 2019-04-03
"agent grpc server quits" show wrong error


teawater:
It should show servErr but not err.

PR #512, opened 2019-04-03
agent: Fix "agent grpc server quits" show wrong error

teawater:
It should show servErr but not err.

Fixes: #511

Signed-off-by: Hui Zhu <teawater@hyper.sh>


jodh-intel:
Good catch @teawater !

Issue #516, opened 2019-04-04
travis: ci: test repository does not chekout to stable branches when needed


chavafg:
If testing stable branch, `clone_tests_repo` must checkout test repository to stable branch

PR #517, opened 2019-04-04
ci: travis: checkout test repo to correct branch

chavafg:
If we are testing an stable branch, we need to checkout
to the correct branch on the tests repository.

Fixes: #516.

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -16,6 +16,10 @@ clone_tests_repo()
 	fi
 
 	go get -d -u "$tests_repo" || true
+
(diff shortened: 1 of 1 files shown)


egernst:
Ah.  I cancelled all the CI (abort) and just kept required, since this doesn't actually touch any code.  Only useful feedback is travis-ci for this PR.

We had a lot of PRs through for bumping versions and not touching code, so we were trying to not clog the CI pipeline, otherwise nothing would complete....

Issue #535, opened 2019-04-17
Add network stats


ZeroMagic:
# Description of problem

Currently KataContainers collect the stats from libcontainer Stats() API.

The libcontainer Stats() API has collected the network stats. However KataContainers do not transfer the network stats now. Can we add network stats ? The network stats will help us to improve our service quality.

```proto
message NetworkStats {
	string name = 1;
	uint64 RxBytes = 2;
	uint64 RxPackets = 3;
	uint64 RxErrors = 4;
	uint64 RxDropped = 5;
	uint64 TxBytes = 6;
	uint64 TxPackets = 7;
	uint64 TxErrors = 8;
	uint64 TxDropped = 9;
}
```

# Expected result

kata-agent will transfer the network stats to kata-runtime. Then I will also do some work in kata-runtime to convert vc.ContainerStats to containerd/cgroups.Metrics. I have proposed a pull request in containerd/cgroups repo. https://github.com/containerd/cgroups/pull/81

bergwolf:
Looks good! We can proceed once containerd merges containerd/cgroups#81

ZeroMagic:
Ok. We can wait for the merging to be completed

PR #538, opened 2019-04-18
proto: Add network stats

ZeroMagic:
Ref: #535 
Add network stats so that kata-runtime can get them and provide them to containerd.

Signed-off-by: ZeroMagic <anthonyliu@zju.edu.cn>


ZeroMagic:
@jodh-intel Done.

PR #543, opened 2019-04-26
# Kata Containers this-is-a-test-2

katabuilder:
- proto: Add network stats

d9aa453 proto: add network stats


chavafg:
is this still needed or can we close? 
@jcvenegas

Issue #544, opened 2019-04-30
updateInterface: fail to hot-add nic to container on arm64


jongwu:
# Description of problem
when hot-add nic to a kata-container on arm64, updateInterface will report error of time out to wait for nic device

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #545, opened 2019-04-30
updateInterface: enable hot-add nic on arm64

jongwu:
For now, update interface in agent will fail when hot-add
    nic to a running containers on arm64 as rescan pci bus will
    occur between uf and bf of shpc hotplug interrupt handling.

    Another problem is that the rootBusPath will be
    "/devices/platform/4010000000.pcie/pci0000:00" on arm64.

    To enable hot-add nic on arm64, rootBusPath should be changed here
    and shpc hotplug should be disabled in guest kernel.

    This patch just change rootBusPath.

Fixes: #544

@bergwolf @jodh-intel @teawater @Pennyzct

### DIFF
--- a/device_amd64.go
+++ b/device_amd64.go
@@ -1,0 +1,12 @@
+//
+// Copyright (c) 2019 Intel Corporation
+// Copyright (c) 2019 ARM Limited
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+const (
+	rootBusPath = "/devices/pci0000:00"
+)
--- a/device_ppc64le.go
+++ b/device_ppc64le.go
@@ -1,0 +1,12 @@
+//
+// Copyright (c) 2019 Intel Corporation
+// Copyright (c) 2019 ARM Limited
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+const (
+	rootBusPath = "/devices/pci0000:00"
+)
--- a/device.go
+++ b/device.go
@@ -35,8 +35,7 @@ const (
 )
 
 const (
-	rootBusPath = "/devices/pci0000:00"
-	pciBusMode  = 0220
+	pciBusMode = 0220
 )
 
 var (
--- a/device_arm64.go
+++ b/device_arm64.go
@@ -1,0 +1,9 @@
+//
+// Copyright (c) 2019 ARM Limited
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+const rootBusPath = "/devices/platform/4010000000.pcie/pci0000:00"
--- a/device_s390x.go
+++ b/device_s390x.go
@@ -1,0 +1,12 @@
+//
+// Copyright (c) 2019 Intel Corporation
+// Copyright (c) 2019 ARM Limited
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+const (
+	rootBusPath = "/devices/pci0000:00"
+)


device_amd64.go
```diff
@@ -0,0 +1,18 @@
...
+package main
+
+const (
+	rootBusPath = "/devices/pci0000:00"
+	pciBusMode  = 0220
+)
+
+var pciBusRescanFile = sysfsDir + "/bus/pci/rescan"
+
+func rescanPciBus() error {
```
jodh-intel: Since hot-add differs only on ARM, how about keeping the new files (`device_amd64.go`, `device_s390x.go`, *etc*), but having all non-ARM implementations call a generic handler like this:

```go
func rescanPciBus() error {
    return genericRescanPCIBus()
}
```

That would avoid the duplication.

jongwu: @jodh-intel thanks, if the implementation for rescanPciBus contains many lines of code, it is better to package them. but for only one line, keep it maybe better, as I fear of that more function call need more stack context switch which may be harmful to performance.

jongwu:
@teawater @bergwolf any comments?

jongwu:
could someone help me retrigger test?

grahamwhaley:
/test
sure

Pennyzct:
Hi~ @grahamwhaley could we add ARM CI machine here?? ;)

grahamwhaley:
@Pennyzct - I expect so. If you are confident it is stable enough, then I don't think anybody will object. The only 'rule' I try to lay down is that if it becomes unstable and starts to block PRs, *and* is not fixed in a reasonable time, then we may disable it again :-)
@chavafg - you OK with this?
So, @Pennyzct - we are also always looking for other members of the community to become helpers with the CI in general - would you like more access to the Jenkins master machine to help us with that? We should also discuss that ;-)

jodh-intel:
@grahamwhaley - could you add those CI arch requirements maybe to https://github.com/kata-containers/ci/blob/master/README.md?

grahamwhaley:
@jodh-intel - sure, I opened https://github.com/kata-containers/ci/issues/156 as a starter.

jongwu:
@jodh-intel yeah,  code duplication is not a good code style.

jongwu:
@grahamwhaley @bergwolf any comments?

grahamwhaley:
np from me. I'd quite like to hear from @devimc on it for the pci/hotplug stuff.
I nudged a rebuild of the jenkins-vsock - the only one failing, and I think that was a generic issue I'm hoping we fixed overnight...

jongwu:
@grahamwhaley thanks.

device_amd64.go
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright (c) 2019 Intel Corporation
```
jcvenegas: This is probably a wrong header

device_ppc64le.go
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright (c) 2019 Intel Corporation
```
jcvenegas: ARM Limited?

jongwu: @jcvenegas I'm confused with these copyright header, those no-arm64 code are all copy from the original file with Intel header. so I'm not sure how to write these header. @jodh-intel @grahamwhaley @bergwolf any idea?

jodh-intel: Thanks for raising this issue @jongwu! I agree that this is confusing, so we need to document this. IANAL so will need to get input from folks more knowledgeable in that area. Since this file is based on an existing one, I would think it is "safe" to always reference the original copyright *as well* (so that you could list both Intel and ARM for 2019).

@grahamwhaley - any thoughts on this? I'm happy to write it up once we have clarity on this subtle area.

As to a location for where to put the words.. I think it would either need to go in:

- https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md

- But, for parity with https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md, it may make sense to create a `https://github.com/kata-containers/documentation/blob/master/Source-File-Requirements.md`

Wherever it goes, it needs to cover the following copyright scenarios:

- Adding a new source file.
- Adding a new source file which is similar to an existing file:
  - Exact copy but for different arch (covering both a tiny fragment (like on this PR) and also a very large and complex file).
  - "Based on" an existing file but with potentially extensive modifications.
- ... ?

jongwu: Thanks @jodh-intel, it is very helpful for me. If I get it right, the copyright header for device_amd64.go, device_ppc64le.go and device_s390x.go should include both intel and ARM. Is that right?
If that is the truth, what's the shape of these header? eg. it should be "Intel Corporation and ARM limited" or "Intel Corporation && ARM limited" or something else?

grahamwhaley: afaik, you would put your new addition copyright on a new line, so like:

```
//
// Copyright (c) 2018 Intel Corporation
// Copyright (c) 2019 <your company/person name here>
//
// SPDX-License-Identifier: Apache-2.0
//
```

and replace `your company` etc. :-)

grahamwhaley: I should probably note - your company may have its own ideas, rules or specific name/format for this btw, and you might want to check internally first.

Pennyzct: // Copyright (c) 2019 ARM Limited
//
// SPDX-License-Identifier: Apache-2.0
//
plz use this. ;)

jongwu: Thanks @grahamwhaley

Pennyzct:
Hi~ @grahamwhaley 
Sorry for the delayed response about adding ARM CI machines on agent repo.
I got sidetracked for a few other issues. 😥
For now, there could only be me full-time dedicated to the ARM CI maintenance, in the future, there may be more arm folks.;) So if anytime ARM CI blocks PR here and doesn't get fixed for a quite time, you sure could bring it down.😄 
And I'm very happy to become helpers with the CI community and have more access to the Jenkins master machine to help you guys.
So All in all, let's bring ARM CI online agent repo. ;)

grahamwhaley:
@Pennyzct - as long as we are happy we have the capacity (as discussed on other Issue), sure - @chavafg maybe we should first add @Pennyzct to the Jenkins master as admin, and then guide on how to add (clone/edit) a new job... thx!

jongwu:
@devimc any comments?

Issue #542, opened 2019-04-24
go >= 1.11 required for proper vsock support


marcov:
(not sure this is really a bug, limitation or just something to note)

When building kata-agent with go < 1.11 (e.g. with go 1.10.4), vsock communication with the host does not work:

```
$ sudo kata-runtime --kata-config configuration-qemu.toml  run --bundle bundle myctr
Failed to check if grpc server is working: rpc error: code = Unavailable desc = transport is closing                                                                                          
```

grahamwhaley:
I would call it a bug today, as the version of golang we have set in our [versions file](https://github.com/kata-containers/runtime/blob/master/versions.yaml#L267) is still 1.10.4 :-)
We should:
 - clarify if we understand why this fails
 - probably bump out minimum golang to 1.11 at least. 1.10.x is now in the ['archived' section](https://golang.org/dl/) of golang releases I think...

marcov:
> I would call it a bug today, as the version of golang we have set in our [versions file](https://github.com/kata-containers/runtime/blob/master/versions.yaml#L267) is still 1.10.4 :-)

The agent does not look at versions.yaml (yet). So it would be worth doing adding the check.

And when building kata-agent in the osbuilder repo, the compiler version is read from [this file](https://github.com/kata-containers/osbuilder/blob/master/rootfs-builder/versions.txt), that as of now it's 1.11 

> We should:
> 
> * clarify if we understand why this fails

Definitely. I am thinking of a way to retrieve agent logs in the guest. Do you have any idea?

grahamwhaley:
I think @marcov if you enable proxy and agent debug, then the agent output will come out in the proxy system logs... I think. /cc @jodh-intel @jcvenegas @devimc

devimc:
@marcov yes , as @grahamwhaley mentioned, you should enable agent+proxy debug. Run `journalctl -b -t kata-proxy` to see the logs

marcov:
@devimc 
But doesn't the agent talk directly with the shim over vsock? 
...and, if the agent does not set up the vsock, then there aren't logs sent to the host.

jcvenegas:
@marcov  does proxy setup work ? if the proxy case does not work probably is something with our code, but if only vsock it may be the code that the agent vendors.

marcov:
The proxy (non-vsock) setup works. The problem is just when vsock is enabled.

devimc:
[...]

PR #551, opened 2019-05-07
vendor: update vsock package version

marcov:
The vsock package needs go >= 1.11. The latest version of the package adds a
compile time check on the go version.

Fixes: #542

Signed-off-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -126,11 +126,11 @@
   revision = "cc6d2ea263b2471faabce371255777a365bf8306"
 
 [[projects]]
-  digest = "1:79062b1e6d284b7a41cf04ba20f5dbc371a4fdd5edaca748ddb827ab07846dbf"
+  digest = "1:b09f78822be5d574e052d4e377b133bac445cdbcbfaf884e0704f870e509ef21"
   name = "github.com/mdlayher/vsock"
   packages = ["."]
   pruneopts = "NUT"
-  revision = "a92c53d5dcabbf6aaaba8fe07983d4f10f812fa7"
+  revision = "7b7533a7ca4eba7dd23dab2de70e25ca6eecf7e2"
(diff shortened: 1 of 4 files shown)


marcov:
done, thanks @jodh-intel

bergwolf:
@marcov We only require at least 1.10.4 (https://github.com/kata-containers/runtime/blob/master/versions.yaml#L267). Please update the runtime repo for the least required version.

bergwolf:
tag `dnm` until versions.yaml is updated in the runtime repo.

marcov:
@bergwolf sure I can do that

When building the rootfs, the agent's Go version is read from [a file in osbuilder](https://github.com/kata-containers/osbuilder/blob/master/rootfs-builder/versions.txt), so that should also be updated to use versions.yaml in the runtime repo. Actually I am not sure if the switch may break something.

jodh-intel:
@marcov - which switch do you mean - changing osbuilder to look at the runtime's `versions.yaml`?

/cc @jcvenegas.

marcov:
@jodh-intel correct, it would require checking the versions.yaml file from the runtime repo instead of a local file in osbuilder. And I am not sure if having osbuilder depend on runtime repo (or on something in that repo), may pose a problem.

jodh-intel:
I don't think it will cause problems, but then if not why did we create the separate `versions.txt`? Since osbuilder only needs that single file, we can probably do what the packaging repo does:

- https://github.com/kata-containers/packaging/blob/master/Makefile#L41

marcov:
In effect the golang binary is downloaded using `curl`, so that means that both `curl` and network access should be available.

bergwolf:
kata-containers/runtime#1693 is merged

Issue #549, opened 2019-05-06
Add a proper rollback path to addStorages()


YongjiXie:
Now we didn't decrement the sandbox storage reference counter when addStorages() failed in some case. This leads to resources leak. Even we might have no chance to mount ephemeral storage again if the first try failed.

PR #568, opened 2019-06-03
mount: Add a proper rollback path to addStorages()

ganeshmaharaj:
We need a proper rollback to decrement the sandbox storage
reference counter when addStorages() failed.

Fixes: #549

Signed-off-by: Xie Yongji <xieyongji@baidu.com>
(cherry picked from commit c66349b6c91156fbf59027e6c57f6aa211f365d0)
Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -317,12 +317,28 @@ func mountStorage(storage pb.Storage) error {
 // associated operations such as waiting for the device to show up, and mount
 // it to a specific location, according to the type of handler chosen, and for
 // each storage.
-func addStorages(ctx context.Context, storages []*pb.Storage, s *sandbox) ([]string, error) {
+func addStorages(ctx context.Context, storages []*pb.Storage, s *sandbox) (mounts []string, err error) {
 	span, ctx := trace(ctx, "mount", "addStorages")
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hmm.. do we need additional checks enabled for stable branch merges @chavafg, @grahamwhaley? This was showing as "green" with zero acks (!)

I'm going to leave this PR until we get one more ack...

grahamwhaley:
@jodh-intel - we do indeed only have the checks enabled for `master` branches, for all repos. The protection dialog *does* take a [branch name pattern match](https://help.github.com/en/articles/configuring-protected-branches) though, so we can fairly trivially change this to `*` or `stable*` for instance. Or, we can add two rules, so `master` requires two acks, but `stable*`s only require one for instance.

jodh-intel:
I think we should still have two acks for every branch. That protects us against scenarios such as where a user raises a PR to merge a change into a `stable*` branch *before* it has landed in `master`.

Thoughts @egernst, @gnawux?

PR #569, opened 2019-06-05
# Kata Containers 1.7.1

katabuilder:
- mount: Add a proper rollback path to addStorages()

ab9b2d0 mount: Add a proper rollback path to addStorages()


jcvenegas:
/ test ?

Issue #573, opened 2019-06-06
mount: virtioblk container rootfs mount for ACRN hypervisor  


vijaydhanraj:
For block devices added by acrn, virtpath is directly updated in storage.source and so no scanning of PCIAddr is required. This is because the virtio-blk device is not hot-plugged but added during VM launch and updated later with container rootfs using block rescan feature.

PR #574, opened 2019-06-07
mount: Virtio-blk container rootfs mount for ACRN hypervisor

vijaydhanraj:
For block devices added by acrn, virtpath is directly udpated
in storage.source and so no scanning of PCIAddr is required.
This is because, the virtio-blk device is not hot-plugged but
added during VM launch and updated later with container rootfs
using block rescan feature.

Fixes: #573

Signed-off-by: Vijay Dhanraj <vijay.dhanraj@intel.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -271,13 +271,28 @@ func virtioFSStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
 
 // virtioBlkStorageHandler handles the storage for blk driver.
 func virtioBlkStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
-	// Get the device node path based on the PCI address provided
-	// in Storage Source
-	devPath, err := getPCIDeviceName(s, storage.Source)
-	if err != nil {
-		return "", err
+
+	// If hot-plugged, get the device node path based on the PCI address else
+	// use the virt path provided in Storage Source
+	if strings.HasPrefix(storage.Source, "/dev") {
+
+		FileInfo, err := os.Stat(storage.Source)
+		if err != nil {
+			return "", err
+		}
+		// Make sure the virt path is valid
+		if FileInfo.Mode()&os.ModeDevice == 0 {
+			return "", err
+		}
+
+	} else {
+		devPath, err := getPCIDeviceName(s, storage.Source)
+		if err != nil {
+			return "", err
+		}
+
+		storage.Source = devPath
 	}
-	storage.Source = devPath
 
 	return commonStorageHandler(storage)
 }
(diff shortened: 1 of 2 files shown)


vijaydhanraj:
Related to: https://github.com/kata-containers/runtime/pull/1779

amshinde:
@vijaydhanraj Seeing this error:
```
mount.go:275: File is not `gofmt`-ed with `-s` (gofmt)
        // If hot-plugged, get the device node path based on the PCI address else
        // use the virt path provided in Storage Source
        if strings.Contains(storage.Source, "/dev") {
                // Make sure the virt path is valid
                _, err := os.Stat(storage.Source)
                if err != nil {
                        return "", err
                }
        } else {
                devPath, err := getPCIDeviceName(s, storage.Source)
                if err != nil {
                        return "", err
                }
```

Can you gofmt the `mount.go` file?

vijaydhanraj:
> @vijaydhanraj Seeing this error:
> 
> ```
> mount.go:275: File is not `gofmt`-ed with `-s` (gofmt)
>         // If hot-plugged, get the device node path based on the PCI address else
>         // use the virt path provided in Storage Source
>         if strings.Contains(storage.Source, "/dev") {
>                 // Make sure the virt path is valid
>                 _, err := os.Stat(storage.Source)
>                 if err != nil {
>                         return "", err
>                 }
>         } else {
>                 devPath, err := getPCIDeviceName(s, storage.Source)
>                 if err != nil {
>                         return "", err
>                 }
> ```
> 
> Can you gofmt the `mount.go` file?

Sure will do.  Thanks!

mount.go
```diff
@@ -271,13 +271,23 @@ func virtioFSStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
...
 func virtioBlkStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
-	// Get the device node path based on the PCI address provided
-	// in Storage Source
-	devPath, err := getPCIDeviceName(s, storage.Source)
-	if err != nil {
-		return "", err
+
+	// If hot-plugged, get the device node path based on the PCI address else
+	// use the virt path provided in Storage Source
+	if strings.Contains(storage.Source, "/dev") {
```
jodh-intel: Please can you add some unit tests for this code? This is a very weak match (not anchored) so would match things like `/home/developer/foo.txt`. Can you add anchoring reliably to guarantee you only match on actual devices?

mount.go
```diff
@@ -271,13 +271,23 @@ func virtioFSStorageHandler(storage pb.Storage, s *sandbox) (string, error) {
...
+		// Make sure the virt path is valid
+		_, err := os.Stat(storage.Source)
```
jodh-intel: If you are matching devices, can you use `os.Stat.Mode()` and `os.ModeDevice` to check you really got one?

vijaydhanraj: done

mount.go
vijaydhanraj: Using HasPrefix() now instead of Contains. Added a few negative test cases to check the flow.

PR #577, opened 2019-06-12
docs: Fix spelling

jodh-intel:
Correct a spelling mistake.

Fixes: #576.


jodh-intel:
@egernst - thanks! I was in fact testing the water here. Although `VCPU` may be "more correct", I think  `vCPU` is more common (and I too prefer it), so I've updated the PR.

PR #578, opened 2019-06-16
agent: mount blockdevices in the guest VM

zklei:
This mount blockdevices in the guest VM instead of passing them through

Fixes: #406

Signed-off-by: <leizhongkai@huawei.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -325,6 +325,13 @@ func mountStorage(storage pb.Storage) error {
 		return err
 	}
 
+	if (storage.Driver == driverSCSIType || storage.Driver == driverBlkType) && !strings.Contains(storage.Fstype, "bind") {
+		// if the "bind" option is not specified, it means that the blk device needs to be mounted into a directory
+		// here we temporarily discard the bind option,
+		// in order to be able to mount the file system of the block device.
+		flags = flags &^ flagList["bind"]
+	}
+
 	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
 }
 


mount.go
```diff
@@ -325,7 +325,17 @@ func mountStorage(storage pb.Storage) error {
...
 
-	return mount(storage.Source, storage.MountPoint, storage.Fstype, flags, options)
+	var fsType = storage.Fstype
+	if (storage.Driver == driverSCSIType || storage.Driver == driverBlkType) && !strings.Contains(storage.Fstype, "bind") {
+		// if the "bind" option is not specified, it means that the blk device needs to be mounted into a directory
+		// here we temporarily discard the bind option,
+		// in order to be able to mount the file system of the block device.
+		// and then reset `storage.Fstype` to "bind" which pass through to the libcontainer pkg.
+		flags = flags &^ flagList["bind"]
+		storage.Fstype = "bind"
```
bergwolf: IMO you can just construct proper storage structure (in kata runtime) to tell agent to mount the block device to a proper container path, rather than doing anything magical here. Did you see any blocker on that path?

zklei: Because I thought in the future, if `storage. Options`(i.e. flags) extends， we can reuse it and keep other mount options, so we just need to drop the "bind" option here.

zklei: Because I thought in the future, if `storage. Options`(i.e. flags) extends， we can reuse it and keep other mount options, so we just need to drop the "bind" option here.

egernst: /cc @amshinde WDYT here?

I tend to agree with @bergwolf

amshinde: Yeah I agree with @bergwolf , its better to cleanly implement this in the runtime itself passing in the right storage structures to the agent, rather than the agent making any assumptions here.

zklei: Alternatively, when the "enable-blk-mount" is turned on in the conf, the rt no longer needs to passes the "bind" opt.   
```
-           vol.Fstype = "bind"
-           vol.Options = []string{"bind"}
+         ac, ok := c.sandbox.config.AgentConfig.(KataAgentConfig)
+         if ok && ac.MountBlkInVM {
+                 // Ensure the block device is formatted, for the devices here are specified as volumes
+                 fsType, err := utils.GetDevFormat(m.Source)
+                 if err != nil || fsType == "" {
+                         k.Logger().WithField("device", id).WithError(err).Error("get device format failed")
+                         return nil
+                 }
+                 vol.Fstype = fsType
+         } else {
+                 vol.Fstype = "bind"
+                 vol.Options = []string{"bind"}
+         }
```
I've updated the runtime code so that agent doesn't need to make any changes, then this PR can be closed.

jodh-intel: For reference, the runtime PR is https://github.com/kata-containers/runtime/pull/882.

jodh-intel:
If this is now being done in the runtime, can we close this PR?

PR #585, opened 2019-06-18
agent: sandbox_pause should not take arguments

bergwolf:
gcc __attribute__ constructors cannot take arguments and cannot
access main function arguments on musl. Such functionality is glibc
specific and only works on x86 and x86_64. As a result, the pause
binary always quits on musl causing sandbox share pidns to malfunction.

### DIFF
--- a/pause.go
+++ b/pause.go
@@ -9,18 +9,36 @@ package main
 /*
 #cgo CFLAGS: -Wall
 #define _GNU_SOURCE
-#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <errno.h>
 
 #define PAUSE_BIN "pause-bin"
 
-void __attribute__((constructor)) sandbox_pause(int argc, const char **argv) {
-	if (argc != 2 || strcmp(argv[1], PAUSE_BIN)) {
-		return;
+void __attribute__((constructor)) sandbox_pause(void) {
+	FILE *f;
+	int len, do_pause = 0;
+	size_t n = 0;
+	char *p = NULL;
+
+	f = fopen("/proc/self/cmdline", "r");
+	if (f == NULL) {
+		perror("failed to open proc");
+		exit(-errno);
 	}
+	while ((len = getdelim(&p, &n, '\0', f)) != -1) {
+		if (len == sizeof(PAUSE_BIN) && !strncmp(p, PAUSE_BIN, sizeof(PAUSE_BIN)-1)) {
+			do_pause = 1;
+			break;
+		}
+	}
+	fclose(f);
+	free(p);
+
+	if (do_pause == 0)
+		return;
 
 	for (;;) pause();
 


devimc:
I faced a similar issue with other cgo implementations that I did in the past, the solution was to read the arguments from `/proc/self/cmdline`, @bergwolf what do you think?

bergwolf:
@devimc Good point! I've updated the PR to follow your suggestion. Thanks!

bergwolf:
Dropping `dnm` as the new fix does not introduce additional memory cost.

pause.go
```diff
@@ -14,18 +14,36 @@ package main
...
-	}
+void __attribute__((constructor)) sandbox_pause() {
+        char buf[512], *p;
+        int fd, len;
 
-	for (;;) pause();
+        fd = open("/proc/self/cmdline", O_RDONLY);
+        if (fd < 0) {
+                fprintf(stderr, "failed to open proc: %d\n", errno);
+                return;
```
liwei: IMO, we should exit here, otherwise, we got two kata-agent processes running in parallel.

pause.go
```diff
@@ -14,18 +14,36 @@ package main
...
+        }
+        len = read(fd, buf, sizeof(buf));
+        close(fd);
+        if (len <= strlen(PAUSE_BIN) || len == sizeof(buf)) {
+                fprintf(stderr, "failed to read buf, len %d\n", len);
+                return;
```
liwei: same here

pause.go
```diff
@@ -14,18 +14,36 @@ package main
...
+        }
+        p = buf + (len - strlen(PAUSE_BIN) - 1);
+        if (strcmp(p, PAUSE_BIN) != 0)
+                return;
```
liwei: same here

pause.go
devimc: I'd prefer to use `strerror(errno)` to be a more verbose

pause.go
```diff
@@ -14,18 +14,36 @@ package main
...
+        char buf[512], *p;
+        int fd, len;
 
-	for (;;) pause();
+        fd = open("/proc/self/cmdline", O_RDONLY);
+        if (fd < 0) {
+                fprintf(stderr, "failed to open proc: %d\n", errno);
+                return;
+        }
+        len = read(fd, buf, sizeof(buf));
```
devimc: this file should be read using `getdelim` since it has `0`'s as delimiters

```C
FILE *f = fopen("/proc/self/cmdline", "rb");
char *arg = 0;
....
while(getdelim(&arg, &size, 0, f) != -1)
{
   // do something with arg
}
free(arg);
```

@bergwolf what do you think?

pause.go
```diff
@@ -14,18 +14,36 @@ package main
...
+        close(fd);
+        if (len <= strlen(PAUSE_BIN) || len == sizeof(buf)) {
+                fprintf(stderr, "failed to read buf, len %d\n", len);
+                return;
+        }
+        p = buf + (len - strlen(PAUSE_BIN) - 1);
```
devimc: `getdelim` can help you here

pause.go
bergwolf: Good point! I'll update.

bergwolf:
PR updated. PTAL @liwei @devimc

pause.go
```diff
@@ -9,18 +9,36 @@ package main
...
+	size_t n = 0;
+	char *p = NULL;
+
+	f = fopen("/proc/self/cmdline", "r");
+	if (f == NULL) {
+		perror("failed to open proc");
+		exit(-errno);
+	}
+	while ((len = getdelim(&p, &n, 0, f)) != -1) {
+		if (len == strlen(PAUSE_BIN) + 1 && strcmp(p, PAUSE_BIN)) {
```
devimc: Two comments here:
* I don't follow, `strcmp` returns 0 if both string are equal, but here you want a non-zero value ?
* Can we use `sizeof` and `strncmp` ? below code should work


```C
if (len == sizeof(PAUSE_BIN) && strncmp(p, PAUSE_BIN, sizeof(PAUSE_BIN)-1) == 0) {
```

pause.go
```diff
@@ -9,18 +9,36 @@ package main
...
+			do_pause = 1;
+			break;
+		}
 	}
+	fclose(f);
+	free(p);
+
+	if (do_pause == 0)
```
devimc: please use braces

pause.go
amshinde: Yeah, this should probably be `! strcmp` here.

lifupan:
Since this pause process will be an init process in its pid namespace, does it need to do reap the exited orphan child processes?

pause.go
bergwolf: Good point! I've updated the PR. ptal.

pause.go
bergwolf: Not really necessary right? C code doesn't need to follow go coding style here.

bergwolf:
@lifupan That's handled by the agent reaper instead.

liwei:
I think the commit subject needs an update 😉

bergwolf:
Ouch! Updated. Thanks @liwei !

lifupan:
> @lifupan That's handled by the agent reaper instead.

Hi @bergwolf,  actually the kata-agent child subreaper cannot
reap a child from different pid namespace. Read from the latest kernel source codes https://elixir.bootlin.com/linux/latest/source/kernel/exit.c#L603, the logic to find a process's reaper is below:

1. give them to another thread in it's thread group, if such a member exists.
2. give it to the first ancestor process which prctl'd itself as a
child_subreaper for its children in the same pid namespace.
3. give it to the init process (PID 1) in our pid namespace
That's why kernel reparent the child process to container's (PID 1) process instead of the kata-agent process.

actually after kernel applying this patch,  https://lkml.org/lkml/2017/1/30/630 , the kata-agent cannot reap the orphan child processes in sub-namespace.

bergwolf:
@lifupan Then it's a bigger problem. Let me take a closer look.

liwei:
@lifupan @bergwolf Maybe we can fix the process reap issue in a separate PR and let this PR keeps going?

lifupan:
> @lifupan @bergwolf Maybe we can fix the process reap issue in a separate PR and let this PR keeps going?

Agree, we can open a separated issue to track it.

jodh-intel:
Odd. One of the Travis builds failed with a timeout... building a golang program!

```
$ go version
go version go1.11.11 linux/ppc64le

    :

go build -buildmode=pie -tags "" -o kata-agent \
	-ldflags "-X main.version=1.8.0-alpha2-f1e4e9d5eda55469d34d1b59f743a059bab8f01b -X main.seccompSupport=no "

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

Any thoughts @nitkon?

nitkon:
@jodh-intel: That was odd. I have restarted the Travis-CI and now it's green. `travis_wait` can be used to exceed the timeout limit but need to root-cause why it happens else it just would fail after the exceeded timeout.

jodh-intel:
Restarted initrd CI which had a transitory failure.

egernst:
flake test for initrd?  
```
Summarizing 1 Failure:

[Fail] state container [It] with workload [true], timeWait 5 
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/functional/state_test.go:45

Ran 3 of 8 Specs in 31.466 seconds
FAIL! -- 2 Passed | 1 Failed | 0 Pending | 5 Skipped --- FAIL: TestFunctional (31.47s)
FAIL
```

egernst:
@bergwolf - this should be backport candidate as well, correct?

The commit message seems to indicate the fix is not using args, whereas you are just grabbing the args now via proc sysfs, right?  Can you update commit message accordingly?

bergwolf:
@egernst Yes, I agree it should go to stable. tag added. Thanks.

egernst:
rebuilding ubuntu-18-04-initrd

devimc:
functional tests are failing again, nothing related to this pr, I think 

```
• Failure [21.258 seconds]
state
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/functional/state_test.go:26
  container
  /tmp/jenkins/workspace/kata-containers-agent-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    with workload [true], timeWait 5 [It]
    /tmp/jenkins/workspace/kata-containers-agent-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected
        <int>: 1
    to equal
        <int>: 0
```

PR #590, opened 2019-06-24
network: setupDNS for sandbox

nitkon:
In order to set hostname, dns etc of the VM, we
need to copy these files from host to VM rootfs
and not container rootfs.

Fixes:  #589

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -1441,7 +1441,7 @@ func (a *agentGRPC) CreateSandbox(ctx context.Context, req *pb.CreateSandboxRequ
 
 	a.sandbox.mounts = mountList
 
-	if err := setupDNS(a.sandbox.network.dns); err != nil {
+	if err := setupDNS(); err != nil {
 		return emptyResp, err
 	}
 
--- a/network.go
+++ b/network.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"reflect"
 	"sync"
+	"syscall"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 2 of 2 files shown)


grpc.go
```diff
@@ -1634,7 +1634,7 @@ func (a *agentGRPC) CopyFile(ctx context.Context, req *pb.CopyFileRequest) (*gpb
 	// container's rootfs is mounted at /run, in order to avoid overwrite guest's rootfs files, only
 	// is possible to copy files to /run
 	if !strings.HasPrefix(path, containersRootfsPath) {
-		return emptyResp, fmt.Errorf("Only is possible to copy files into the %s directory", containersRootfsPath)
+		agentLog.Info("Copying files to guest's rootfs")
```
jodh-intel: I think we'd better get @devimc's comments on this behavioral change.

grahamwhaley: coz, it feels like a potential big security hole to me... which I suspect is why we restrict the destination path current;y.
@nitkon ooi, why do you need to set the VM hostname etc.??
/cc @amshinde

amshinde: @nitkon Same questions as @grahamwhaley. 
Why do you need to set the VM hostname? Can you explain the use-case?

bergwolf: Can you copy file to `/run` and then bind mount it to `/etc/hostname` in `CreateSandbox` request if necessary?

nitkon: @bergwolf : Ptal ...  I have copied the file in `kataShared` on the host and then bind mounting it in the sandbox `/etc/resolv.conf` .

egernst:
@nitkon how does this relate to  #589 ?

PR #610, opened 2019-07-19
agent: delete element of sandbox.deviceWatchers with right key

### LINKED ISSUE #609: Problems found by code review
# Description of problem

`sandbox.deviceWatchers` is a map with the key `pciAddr`, but when receive the uEvent in the agent, the element of map is deleted with wrong key:

https://github.com/kata-containers/agent/blob/master/device.go#L152
`s.deviceWatchers[pciAddr] = notifyChan`

https://github.com/kata-containers/agent/blob/master/agent.go#739
`delete(s.deviceWatchers, uEv.DevName)`

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

The element want to be deleted will never be deleted.

zklei:
Fixes #609

Signed-off-by: leizhongkai <leizhongkai@huawei.com>

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -736,7 +736,7 @@ func (s *sandbox) listenToUdevEvents() {
 				if ch != nil && strings.HasPrefix(uEv.DevPath, filepath.Join(rootBusPath, devPCIAddress)) {
 					ch <- uEv.DevName
 					close(ch)
-					delete(s.deviceWatchers, uEv.DevName)
+					delete(s.deviceWatchers, devPCIAddress)
 				}
 			}
 


agent.go
```diff
@@ -736,7 +736,7 @@ func (s *sandbox) listenToUdevEvents() {
 				if ch != nil && strings.HasPrefix(uEv.DevPath, filepath.Join(rootBusPath, devPCIAddress)) {
 					ch <- uEv.DevName
 					close(ch)
-					delete(s.deviceWatchers, uEv.DevName)
+					delete(s.deviceWatchers, devPCIAddress)
```
lifupan: I think actually there is no need to do the delete here,  it seems https://github.com/kata-containers/agent/blob/581bcf2404b23b53d0f1fbd9b833f7559fe2baa9/device.go#L162 will do this delete.

zklei: yeah,  `agent/device.go#L162` deleted just when  time out happens.

devimc:
good catch @zklei

PR #547, opened 2019-05-02
agent: support debug console

devimc:
Start a debug console if the `agent.debug_console` option is part of the kernel
command line and a shell (bash or sh) is installed in the rootfs.
The debug console is useful when the developer wants to debug the agent process
or any other processes (even containers). The current way to do this is by
modifying a systemd target + creating a new system unit, the problem with this
approach is that it's not user-friendly and we depend on systemd, since it must
be installed in the rootfs.

fixes #546

Signed-off-by: Julio Montes <julio.montes@intel.com>

PR #616, opened 2019-07-30
agent: add support for loading kernel modules

### LINKED ISSUE #615: agent: add support for loading kernel modules

devimc:
Some containers may require HW devices to improve its performance. In Kata
containers these devices can be passed though VFIO and used in the container
only if the guest kernel has the drivers built-in for them. When kernel
fragment support landed (kata-containers/packaging#646), kata lost support
for several HW devices like Intel GPUs, this was expected since one of the
goals of this change was to have a minimal kernel in order to improve boot
time, memory footprint and reduce the attack surface by enabling just the
features required to run kata containers.
Currently if the users want to use a HW device in kata container, He/She has
to build a custom kernel and create a custom image that includes the modules
that will be loaded by systemd. From user's perspective this is a pain, since
not all users know how to create a custom image or configure/build a custom
kernel.
Same as the `debug_console` feature (kata-containers/agent#547), the goal of
this change is to facilitate the things to the final user and *do not* depend
on systemd.

fixes #615

see kata-containers/osbuilder#77


config.go
```diff
@@ -125,6 +127,8 @@ func (c *agentConfig) parseCmdlineOption(option string) error {
 			agentLog.Debug("Param passed to NOT use vsock channel")
 			commCh = serialCh
 		}
+	case kmodulesFlag:
+		kmodules = strings.Split(split[valuePosition], kmodulesSeparator)
```
bergwolf: What about adding the module list to `CreateSandboxRequest`?

bergwolf: Also we should add module options for each module. Then grpc is a better place so that we have structured parameters. sth. like:
```go
type KernelModule struct{
        Name string
        options []string
}

type CreateSandboxRequest struct {
<snip...>
        KernelModules []KernelModule
<snip...>
}
```

agent.go
```diff
@@ -1304,6 +1308,19 @@ func init() {
...
+func loadKernelModules() {
+	for _, m := range kmodules {
+		agentLog.WithField("module", m).Debug("loading module")
+		cmd := exec.Command(modprobePath, "-v", m)
+		output, err := cmd.CombinedOutput()
+		if err != nil {
+			agentLog.WithError(err).WithFields(logrus.Fields{
+				"output": string(output),
+				"module": m}).Error("could not load module")
+		}
```
bergwolf: We should fail if specified modules fail to load.

agent.go
```diff
@@ -1304,6 +1308,19 @@ func init() {
 	}
 }
 
+func loadKernelModules() {
```
jodh-intel: I agree with @bergwolf - this function should return an `error` (which will help us unit test this... :)

config.go
jodh-intel: Talking about gRPC, how about updating `agentGRPC.GetGuestDetails()` so that it optionally returns the contents of `/proc/modules`? That could be useful for debug given that loading "module a" might auto-load "modules b-e".

devimc: @jodh-intel @bergwolf  I agree, I like both ideas

amshinde:
@devimc You mentioned in the commit message that complications of a custom rootfs/kernel, but wouldnt the modules still need to be added to the rootfs?
I agree with @bergwolf's suggestion that we need to support options as well.
Lastly, would be useful to have this documented once we have this feature.

devimc:
@amshinde 

> @devimc You mentioned in the commit message that complications of a custom rootfs/kernel, but wouldnt the modules still need to be added to the rootfs?

If this patch is merged then we can move forward and may be include some modules in the image that we ship, or at least to provide an easy way to allow users create an image with the modules needed in the  container.

> I agree with @bergwolf's suggestion that we need to support options as well.
Lastly, would be useful to have this documented once we have this feature.

Yes, I agree

devimc:
@bergwolf @jodh-intel @amshinde pr updated, thanks

grpc.go
```diff
@@ -62,6 +62,8 @@ var (
 
 	// set when StopTracing() is called.
 	stopTracingCalled = false
+
+	modprobePath = "/sbin/modprobe"
```
ganeshmaharaj: Super dumb question. Is the path of `modprobe` the same for all distros? I know our images are today made with clear, but would this work if someone chooses to create their base image with a different distro?

bergwolf:
One more question, how do we determine which modules are to be loaded at the runtime side? I'm wondering if we need special grpc API to load modules, or just embed it in `CreateSandboxRequest`.

grpc.go
jodh-intel: Not a dumb question at all and good to ask it!

In summary, "yes" since `modprobe(8)` is an admin command, so the binary must live in `/sbin` (which might of course be linked to `/usr/sbin`). Generally, hard-coding paths should be discouraged, but in this scenario, it's required to ensure we're calling the correct "official" binary (security).

protocols/grpc/agent.proto
```diff
@@ -270,6 +270,13 @@ message TtyWinResizeRequest {
 	uint32 column = 4;
 }
 
+message KernelModule {
+	// This field is the name of the kernel module.
+	string name = 1;
+	// This field are the options for the kernel module.
```
jodh-intel: Nit: might be worth mentioning in the comment that the options/parameters are whitespace-delimited key=value pairs which are passed verbatim to `modprobe(8)`.

devimc:
@bergwolf 
> One more question, how do we determine which modules are to be loaded at the runtime side? 

I plan to add an option in the `configuration.toml` and a new annotation, what do you think ?

devimc:
/test 

@jodh-intel change applied, thanks

grpc.go
ganeshmaharaj: Thanks @jodh-intel

raravena80:
Can this be merged?

PR #622, opened 2019-08-06
network: Setup DNS for sandbox

nitkon:
Sandbox has network but not dns. Set
dns in the sandbox.

Fixes: #589

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>

### DIFF
--- a/network.go
+++ b/network.go
@@ -10,8 +10,12 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
+	"path/filepath"
 	"reflect"
+	"strings"
 	"sync"
+	"syscall"
 
 	"golang.org/x/sys/unix"
 
@@ -24,11 +28,13 @@ import (
 )
 
 var (
-	errNoHandle = grpcStatus.Errorf(codes.InvalidArgument, "Need network handle")
-	errNoIF     = grpcStatus.Errorf(codes.InvalidArgument, "Need network interface")
-	errNoLink   = grpcStatus.Errorf(codes.InvalidArgument, "Need network link")
-	errNoMAC    = grpcStatus.Errorf(codes.InvalidArgument, "Need hardware address")
-	errNoRoutes = grpcStatus.Errorf(codes.InvalidArgument, "Need network routes")
+	errNoHandle        = grpcStatus.Errorf(codes.InvalidArgument, "Need network handle")
+	errNoIF            = grpcStatus.Errorf(codes.InvalidArgument, "Need network interface")
+	errNoLink          = grpcStatus.Errorf(codes.InvalidArgument, "Need network link")
+	errNoMAC           = grpcStatus.Errorf(codes.InvalidArgument, "Need hardware address")
+	errNoRoutes        = grpcStatus.Errorf(codes.InvalidArgument, "Need network routes")
+	kataGuestSharedDir = "/run/kata-containers/shared/containers"
+	kataGuestFileDNS   = "/etc/resolv.conf"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


nitkon:
@amshinde @bergwolf : This is another approach wherein 
1. No changes to runtime are needed.
2. The resolv.conf and other files are not present at the time of CreateSandbox. The `kataGuestSharedDir = "/run/kata-containers/shared/containers"` is empty. Hence I have called `setupDNS` at the time of `CreateContainer` when those files are present. 

3. As CreateContainer gets called n+1 times where n is the number of containers, I have checked if resolv.conf is already mounted. 

4. In the runtime repo, at the time of CreateSandboxRequest, there are no files present in `/run/kata-containers/shared/sandboxes/sandbox-id`. Hence cannot send the info in `sandbox.network.dns`

Is this alternate approach acceptable?

grahamwhaley:
travis linter not happy with you @nitkon ;-)

```
/home/travis/gopath/src/github.com/kata-containers/agent/protocols/mockserver
network.go:37:2: var `kataGuestFileDns` should be `kataGuestFileDNS` (golint)
	kataGuestFileDns   = "/etc/resolv.conf"
	^
network.go:635:6: func getSandboxDnsFile should be getSandboxDNSFile (golint)
func getSandboxDnsFile(sandboxId string) (string, error) {
     ^
network.go:664:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (golint)
	} else {
	       ^
network.go:665:3: var `kataGuestSharedFileDns` should be `kataGuestSharedFileDNS` (golint)
		kataGuestSharedFileDns, err := getSandboxDnsFile(id)
		^
network.go:654:2: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (gosimple)
	if len(res) > 0 {
	^
The command ".ci/static-checks.sh" failed and exited with 1 during .
```

nitkon:
Thnx @grahamwhaley ;-) Fixed!

network.go
```diff
@@ -574,9 +580,91 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *types.Route, add
...
+		if strings.Contains(file, sandboxID) && strings.Contains(file, "resolv.conf") {
+			return file, nil
+		}
+	}
+
+	return "", nil
+}
+
+func Mounted(mountpoint string) bool {
+	res, _ := exec.Command("sh", "-c", "mount | grep /etc/resolv.conf").Output()
```
devimc: please don't use `sh`, `mount` and `grep`, instead read the file and parse it

network.go
```diff
@@ -574,9 +580,91 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *types.Route, add
...
+
+func getSandboxDNSFile(sandboxID string) (string, error) {
+
+	var files []string
+	err := filepath.Walk(kataGuestSharedDir, traverseFiles(&files))
+	if err != nil {
+		return "", err
+	}
+	for _, file := range files {
+		logrus.Printf("Files are %v", file)
```
devimc: use `WithField`

network.go
amshinde: Yes please try to use a system call instead of executing a binary.

amshinde:
@nitkon I see there are 3 PRs open to address the same issue.
You have mentioned the approach taken here, but can you describe other approaches for this that you have considered, the reasons for taking a particular approach,  so that we can compare and understand pros/cons.

network.go
```diff
@@ -574,9 +580,91 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *types.Route, add
...
+		}
+		*files = append(*files, path)
+		return nil
+	}
+}
+
+func getSandboxDNSFile(sandboxID string) (string, error) {
+
+	var files []string
+	err := filepath.Walk(kataGuestSharedDir, traverseFiles(&files))
```
amshinde: Are you doing a walkthrough of the entire file-system of the sandbox here ?
The /etc/resolve file should be present at a fixed path in the shared directory, there is no need to list out all the files and then search.

network.go
```diff
@@ -574,9 +580,91 @@ func (s *sandbox) updateRoute(netHandle *netlink.Handle, route *types.Route, add
 /////////
 // DNS //
 /////////
+func mountFile(source, destination, fsType string, flags int, options string) error {
```
amshinde: I believe there exists a function to perform bindmount already in the agent repo. Can you please reuse that?

Issue #623, opened 2019-08-06
Do not delete routes with proto as "kernel"


amshinde:
See: 
https://github.com/kata-containers/runtime/pull/1936
and
https://github.com/kata-containers/runtime/issues/1811

PR #624, opened 2019-08-06
network: While updating routes, do not delete routes with proto "kernel"

amshinde:
Routes with proto kernel are automatically added by the kernel
when an interface is added with ip assigned as a subnet address.
With github.com/kata-containers/runtime#1936, these routes will no
longer be sent from the runtime.

Depends-on: github.com/kata-containers/runtime#1936

Fixes: #623

### DIFF
--- a/network.go
+++ b/network.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"os"
 	"reflect"
+	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


amshinde:
PR updated to allow replacing routes.
@egernst PTAL.

PR #627, opened 2019-08-13
Remove certain log messages

amshinde:
logs: Remove certain logs that are repeatedly printed.
    
 When using static cpu policy, grpc request to update CPUs is
 being called over and over again. These logs create a lot of
 chatter in the journal as well make the debug console quite
 unusable. So get rid of these.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -23,7 +23,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/gogo/protobuf/proto"
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 	"github.com/kata-containers/agent/pkg/uevent"
 	pb "github.com/kata-containers/agent/protocols/grpc"
@@ -948,9 +947,6 @@ func (s *sandbox) initChannel() error {
 
 func makeUnaryInterceptor() grpc.UnaryServerInterceptor {
 	return func(origCtx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
-		var start time.Time
-		var elapsed time.Duration
-		var message proto.Message
 
 		grpcCall := info.FullMethod
 		var ctx context.Context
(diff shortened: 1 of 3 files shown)


amshinde:
@jcvenegas Please take a look at these and let me know if this is ok with you.

cc @egernst

agent.go
```diff
@@ -969,33 +965,13 @@ func makeUnaryInterceptor() grpc.UnaryServerInterceptor {
...
-		if !tracing {
-			// Just log call details
-			elapsed = time.Since(start)
-			message = resp.(proto.Message)
-
-			logger := agentLog.WithFields(logrus.Fields{
-				"request":  info.FullMethod,
-				"duration": elapsed.String(),
-				"resp":     message.String()})
-			logger.Debug("request end")
```
bergwolf: These are debug logs. Are you seeing it affecting CPUs in non-debug mode?

cgroup.go
```diff
@@ -48,11 +48,6 @@ func getAvailableCpusetList(cpusetReq string) (string, error) {
 
 	for k := range cpusetListReq {
 		if !cpusetGuestList[k] {
-			agentLog.WithFields(logrus.Fields{
```
grahamwhaley: I wonder if rather than dropping completely, we could wrap them in a golang [once](https://golang.org/pkg/sync/#Once) ? But:
- maybe seeing it only once makes no sense
- we'd be adding a small overhead for the once checking
- you might have to pull the log code out to a new func so we can wrap it, which is just niggling.

agent.go
amshinde: @bergwolf I am seeing these logs in the debug mode only.  For the static policy, I was using, I saw a lot of these messages being printed to the journal, which made it quite difficult to see any logs I was interested in.

amshinde:
> For debug console to be usable, IMO we'd better disassociate it with agent logger debug mode, instead of trying to remove these debug logs.

@bergwolf Let me see how we can do that for the console itself.

agent.go
jodh-intel: Currently we have two logging scenarios:

- minimal logging (default)
- "firehose mode" / full logging (debug mode).

I think we need to adopt a more granular approach and allow multiple debug levels. We could do this by adding a new `agent.debug_level=[0-9]` (or maybe just `agent.debug_level={low,medium,high}` as we really shouldn't need 10 levels of debug :smile:?) These messages could then only be displayed for the "full" or "high" level of debug logging. All debug logs should then have an `debug-level=XXX` field added to allow admins to filter of course.

The question is what we set `agent.debug_level=` to if the agent is only passed an `agent.log=debug` option (probably "full"/"high" as that's what users would expect I think).

jodh-intel:
@amshinde - any update on this?

jodh-intel:
Ping @amshinde.

amshinde:
@jodh-intel I am going to close this for now. Will open it again if I come up with a better approach.

Issue #659, opened 2019-10-10
remove the NewAgentClient()'s  parameter of 'enableYamux'


lifupan:
# Description of problem

For func NewAgentClient(), the parameter of 'enableYamux'
    has nothing to do with 'Proxybuiltin' and it can be figured
    out by the dialer's address scheme. Thus there is no need to
    pass the parameter 'enableYamux'.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #660, opened 2019-10-10
client: remove the parameter of 'enableYamux'

lifupan:
For func NewAgentClient(), the parameter of 'enableYamux'
has nothing to do with 'Proxybuiltin' and it can be figured
out by the dialer's address scheme. Thus there is no need to
pass the parameter 'enableYamux'.

Fixes: #659

Signed-off-by: lifupan <lifupan@gmail.com>


bergwolf:
Please add `Depends-on:` tag so that the PR can be properly tested by CI.

Issue #695, opened 2019-11-28
Rust agent(grpc-rs) cannot work with firecacker hybrid vsock


teawater:
https://github.com/kata-containers/kata-containers/issues/79

PR #696, opened 2019-11-28
client.go: HybridVSockDialer: Change Read EOT to recv peek

teawater:
Rust agent(grpc-rs) cannot work with firecacker because the first package
will be dropped or truncated by HybridVSockDialer read EOT code.
This issue was reported in
https://github.com/kata-containers/kata-containers/issues/79

Golang agent(grpc-golang) is also affected by this issue.  But it is first
package is 9 bytes FRAME_SETTINGS.  Drop it will not affect subsequent
communications.

This commit change to use recv peek that will do same check with read
EOT but not drop any package.

Fixes: #695

Signed-off-by: Hui Zhu <teawater@antfin.com>

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -13,12 +13,14 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 	"github.com/hashicorp/yamux"
 	"github.com/mdlayher/vsock"
 	opentracing "github.com/opentracing/opentracing-go"
+	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	grpcStatus "google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


protocols/client/client.go
```diff
@@ -407,12 +409,20 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
 			return nil, err
 		}
 
-		// Read EOT (End of transmission) byte
-		eot := make([]byte, 32)
-		if _, err = conn.Read(eot); err != nil {
-			// Just close the connection, gRPC will dial again
-			// without errors
-			conn.Close()
+		// Receive the package from the connection without remove this
```
jodh-intel: Nit: s/package/packet/.

Just an idea but you could simplify this comment to something like:

> Receive the packet from the connection without removing it from the receive queue (MSG_PEEK), ensuring the connection is usable.

teawater: @jodh-intel Thanks!  Pushed a new version according to your comments.

jodh-intel:
Thanks again @teawater.

Ping @kata-containers/agent.

devimc:
restarting firecracker CI

PR #698, opened 2019-11-30
client.go: HybridVSockDialer: Check return size n of unix.Recvfrom

teawater:
I found that unix.Recvfrom will not return err to io.EOF if it get
an EOF.  It just set return size n to 0.
That will make function commonDialer return. Grpc dialer will handle
it.

This commit add code to check return size n and set err to io.EOF if
n is 0.

Fixes: #697

Signed-off-by: Hui Zhu <teawater@antfin.com>

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -9,6 +9,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


teawater:
jenkins-ci-ubuntu-1804-firecracker fails because https://github.com/kata-containers/tests/issues/2136

bergwolf:
The failed CI `jenkins-ci-ubuntu-1804-firecracker` is unrelated and tracked at kata-containers/tests#2136

PR #706, opened 2019-12-27
FC: Add new vsock connection handshake

Pennyzct:
**Which feature do you think can be improved?**
New Firecracker `v0.20.0` has changed the host-initiated vsock connection protocol to include a trivial handshake.

The new protocol looks like this:
```
    - [host] CONNECT <port><LF>
    - [guest/success] OK <assigned_host_port><LF>
```
See PR [firecracker-microvm/firecracker#1472](https://github.com/firecracker-microvm/firecracker/pull/1472) for detailed info.

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -7,6 +7,7 @@
 package client
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"net"
@@ -407,12 +408,19 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
 			return nil, err
 		}
 
-		// Read EOT (End of transmission) byte
-		eot := make([]byte, 32)
-		if _, err = conn.Read(eot); err != nil {
-			// Just close the connection, gRPC will dial again
-			// without errors
+		// A trivial handshake is included in the host-initiated vsock connection protocol.
+		// It looks like this:
+		// - [host] CONNECT <port><LF>
+		// - [guest/success] OK <assigned_host_port><LF>
+		reader := bufio.NewReader(conn)
+		if response, err := reader.ReadString('\n'); err != nil {
 			conn.Close()
+			// for now, we temporarily rely on the backoff strategy from GRPC for more stable CI.
+			return conn, nil
+		} else if !strings.Contains(response, "OK") {
+			conn.Close()
+			// for now, we temporarily rely on the backoff strategy from GRPC for more stable CI.
+			return conn, nil
 		}
 
 		return conn, nil
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -20,6 +21,7 @@ import (
 	"github.com/hashicorp/yamux"
 	"github.com/mdlayher/vsock"
 	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	grpcStatus "google.golang.org/grpc/status"
@@ -37,6 +39,14 @@ var defaultDialTimeout = 15 * time.Second
 var defaultCloseTimeout = 5 * time.Second
 
 var hybridVSockPort uint32
+
+var agentClientFields = logrus.Fields{
+	"name":   "agent-client",
+	"pid":    os.Getpid(),
+	"source": "agent-client",
+}
+
+var agentClientLog = logrus.WithFields(agentClientFields)
 
 // AgentClient is an agent gRPC client connection wrapper for agentgrpc.AgentServiceClient
 type AgentClient struct {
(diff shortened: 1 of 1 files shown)


protocols/client/client.go
```diff
@@ -410,24 +409,17 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
+		// It looks like this:
+		// - [host] CONNECT <port><LF>
+		// - [guest/success] OK <assigned_host_port><LF>
+		reader := bufio.NewReader(conn)
+		if response, err := reader.ReadString('\n'); err != nil {
+			conn.Close()
+			return nil, err
+		} else if !strings.Contains(response, "OK") {
+			conn.Close()
+			return nil, errors.New("malformed response code")
```
jodh-intel: I wonder if it would make sense to log `response` here if debug is enabled?

Pennyzct: Hi~ @jodh-intel yep. And since this file is missing logging part, I've added another commit to fulfill it. ;).

Pennyzct:
Hi guys
FC CI is definitely failing here, since this change is included only in v0.20.0. ;). 
I've added another [PR](https://github.com/kata-containers/runtime/pull/2379) to do the FC version bump.

devimc:
@Pennyzct thanks, but please add `Depends-on: github.com/kata-containers/runtime#2379` in your commit message, that way we can see the CI using the latest version of firecracker

Pennyzct:
Hi~ @devimc 
FWIT, even if I add the `Depends-on: github.com/kata-containers/runtime#2379`, if we don't do the `runtime/` vendor update to include this commit, FC CI will still fail. ;).

devimc:
@Pennyzct thanks, now I can see that fc 0.20 is installed

```
[Hypervisor]
  MachineType = ""
  Version = "firecracker 0.20.0"
```

unfortunately FC CI is failing

```
Stderr: docker: Error response from daemon: OCI runtime create failed: Failed to check if grpc server is working: rpc error: code = DeadlineExceeded desc = context deadline exceeded: unknown.
```

Pennyzct:
**Updates:** 
The debugging part should be only running after already getting right response code, since otherwise we may encounter JSON decode error. 
```
dockerd[16053]: time="2020-01-10T03:34:14.461867307Z" level=error msg="Handler for POST 
/v1.38/containers/7a52caa16c60249ceb09ede8b65a8161a0ad7bfe6c1cc206040456ba9737541b/start 
returned error: OCI runtime create failed: unable to retrieve OCI runtime error (invalid character 'i' in 
literal true (expecting 'r')): /usr/local/bin/kata-runtime did not terminate sucessfully
```

devimc:
@Pennyzct good catch!

sboeuf:
LGTM

@jcvenegas @likebreath I think the right move here is to make sure you update Kata to support latest CH from master. This way, I think the random failures will go away, as the PR is effectively reverting the 3 patches we've been reverting so far (through our testing).

@devimc I think you need to bump FC to 0.20.0 so that it does not get broken when this PR is merged.

devimc:
@sboeuf what do you mean? CI installs firecracker 0.20 and is failing

```
• Failure [152.692 seconds]
docker kill
/tmp/jenkins/workspace/kata-containers-agent-ubuntu-1804-PR-firecracker/go/src/github.com/kata-containers/tests/integration/docker/kill_test.go:79
  killing container
  /tmp/jenkins/workspace/kata-containers-agent-ubuntu-1804-PR-firecracker/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    with '22'(stopped (tty output)) signal [It]
    /tmp/jenkins/workspace/kata-containers-agent-ubuntu-1804-PR-firecracker/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:43

    Expected error:
        <*errors.errorString | 0xc000057150>: {
            s: "Timeout reached after 120s",
        }
        Timeout reached after 120s
    not to have occurred

    /tmp/jenkins/workspace/kata-containers-agent-ubuntu-1804-PR-firecracker/go/src/github.com/kata-containers/tests/integration/docker/kill_test.go:144
```

sboeuf:
@Pennyzct the PR is failing with latest FC 0.20. Are you still trying to make this pass?

protocols/client/client.go
```diff
@@ -410,25 +419,20 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
-			}
+		// A trivial handshake is included in the host-initiated vsock connection protocol.
+		// It looks like this:
+		// - [host] CONNECT <port><LF>
+		// - [guest/success] OK <assigned_host_port><LF>
+		reader := bufio.NewReader(conn)
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			conn.Close()
+			return nil, err
```
sboeuf: Try removing this line, this way the function will return `conn, nil` and the GRPC code will handle dialing again.

protocols/client/client.go
```diff
@@ -410,25 +419,20 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
+		} else if !strings.Contains(response, "OK") {
+			conn.Close()
+			return nil, errors.New("malformed response code")
```
sboeuf: Try removing this line, this way the function will return `conn, nil` and the GRPC code will handle dialing again.

amshinde:
@Pennyzct Can you implement @sboeuf suggestions and see if the CI for firecracker passes with those.

Pennyzct:
Hi~ @sboeuf @amshinde 
Sorry for the late response. 😢. I'm still on the Lunar New Year Holiday, so the response could be slow.;).
I'm still working on these patches, it's odd, the all docker-related integration tests have all passed, but failing on `ctr`. I'll try to reproduce and debug it.

protocols/client/client.go
Pennyzct: I'm confused, if reading was failing, we should return error instantly. Then here in func [commonDialer](https://github.com/kata-containers/agent/blob/master/protocols/client/client.go#L330)
```
	go func() {
		for {
			select {
			case <-cancel:
				// canceled or channel closed
				return
			default:
			}

			conn, err := dialFunc()
			if err == nil {
				// Send conn back iff timer is not fired
				// Otherwise there might be no one left reading it
				if t.Stop() {
					ch <- conn
				} else {
					conn.Close()
				}
				return
			}
		}
	}()
```
we could re-connect(`conn, err := dialFunc()`) again.

sboeuf: Yes I agree, but this quick re-connection performed by `commonDialer()` tend to get things to fail. Instead, when relying on the backoff strategy from GRPC (which happens when we return `conn, nil`), things are more stable.
Just to clarify, I'm not saying this is a long term solution, but while we look actively for the root cause, relying on the backoff strategy from GRPC gives us a more stable CI for both Firecracker and Cloud-Hypervisor.

jcvenegas:
@Pennyzct ping could you rebase this PR?

Pennyzct:
Hi~ guys
Sorry for the delay.😢. I'm still on the Lunar New Year holidays, and the response could be slow.
Thanks for the suggestion from @sboeuf, I've refined the code to use the back-off strategy from grpc. And I've also re-based the code here. ;).

PR #717, opened 2020-01-21
protocols: Fix client hybrid vsock protocol

### LINKED ISSUE #716: Stabilize Cloud-Hypervisor CI
The CI for Cloud-Hypervisor is not really stable since the hybrid vsock connection needs some fix.

sboeuf:
The previous patches related to the hybrid vsock implementation of the
GRPC dialer broke the case where the agent is not connected to the vsock
socket when the runtime tries to initiate the connection.

This commit restores the proper functionality as it does not close the
connection when recvfrom() syscall returns 0. Additionally, it does not
return an error when the syscall fails, as this lets the GRPC handle the
retries on its own.

Fixes #716

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -9,19 +9,16 @@ package client
 import (
 	"context"
 	"fmt"
-	"io"
 	"net"
 	"net/url"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 	"github.com/hashicorp/yamux"
 	"github.com/mdlayher/vsock"
 	opentracing "github.com/opentracing/opentracing-go"
-	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	grpcStatus "google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


amshinde:
@sboeuf Patch looks good, thought the cloud hypervisor CI still seems to be failing.

sboeuf:
@amshinde yes unfortunately it seems that's not the fix. The same patch has been passing the CI, but now it fails :(

sboeuf:
@chavafg @amshinde because this is a change related to `protocols/client/client.go` (used by the runtime only), are we sure the CI compiles the runtime based on this new version?

teawater:
https://github.com/kata-containers/agent/pull/706
This is the PR to fix this issue from @Pennyzct 
I think this one is better because it check the reply from fc.

sboeuf:
@teawater @jodh-intel @amshinde @jcvenegas @likebreath @egernst 
>#706
This is the PR to fix this issue from @Pennyzct
I think this one is better because it check the reply from fc.

I agree the PR from @Pennyzct is the right way to go in the long term. But please notice that both Firecracker and Cloud-Hypervisor must be updated to their latest version before this can work. The new handshake introduced by the hybrid vsock solution is not backward compatible, hence we should decide what needs to be bumped first.

If we merge @Pennyzct's PR, then someone has to work on making sure that CH and FC are updated and working with this new code.

IMO, it's good to merge @Pennyzct's PR since we want to move to latest FC and CH anyway, so let's not take some detours.

PR #733, opened 2020-02-04
protocols: Fix Cloud-Hypervisor CI regarding hvsock

### LINKED ISSUE #732: Cloud-Hypervisor sporadically fails
Some failures related to Cloud-Hypervisor have been reported [here](https://github.com/kata-containers/tests/issues/2141#issuecomment-581974147).

sboeuf:
This PR fixes Cloud-Hypervisor CI by increasing the connection retry timeout from 15s to 60s as it can take some time for the agent to listen onto the vosck socket when running in nested virtualization environment.

Because of this fix, the PR also re-enable the proper error handling, leading to a faster connection retry, instead of relying on the GRPC backoff strategy.

Fixes #732

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -9,6 +9,7 @@ package client
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -427,13 +428,11 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
 		if err != nil {
 			conn.Close()
 			agentClientLog.WithField("Error", err).Debug("HybridVsock trivial handshake failed")
-			// for now, we temporarily rely on the backoff strategy from GRPC for more stable CI.
-			return conn, nil
+			return nil, err
 		} else if !strings.Contains(response, "OK") {
 			conn.Close()
 			agentClientLog.WithField("response", response).Debug("HybridVsock trivial handshake failed with malformd response code")
-			// for now, we temporarily rely on the backoff strategy from GRPC for more stable CI.
-			return conn, nil
+			return nil, errors.New("hvsock response does not contain \"OK\"")
 		}
 		agentClientLog.WithField("response", response).Debug("HybridVsock trivial handshake")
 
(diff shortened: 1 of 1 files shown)


jcvenegas:
What would happen if we reduce the retry timeout ? just to try to be more agressive and reduce the time frame to connect ?  like simulate what grpc does?  I dont know if make sense what I ask

sboeuf:
@jcvenegas 
>What would happen if we reduce the retry timeout ? just to try to be more agressive and reduce the time frame to connect ? like simulate what grpc does? I dont know if make sense what I ask

The number of times you retry during the 15s interval does not matter. There are some cases where the agent takes 15s to properly listen onto the vsock socket (and I think this is due to nested virt).
Now, because we now have the problem fixed, I think it's important to go back to the aggressive retry strategy, so that we don't penalize the common use case where we want to connect as fast as possible to the agent.

amshinde:
@sboeuf @jcvenegas CI still looks unhappy. 
@sboeuf Have we measured how long it takes for clh to boot in a nested environment and for the vsock connection from the agent to be available?

sboeuf:
@amshinde 
>@sboeuf @jcvenegas CI still looks unhappy.

That's expected since this patch effectively does nothing from an agent perspective. The change affects `protocols/client/client.go`, which is used from the runtime.

>@sboeuf Have we measured how long it takes for clh to boot in a nested environment and for the vsock connection from the agent to be available?

No that's something we're planning to look at.

sboeuf:
@jcvenegas you might need to add a `Depends-on` to get the CI passing.

jcvenegas:
@sboeuf in this case depends on does not work as needs to update the vendor not just clone an specific code, but I have a branch testing all the changes we have pending 

https://github.com/kata-containers/runtime/pull/2440

CI looks statable

@amshinde  talking with @sboeuf this may not be the most elegant solution, next step is to do performance testing  on clh as this may be related with the slow container boot times we have using cloud hypervisor

amshinde:
> @amshinde talking with @sboeuf this may not be the most elegant solution, next step is to do performance testing on clh as this may be related with the slow container boot times we have using cloud hypervisor

@jcvenegas Sounds good. Would be good to have this data for firecracker as well and see if there are any discrepancies there as they have the same communication mechanism.

PR #734, opened 2020-02-13
protocols: client: Add timeout for hybrid vsock handshake

jcvenegas:
When the client tries to connect sometimes a race condition
could happen when the between bind and listen calls in the agent
vsock side.

This will block the hypervisor wanting for a response and
as consequence the agent client where it checks for an OK
response.

This case needs to be handled by the guest kernel, see
https://lore.kernel.org/netdev/668b0eda8823564cd604b1663dc53fbaece0cd4e.camel@intel.com/

As an extra protection make the agent client timeout if no OK response
is given. The response should be quick so is OK to wait a few seconds
and then timeout.

This also allow to return an error from the dialler function so retry
does not fallback on grpc retry making retries faster.

Fixes: #372

### DIFF
--- a/protocols/client/client.go
+++ b/protocols/client/client.go
@@ -9,6 +9,7 @@ package client
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -400,6 +401,7 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
 	}
 
 	dialFunc := func() (net.Conn, error) {
+		handshakeTimeout := 10 * time.Second
 		conn, err := net.DialTimeout("unix", udsPath, timeout)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


jcvenegas:
https://github.com/kata-containers/runtime/pull/2454 test of code in runtime

protocols/client/client.go
```diff
@@ -418,26 +420,39 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
+		case err = <-errChan:
 			conn.Close()
 			agentClientLog.WithField("Error", err).Debug("HybridVsock trivial handshake failed")
-			// for now, we temporarily rely on the backoff strategy from GRPC for more stable CI.
-			return conn, nil
-		} else if !strings.Contains(response, "OK") {
+			return nil, err
+		case <-time.After(handshakeTimeout):
+			// Timeout: kernel vsock implementation has a race condition, where no response is given
+			// Insted of wait forever for a response, timeout after a fair amount of time.
```
sboeuf:
```suggestion
			// Instead of waiting forever for a response, timeout after a fair amount of time.
```

protocols/client/client.go
```diff
@@ -418,26 +420,39 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
-		response, err := reader.ReadString('\n')
-		if err != nil {
+		successfulConnectionChan := make(chan bool)
+		errChan := make(chan error)
+
+		go func() {
+			reader := bufio.NewReader(conn)
+			response, err := reader.ReadString('\n')
+			if err != nil {
+				errChan <- err
```
sboeuf:
```suggestion
				errChan <- err
				return
```
We want to make sure the go routine terminates after the error happened.

protocols/client/client.go
```diff
@@ -400,6 +401,7 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
 	}
 
 	dialFunc := func() (net.Conn, error) {
+		handshakeTimeout := 3 * time.Second
```
sboeuf: I think this is not enough as it might introduce some potential errors if for some use cases (nested environment) it was really taking some time to get the handshake properly done.
I think 10s would give more slack.

protocols/client/client.go
```diff
@@ -418,26 +420,39 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
+			if err != nil {
+				errChan <- err
+			}
+
+			agentClientLog.WithField("response", response).Debug("HybridVsock trivial handshake")
+
+			if strings.Contains(response, "OK") {
+				successfulConnectionChan <- true
+			} else {
+				errChan <- errors.New("HybridVsock trivial handshake failed with malformd response code")
```
chavafg: typo: malformed?

jcvenegas:
@sboeuf @chavafg @amshinde @devimc @jodh-intel @grahamwhaley this PR is ready folks, please take a look

chavafg:
restarted CI jobs that failed. 
The fedora one failed with 
```
Stderr: Failed to check if grpc server is working: rpc error: code = Unavailable desc = transport is closing
```

protocols/client/client.go
```diff
@@ -418,26 +420,40 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
+			response, err := reader.ReadString('\n')
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			agentClientLog.WithField("response", response).Debug("HybridVsock trivial handshake")
+
+			if strings.Contains(response, "OK") {
+				successfulConnectionChan <- true
```
devimc: can we reuse the `errChan` ?  `errChan <- nil`

protocols/client/client.go
```diff
@@ -418,26 +420,40 @@ func HybridVSockDialer(sock string, timeout time.Duration) (net.Conn, error) {
...
+			} else {
+				errChan <- errors.New("HybridVsock trivial handshake failed with malformed response code")
+			}
+		}()
+
+		select {
+		case <-successfulConnectionChan:
+			return conn, nil
+		case err = <-errChan:
 			conn.Close()
```
devimc: if `errChan` can be reused, here we need to check `if err != nil` , right?

jcvenegas:
https://github.com/kata-containers/runtime/pull/2454 CI passing with latest changes, @devimc  PR updated

Issue #735, opened 2020-02-14
Add support for configuring nvidia GPU access for container


RyanAoh:
### some related issues ：
https://github.com/kata-containers/runtime/issues/2269
https://github.com/kata-containers/agent/issues/709
https://github.com/kata-containers/documentation/issues/517
https://github.com/kata-containers/runtime/issues/1876

There are two questions：
1. When to load nvidia-kernel module
Nvidia kernel module can't be loaded when there is no device to drive, so , it can't be loaded using kata configuration file  or systemd scripts. We need to distinguish the container  which has gpu vfio device invoked  with '--device' , and load nvidia kernel modules at the right time(after devices added).
2. How to share the driver files（libs, bins）at kata-vm（container‘s host） to container
Reference [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-runtime) ：“ A modified version of runc adding a custom pre-start hook to all containers.If environment variable NVIDIA_VISIBLE_DEVICES is set in the OCI spec, the hook will configure GPU access for the container by leveraging nvidia-container-cli from project libnvidia-container”.
We need to do the same thing to configure GPU access for the container. Because we can't let user  modify Dockerfile to install driver files every time when a container is started, it's  inefficient and complicated for users. 

### Solution：
1. Like nvidia-docker， set env “NVIDIA_VISIBLE_DEVICES”（in device-plugin, Dockerfile, or command args）to distinguish the gpu container, then call function to setup nvidia driver for the container.
2. setup nvidia driver for container :
  + exec "[nvidia-container-cli](https://github.com/NVIDIA/libnvidia-container) --kernel-modules info " to load nvidia kernel modules and get gpus info
  + exec "nvidia-container-cli list --binaries","nvidia-container-cli list --libraries" list the neccessary driver files
  + mount the binaries and libraries listed before to the container's rootfs
  + exec "ldconfig" to configure dynamic linker run-time bindings
> nvidia-container-cli and nvidia kernel modules should be instlled to kata-vm when building the rootfs.

#### Test result
Actually, I have done some tests in the way above，it can achieve the same effect as nvidia-docker.

+ Docker:
![image](https://user-images.githubusercontent.com/13352465/74536749-a7024a80-4f73-11ea-9916-4dbe8ca541dc.png)
![image](https://user-images.githubusercontent.com/13352465/74536860-e16be780-4f73-11ea-90dc-c89fb63e4092.png)

+ kubernetes(with my own device-plugin)
[...]

PR #736, opened 2020-02-16
agent: configure nvidia GPU access for the container

RyanAoh:
With nvidia-container-cli and nvidia driver installed into the rootfs of kata vm, if environment variable NVIDIA_VISIBLE_DEVICES is set in the OCI spec, the agent will configure GPU access for the container by leveraging nvidia-container-cli from project libnvidia-container.

Fixes: #735

Signed-off-by: Aohan Yang <aohan@nuaa.edu.cn>

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -668,6 +668,16 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	for _, e := range ociSpec.Process.Env {
+		if strings.HasPrefix(e, "NVIDIA_VISIBLE_DEVICES=") {
+			err := setupNvidiaDriver(ociSpec.Root.Path)
+			if err != nil {
+				return emptyResp, err
+			}
+			break
+		}
+	}
+
 	if a.sandbox.guestHooksPresent {
 		// Add any custom OCI hooks to the spec
 		a.sandbox.addGuestHooks(ociSpec)
--- a/nvidia.go
+++ b/nvidia.go
@@ -1,0 +1,116 @@
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"golang.org/x/sys/unix"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+const ldconfPath = "/etc/ld.so.conf"
+
+func setupNvidiaDriver(containerRootfs string) (err error) {
+	bins, libs, err := getNvidiaFiles()
+	if err != nil {
+		return
+	}
+
+	// mount driver binaries
+	for _, binFile := range bins {
+		if err = mount(binFile, filepath.Join(containerRootfs, binFile), "bind", unix.MS_BIND|unix.MS_REC|unix.MS_RDONLY, "rbind"); err != nil {
+			return
+		}
+	}
+
+	// mount driver libraries
+	libDirsMap := make(map[string]interface{})
+	for _, libFile := range libs {
+		if err = mount(libFile, filepath.Join(containerRootfs, libFile), "bind", unix.MS_BIND|unix.MS_REC|unix.MS_RDONLY, "rbind"); err != nil {
+			return
+		}
+		libDir := filepath.Dir(libFile)
+		libDirsMap[libDir] = nil
+	}
+
+	// configure dynamic linker run-time bindings
+	ldArgs := []string{"-r", containerRootfs}
+	for dir := range libDirsMap {
+		ldArgs = append(ldArgs, filepath.Join(containerRootfs, dir))
+	}
+	ldCmd := exec.Command("ldconfig", ldArgs...)
+	err = ldCmd.Run()
+	if err != nil {
+		return
+	}
+	// write libdirs to ld.so.conf
+	confFile, err := os.OpenFile(filepath.Join(containerRootfs, ldconfPath), os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
+	if err != nil {
+		return
+	}
+	defer confFile.Close()
+	for dir := range libDirsMap {
+		if _, err = fmt.Fprintln(confFile, dir); err != nil {
+			return
+		}
+	}
+
+	return nil
+}
+
+func getNvidiaFiles() (bins []string, libs []string, err error) {
+	// "--load-kmods": Load kernel modules
+	// "info": Report information about the driver and devices
+	infoCmd := exec.Command("nvidia-container-cli", "--load-kmods", "info")
+	if err = infoCmd.Run(); err != nil {
+		return
+	}
+
+	// List driver binaries
+	lbinCmd := exec.Command("nvidia-container-cli", "list", "--binaries")
+	if bins, err = runGetOutputLines(lbinCmd); err != nil {
+		return
+	}
+
+	//  List driver libraries
+	llibCmd := exec.Command("nvidia-container-cli", "list", "--libraries")
+	if libs, err = runGetOutputLines(llibCmd); err != nil {
+		return
+	}
+
+	return
+}
+
+func runGetOutputLines(cmd *exec.Cmd) (outputLines []string, err error) {
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return
+	}
+	if err = cmd.Start(); err != nil {
+		return
+	}
+
+	outputBuf := bufio.NewReader(stdout)
+	var line []byte
+	for {
+		line, _, err = outputBuf.ReadLine()
+		if err != nil {
+			if err.Error() == "EOF" {
+				break
+			}
+			return
+		}
+		outputLines = append(outputLines, string(line))
+	}
+
+	if err = cmd.Wait(); err != nil {
+		return
+	}
+
+	return
+}
--- a/nvidia_test.go
+++ b/nvidia_test.go
@@ -1,0 +1,42 @@
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"os/exec"
+	"reflect"
+	"testing"
+)
+
+func TestRunGetOutputLines(t *testing.T) {
+	type args struct {
+		cmd *exec.Cmd
+	}
+	tests := []struct {
+		name            string
+		args            args
+		wantOutputLines []string
+		wantErr         bool
+	}{
+		{
+			name:            "successfulCase",
+			args:            args{cmd: exec.Command("echo", "-e", "abc\ndef")},
+			wantOutputLines: []string{"abc", "def"},
+			wantErr:         false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotOutputLines, err := runGetOutputLines(tt.args.cmd)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("runGetOutputLines() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(gotOutputLines, tt.wantOutputLines) {
+				t.Errorf("runGetOutputLines() gotOutputLines = %v, want %v", gotOutputLines, tt.wantOutputLines)
+			}
+		})
+	}
+}


grahamwhaley:
Nice. It's quite a hardware specific change to the agent, but, I'm not sure we have any other alternative mechanism (like plugins) right now...

Let's see if we can track down some NVIDIA folks to pull in - maybe @guptaNswati  and @RenaudWasTaken ??

grpc.go
```diff
@@ -668,6 +668,16 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	for _, e := range ociSpec.Process.Env {
+		if strings.Contains(e, "NVIDIA_VISIBLE_DEVICES") {
```
devimc: then `foo=NVIDIA_VISIBLE_DEVICES` and  `vNVIDIA_VISIBLE_DEVICESar="foo" ` are valid? 
how about using `strings.HasPrefix(e, "NVIDIA_VISIBLE_DEVICES=")` ?

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}
+
+	return nil
+}
+
+func getNvidiaFiles() (bins []string, libs []string, err error) {
+	// "-k": Load kernel modules
+	// "info": Report information about the driver and devices
+	infoCmd := exec.Command("nvidia-container-cli", "--load-kmods", "info")
```
devimc: missing `-k` option? you mention it in the comments

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+
+func setupNvidiaDriver(containerRootfs string) (err error) {
+	bins, libs, err := getNvidiaFiles()
+	if err != nil {
+		return
+	}
+
+	// mount driver binaries
+	for _, binFile := range bins {
+		if err = mount(binFile, containerRootfs+binFile, "bind", unix.MS_BIND|unix.MS_REC, "rbind"); err != nil {
```
devimc: how about using `filepath.Join(containerRootfs, binFile)`?  
security concern: have you tried to modify the binary form the container? I'd prefer to copy the binary

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+			return
+		}
+	}
+
+	// mount driver libraries
+	libDirsMap := make(map[string]interface{})
+	for _, libFile := range libs {
+		if err = mount(libFile, containerRootfs+libFile, "bind", unix.MS_BIND|unix.MS_REC, "rbind"); err != nil {
```
devimc: same questions

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+	for dir := range libDirsMap {
+		ldArgs = append(ldArgs, containerRootfs+dir)
+	}
+	ldCmd := exec.Command("ldconfig", ldArgs...)
+	err = ldCmd.Run()
+	if err != nil {
+		return
+	}
+	// write libdirs to ld.so.conf
+	confFile, err := os.OpenFile(containerRootfs+ldconfPath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
```
devimc: `filepath.Join`

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// mount driver binaries
+	for _, binFile := range bins {
+		if err = mount(binFile, containerRootfs+binFile, "bind", unix.MS_BIND|unix.MS_REC, "rbind"); err != nil {
+			return
+		}
+	}
+
+	// mount driver libraries
+	libDirsMap := make(map[string]interface{})
```
devimc: are you using a map to skip duplicates? if so, you should check if the path already exist before bind-mount it again (next lines)

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+		libDir := filepath.Dir(libFile)
+		libDirsMap[libDir] = nil
+	}
+
+	// configure dynamic linker run-time bindings
+	ldArgs := []string{"-r", containerRootfs}
+	for dir := range libDirsMap {
+		ldArgs = append(ldArgs, containerRootfs+dir)
+	}
+	ldCmd := exec.Command("ldconfig", ldArgs...)
```
devimc: I'm curious, why the agent should be aware of  this new libraries?

grpc.go
RyanAoh: yes, you are right. I'll fix it.

nvidia.go
RyanAoh: '-k' is just '--load-kmods', I have changed the code but forgot to change the comment.  I'll fix it.

nvidia.go
RyanAoh: To fix

nvidia.go
RyanAoh: To fix.(filepath.Join)
As for copying or mounting, I  have tested to modify the binary form the container, it can't be modified because it's read only( as the vm is read only file system). And the binaries are also isolated by kata vm. Change in one container will not affect other containers. So I used mounting.

nvidia.go
RyanAoh: I use a map to skip duplicate lib dirs but not lib files. These files are not duplicated. The mount function also checks if the path already exists.
"nvidia-container-cli list -l" output like follow:
```
/usr/lib64/libnvidia-ml.so.418.116.00
/usr/lib64/libnvidia-cfg.so.418.116.00
/usr/lib64/vdpau/libvdpau_nvidia.so.418.116.00
/usr/lib64/libnvidia-encode.so.418.116.00
...
```
mount every file
Dirs "/usr/lib64/" "/usr/lib64/vdpau/" are saved to use as args of ldconfig and write into "/etc/ld.so.conf"

nvidia.go
RyanAoh: ldconfig comand as follow:
`ldconfig -r  path/to/rootfs  path/to/rootfs/usr/lib64  path/to/rootfs/usr/lib64/vdpau ...(lib dirs)`
Use -r to change to and use containers's rootfs as root directory.
Specified the directories(`filepath.Join(containerRootfs, dir`)  on the command to be processed.

RenaudWasTaken:
Hello!

A note on devices generally, we are working on CDI (https://github.com/RenaudWasTaken/cdi) with a few other runtime maintainers so that runtimes don't have to write vendor specific code. It seems that this might be useful here :)

Outside of that, on the actual PR, it seems like you might just want to run the nvidia-container-cli, or more likely the nvidia-container-toolkit, as a prestart hook since:
* All the work mounting libraries, devices and executable is already handled by that binary
* We do a bit more work around "preflight checks" ensuring that the container can actually run on that GPU
* You'll get the latest features container features
* It will give you a more consistent behavior with Docker / Podman / LXC

RyanAoh:
> Outside of that (since CDI is not finished), on the actual PR, it seems like you might just want to run the nvidia-container-cli, or more likely the nvidia-container-toolkit, as a prestart hook since:
> 

I have tried to add nvidia-container-runtime-hook as a prestart hook just like what [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-runtime) did, but the container failed. As I am not really know how nvidia-container-runtime-hook works，I can‘t work out the error. At last, I took the approach in the PR.

grpc.go
```diff
@@ -668,6 +668,16 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		return emptyResp, err
 	}
 
+	for _, e := range ociSpec.Process.Env {
+		if strings.HasPrefix(e, "NVIDIA_VISIBLE_DEVICES") {
```
devimc: I think you should be a little bit more specific here,  how about using `"NVIDIA_VISIBLE_DEVICES="`?

nvidia.go
```diff
@@ -0,0 +1,116 @@
+//
```
devimc: not sure, but I think this header should have a copyright, maybe  `Copyright (c) 2020 Aohan Young` .. ?

cc @grahamwhaley

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+
+func setupNvidiaDriver(containerRootfs string) (err error) {
+	bins, libs, err := getNvidiaFiles()
+	if err != nil {
+		return
+	}
+
+	// mount driver binaries
+	for _, binFile := range bins {
+		if err = mount(binFile, filepath.Join(containerRootfs, binFile), "bind", unix.MS_BIND|unix.MS_REC, "rbind"); err != nil {
```
devimc: just to make sure the binaries can't be modified (in case a user adds `rw` in the kernel command line), could you add `unix.MS_RDONLY` and check if this still works for you?

nvidia.go
```diff
@@ -0,0 +1,116 @@
...
+			return
+		}
+	}
+
+	// mount driver libraries
+	libDirsMap := make(map[string]interface{})
+	for _, libFile := range libs {
+		if err = mount(libFile, filepath.Join(containerRootfs, libFile), "bind", unix.MS_BIND|unix.MS_REC, "rbind"); err != nil {
```
devimc: same here

RenaudWasTaken:
> Thanks @RenaudWasTaken. Actually, I have tried to add nvidia-container-runtime-hook as a prestart hook just like what nvidia-container-runtime did at first, but the container failed. 

Happy to help this work! Do you have any error logs ?

RyanAoh:
@RenaudWasTaken 
Here are the logs collected by running kata-collect-data.sh. Kata runtime and agent debug enabled. Hope it  helps.
[out.txt](https://github.com/kata-containers/agent/files/4223554/out.txt)

nvidia.go
RyanAoh: Not sure whether it's necessary.Maybe later to add.

grpc.go
RyanAoh: My negligence. Thanks for your patience.
To fix

nvidia.go
RyanAoh: To add.

nvidia.go
RyanAoh: yeah，it works. 
To add.

nvidia.go
jodh-intel: @devimc is correct - please add a copyright comment to all new files.

Related: I've just created https://github.com/kata-containers/tests/issues/2323.

/cc @grahamwhaley

flx42:
@jodh-intel
@RenaudWasTaken should be your point of contact now :)

RenaudWasTaken:
> Here are the logs collected by running kata-collect-data.sh. Kata runtime and agent debug enabled. Hope it helps.

Looks like your logs are too late, they only show the delete commands after the container failed. It would also be nice to have the stdout of running the container :)

The simplest thing you can do is try to run through the kata CLI!

jodh-intel:
Thanks @flx42 and @RenaudWasTaken.

@RyanAoh - if you haven't already found it, here's how you can run Kata "stand-alone" (without a  container manager):

- https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#running-standalone

RyanAoh:
@RenaudWasTaken  I got the following error message. I think it could be helpful :
`kata-gpu-test-1 containerd[5229]: time="2020-02-27T08:56:43.968868354-05:00" level=debug msg="reading guest console" ID=282803f53f697766a9fbb31b46b45ac67f275366d29f3555b37d21df539a8fb8 sandbox=282803f53f697766a9fbb31b46b45ac67f275366d29f3555b37d21df539a8fb8 source=virtcontainers subsystem=kata_agent vmconsole="time=\"2020-02-27T13:56:43.961858575Z\" level=debug msg=\"create container err: rpc error: code = Internal desc = Could not run process: container_linux.go:346: starting container process caused \\\"process_linux.go:449: container [init](url) caused \\\\\\\"process_linux.go:432: running prestart hook 0 caused \\\\\\\\\\\\\\\"error running hook: exit status 1, stdout: , stderr: could not open OCI spec: open /config.json: no such file or directory\\\\\\\\\\\\\\\\n\\\\\\\\\\\\\\\"\\\\\\\"\\\"\" debug_console=false name=kata-agent pid=156 sandbox=282803f53f697766a9fbb31b46b45ac67f275366d29f3555b37d21df539a8fb8 source=agent"`

RyanAoh:
@RenaudWasTaken  I got the following error message. I think it could be helpful :
`kata-gpu-test-1 containerd[5229]: time="2020-02-27T08:56:43.968868354-05:00" level=debug msg="reading guest console" ID=282803f53f697766a9fbb31b46b45ac67f275366d29f3555b37d21df539a8fb8 sandbox=282803f53f697766a9fbb31b46b45ac67f275366d29f3555b37d21df539a8fb8 source=virtcontainers subsystem=kata_agent vmconsole="time=\"2020-02-27T13:56:43.961858575Z\" level=debug msg=\"create container err: rpc error: code = Internal desc = Could not run process: container_linux.go:346: starting container process caused \\\"process_linux.go:449: container [init](url) caused \\\\\\\"process_linux.go:432: running prestart hook 0 caused \\\\\\\\\\\\\\\"error running hook: exit status 1, stdout: , stderr: could not open OCI spec: open /config.json: no such file or directory\\\\\\\\\\\\\\\\n\\\\\\\\\\\\\\\"\\\\\\\"\\\"\" debug_console=false name=kata-agent pid=156 sandbox=282803f53f697766a9fbb31b46b45ac67f275366d29f3555b37d21df539a8fb8 source=agent"`

RenaudWasTaken:
Yep! That's a helpful error, looks like the [hook state](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/specs-go/state.go#L3-L17) doesn't contain the path to the container bundle :) 

We need to look into how the runtime invokes hooks :)

RyanAoh:
> Yep! That's a helpful error, looks like the [hook state](https://github.com/opencontainers/runtime-spec/blob/v1.0.0/specs-go/state.go#L3-L17) doesn't contain the path to the container bundle :)
> 

Thanks @RenaudWasTaken. I solved it refering to your reply. I missed the step to write the OCI spec to a file so that hooks can read it.(em...Since I am not very know how the runtime invokes hooks ).
Not, It can work well with  adding nvidia-container-runtime-hook as a prestart hook. Maybe  an update for this PR later.

RyanAoh:
Huh，with the agent working well with adding nvidia-container-runtime-hook. I got the the best way to configure nvidia GPU access for the container with no code changes.
1. config `guest_hook_path = "/usr/share/oci/hooks"` in the configuration file
2. locate "nvidia-container-toolkit" to "${ROOTFS_DIR}/usr/share/oci/hooks/prestart" (not nvidia-container-runtime-hook, since a symbolic is invalid)
Then the agent will add it. And the nvidia-container-toolkit will decide whether to configure nvidia GPU access by calling nvidia-container-cli  accord the env "NVIDIA_VISIBLE_DEVICES"

So, I think this PR and the issue #735 coud be closed. But maybe we need a detailed document to guide us how to use nvidia gpu device with kata containers.

jmsmkn:
> Huh，with the agent working well with adding nvidia-container-runtime-hook. I got the the best way to configure nvidia GPU access for the container with no code changes.
> 
> 1. config `guest_hook_path = "/usr/share/oci/hooks"` in the configuration file
> 2. locate "nvidia-container-toolkit" to "${ROOTFS_DIR}/usr/share/oci/hooks/prestart" (not nvidia-container-runtime-hook, since a symbolic is invalid)
>    Then the agent will add it. And the nvidia-container-toolkit will decide whether to configure nvidia GPU access by calling nvidia-container-cli  accord the env "NVIDIA_VISIBLE_DEVICES"
> 
> So, I think this PR and the issue #735 coud be closed. But maybe we need a detailed document to guide us how to use nvidia gpu device with kata containers.
> 
> Thanks everyone :)

Hi @RyanAoh, maybe you could help as I'm trying to get this working, and you're comment is very promising!

I cannot get the 2nd step to work - I added `nvidia-container-toolkit` to the root fs and rebuilt the kata image, set the hooks, but then when running the container the hook complains about missing `nvidia-container-cli`. So, I then copy all of `nvidia-container-cli list --binaries` and `nvidia-container-cli list --libraries` listed in your initial #704 comment, and then get the problem with `libnvidia-container.so.1: cannot open shared object file: No such file or directory`. Where would I run `ldconfig` without customising the agent like you did in this PR?

RyanAoh:
>Where would I run `ldconfig` without customising the agent like you did in this PR?
> 
emm, try to run `ldconfig /path/to/your_copy_dest` in the container before running `nvidia-smi` ?

Issue #737, opened 2020-02-17
Use "virtiofs" instead of "virtio_fs" for the virtio-fs module


fidencio:
# Description of problem

n the early stages of virtio-fs development, its module was called "virtio_fs". However, it's changed to "virtiofs" and that's the name used upstream.

Whenever we're trying to use a new enogh kernel, we'd look for virtio_fs instead of "virtiofs" causing a failure to load the module.

# Expected result
Work with the new enough kernels.

# Actual result
Doesn't work with new enough kernels.

NOTE: It's important to note that I'm not exactly familiar with the kata project with regards to when the kernel used by the project would / should be updated in order to reflect the changes I'm asking for in this bug report.

PR #738, opened 2020-02-17
mount: Use "virtiofs" instead of "virtio_fs"

fidencio:
virtio_fs was the name used for the module in the very early stages of
its development.

Resolves: #737

Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	type9pFs       = "9p"
-	typeVirtioFS   = "virtio_fs"
+	typeVirtioFS   = "virtiofs"
 	typeRootfs     = "rootfs"
 	typeTmpFs      = "tmpfs"
 	procMountStats = "/proc/self/mountstats"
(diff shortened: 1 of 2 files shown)


devimc:
thanks @fidencio , but I think we should keep backward compatibility, like @grahamwhaley said

fidencio:
> thanks @fidencio , but I think we should keep backward compatibility, like @grahamwhaley said

Makes sense. I will re-work both patches then.

fidencio:
@devimc, please take a look at https://github.com/kata-containers/runtime/pull/2463#issuecomment-587362867

I guess it's the very same case for this one.

amshinde:
> thanks @fidencio , but I think we should keep backward compatibility, like @grahamwhaley said

Are you planning to address this @fidencio ?

fidencio:
> > thanks @fidencio , but I think we should keep backward compatibility, like @grahamwhaley said
> 
> Are you planning to address this @fidencio ?

Sincerely, nops. I see no point to keep backwards compat with an option that was never ever released.

amshinde:
Merging this as https://github.com/kata-containers/runtime/pull/2604 has been merged.

amshinde:
@fidencio Can you revendor this in your shim PR: github.com/kata-containers/shim#233?

Issue #739, opened 2020-02-18
build: Add support to strip the agent golang binary


justin-he:
Which feature do you think can be improved?

Provide a flag "STRIP" for user to strip the golang binary

How can it be improved?

Additional Information

Before raising this feature request

Have you looked at the limitations document?
yes

PR #740, opened 2020-02-18
build: Add support to strip the agent golang binary

justin-he:
This provide a flag "STRIP=yes" to strip the golang binary
After make STRIP=yes
24757576 kata-agent.nostrip*
19651136 kata-agent.striped*
It reduces the golang binary size ~5MB

Fixs: #739

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ INIT := no
 # Set to "yes" if agent should support OpenTracing with http://jaegertracing.io.
 TRACE := no
 
+# Set to "yes“ if binary stripping is needed.
+STRIP := no
+
 # Tracing cannot currently be supported when running the agent as PID 1 since
 # the tracing requires additional services to be started _before_ the agent
 # process starts.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -85,6 +85,11 @@ else
 	BUILDFLAGS := -buildmode=pie
 endif
 
+# whether stipping the binary
```
grahamwhaley: Looking good. Can I ask we also set a default `no` at the top of the file, along with the other defaults, to make it easier to see what the set of build options are? Because, afaict, we don't have actual documentation on what the agent build options are ;-). So, add to this section:

```makefile
# Define if agent will be installed as init
INIT := no

# Set to "yes" if agent should support OpenTracing with http://jaegertracing.io.
TRACE := no
```

justin-he: Sure. will update it

Makefile
```diff
@@ -85,6 +88,11 @@ else
...
+ifeq ($(STRIP),yes)
+	LDFLAGS += -w -s
```
jcvenegas: Could you rename it with KATA_LDFLAGS LDFLAGS empty will use the ones from the host and some of them are not compatible with golang. In the agent may not be a big issue as we build it in a container based in recent experience could be  better if yes, will need to change https://github.com/kata-containers/agent/pull/740/files#diff-b67911656ef5d18c4ae36cb6741b7965R105  to use it

PR #754, opened 2020-02-28
agent: support pmem/nvdimm hotplug

### LINKED ISSUE #753: agent: support pmem/nvdimm hotplug
same as SCSI devices, the agent should wait for hotplugged pmem/nvdimm devices to appear in the guest

devimc:
Agent has to wait for pmem/nvdimm devices to appear in the guest.
The agent can use udev and the ACPI device path to identify when a pmem/nvdimm
device has been hotplugged in the guest.
For example when `/dev/pmem1` is hotplugged, the ACPI dev path may look like:

```
/devices/LNXSYSTM:00/LNXSYBUS:00/ACPI0012:00/ndbus0/region1/pfn1.1/block/pmem1
```

The udev routine catches the above path and return the full path to the device:
`/dev/pmem1`

fixes #753

cc @amshinde @lifupan @jodh-intel

### DIFF
--- a/device_amd64.go
+++ b/device_amd64.go
@@ -9,4 +9,15 @@ package main
 
 const (
 	rootBusPath = "/devices/pci0000:00"
+
+	// From https://www.kernel.org/doc/Documentation/acpi/namespace.txt
+	// The Linux kernel's core ACPI subsystem creates struct acpi_device
+	// objects for ACPI namespace objects representing devices, power resources
+	// processors, thermal zones. Those objects are exported to user space via
+	// sysfs as directories in the subtree under /sys/devices/LNXSYSTM:00
+	acpiDevPath = "/devices/LNXSYSTM"
+
+	// /dev/pmemX devices exported in the ACPI sysfs (/devices/LNXSYSTM) are
+	// in a subdirectory whose prefix is pfn (page frame number).
+	pfnDevPrefix = "/pfn"
 )
--- a/mount.go
+++ b/mount.go
@@ -217,6 +217,7 @@ var storageHandlerList = map[string]storageHandler{
 	driverSCSIType:      virtioSCSIStorageHandler,
 	driverEphemeralType: ephemeralStorageHandler,
 	driverLocalType:     localStorageHandler,
+	driverNvdimmType:    nvdimmStorageHandler,
 }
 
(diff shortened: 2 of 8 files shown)


grahamwhaley:
Silly question - do we have `udev` in all the distros we can use for the VM image? I dunno....

devimc:
@grahamwhaley yes, kata-agent throws a goroutine for monitoring udev events (inhouse implementation of udev)

devimc:
@amshinde @jodh-intel  please take a look

device_amd64.go
```diff
@@ -8,5 +8,7 @@
 package main
 
 const (
-	rootBusPath = "/devices/pci0000:00"
+	rootBusPath  = "/devices/pci0000:00"
+	acpiDevPath  = "/devices/LNXSYSTM"
```
amshinde: I would add docs around this, something like
```
The Linux kernel's core ACPI subsystem creates struct acpi_device
   objects for ACPI namespace objects representing devices, power resources
   processors, thermal zones.  Those objects are exported to user space via
   sysfs as directories in the subtree under /sys/devices/LNXSYSTM:00.
```

device_amd64.go
```diff
@@ -8,5 +8,7 @@
...
+	pfnDevPrefix = "/pfn"
```
amshinde: Can we add some explanation comments for this?

mount.go
```diff
@@ -299,10 +299,20 @@ func virtioBlkStorageHandler(_ context.Context, storage pb.Storage, s *sandbox)
 	// If hot-plugged, get the device node path based on the PCI address else
 	// use the virt path provided in Storage Source
 	if strings.HasPrefix(storage.Source, "/dev") {
-
+		// waiting for a pmem/nvdimm device
```
amshinde: Why did you go with adding this logic in the virtioBlkStorageHandler @devimc ?
How about having a separate handler for pmem devices?

devimc: because it's the same logic that we have in the runtime

https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu.go#L1031-L1048

`hotplugAddBlockDevice` -> `ExecuteNVDIMMDeviceAdd`

devimc: @amshinde  ignore my comment, I know what you mean

devimc:
@amshinde changes applied, thanks

Issue #767, opened 2020-04-02
SCSI device scan fails on some kernels


errordeveloper:
While testing custom kerenls, I often noticed seeing the following error in pod status:

```
write /sys/class/scsi_host/host1/scan invalid argument
```

This is caused by:

https://github.com/kata-containers/agent/blob/c1d24f92e9e2fb454ffc70fe8583e49ca80ed893/device.go#L321-L348

I wonder if this is provides some critical functionality in Kubernetes context, or a configuration parameter could be added to turn it off without a major compromise? From the original PR https://github.com/kata-containers/agent/pull/163 it sounds as if this was not a critical feature as such, but I'm not sure if it has become one since then.

I guess aside from adding a toggle for this, there could be a way to fix this that e.g. by turning this error into a warning or being less opportunistic on how the SCSI scan is performed?

errordeveloper:
@jodh-intel @devimc  I have enabled debug logs, but not been able to tell as what comes out in the journal is rather very verbose and mostly unreadable as most messages get doubly quoted etc. Is there a tool that people use to parse it?
To be clear, my pod doesn't have any volumes, it would help to know if the this arises from the fact that kata passes default secret as a SCSI device to the VM, I have no idea how to tell whether this is the case or it's something else.

errordeveloper:
I've double checked the [kernel config](https://github.com/linuxkit/linuxkit/blob/87f2e7542484baf3a223aef7b3a61c7ba0f8cad7/kernel/config-4.19.x-x86_64#L1746-L1767) that I used, and there are indeed a few difference, albeit SCSI is generally enabled.

```
#
# SCSI device support
#
CONFIG_SCSI_MOD=y
CONFIG_RAID_ATTRS=m
CONFIG_SCSI=y
CONFIG_SCSI_DMA=y
# CONFIG_SCSI_MQ_DEFAULT is not set
# CONFIG_SCSI_PROC_FS is not set

#
# SCSI support type (disk, tape, CD-ROM)
#
CONFIG_BLK_DEV_SD=y
# CONFIG_CHR_DEV_ST is not set
# CONFIG_CHR_DEV_OSST is not set
CONFIG_BLK_DEV_SR=y
# CONFIG_BLK_DEV_SR_VENDOR is not set
CONFIG_CHR_DEV_SG=y
# CONFIG_CHR_DEV_SCH is not set
# CONFIG_SCSI_CONSTANTS is not set
# CONFIG_SCSI_LOGGING is not set
[...]

PR #768, opened 2020-04-02
Add flags to ignore device attachement errors

errordeveloper:
- `agent.ingnore_scsi_bus_scan_errors` to avoid #767
- `agent.ingnore_all_device_errors` to avoid any device errors  causing agent to die

### DIFF
--- a/device.go
+++ b/device.go
@@ -352,7 +352,10 @@ func scanSCSIBusImpl(scsiAddr string) error {
 // associated with the disk.
 func getSCSIDevPathImpl(s *sandbox, scsiAddr string) (string, error) {
 	if err := scanSCSIBus(scsiAddr); err != nil {
-		return "", err
+		if !ignoreSCSIBusScanErrs {
+			return "", err
+		}
+		agentLog.WithError(err).Warnf("SCSI bus scan failed")
 	}
 
 	devPath := filepath.Join(scsiHostChannel+scsiAddr, scsiBlockSuffix)
@@ -422,9 +425,11 @@ func addDevices(ctx context.Context, devices []*pb.Device, spec *pb.Spec, s *san
 
 		err := addDevice(ctx, device, spec, s)
 		if err != nil {
-			return err
+			if !ignoreAllDeviceErrs {
+				return err
+			}
+			agentLog.WithError(err).Warnf("Could not attach %s device", device.Type)
 		}
-
 	}
 
 	return nil
--- a/agent.go
+++ b/agent.go
@@ -190,6 +190,9 @@ var unifiedCgroupHierarchy = false
 // Size in bytes of the stdout/stderr pipes created for each container.
 var containerPipeSize = uint32(0)
 
+var ignoreSCSIBusScanErrs = false
+var ignoreAllDeviceErrs = false
+
 // commType is used to denote the communication channel type used.
 type commType int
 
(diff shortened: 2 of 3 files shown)


device.go
```diff
@@ -352,7 +352,10 @@ func scanSCSIBusImpl(scsiAddr string) error {
 // associated with the disk.
 func getSCSIDevPathImpl(s *sandbox, scsiAddr string) (string, error) {
 	if err := scanSCSIBus(scsiAddr); err != nil {
-		return "", err
+		if !ignoreSCSIBusScanErrs {
+			return "", err
+		}
+		agentLog.WithError(err).Warnf("SCSI bus scanl failed")
```
dagrh, line 358: typo 'scanl' rather than 'scan'

devimc:
@errordeveloper thanks for raising this, however I'm not sure if we really want these new parameters, if there is an error adding a device or re-scannning a bus then we'll end up having an incomplete sandbox/container, isn't it?
Rootfs block devices hot added may not be detected (devicemapper case) because of the lack of SCSI support in the guest, did  you try it out?
IMO guest kernel must fully support SCSI whether it's intended to be used as kernel for a kata container

errordeveloper:
@davimc the guest kernel I was testing does support SCSI from what I can tell by the [config](https://github.com/linuxkit/linuxkit/blob/87f2e7542484baf3a223aef7b3a61c7ba0f8cad7/kernel/config-4.19.x-x86_64#L1748). From my point of view these params are useful as means of debugging, this way I was actually able to use what I have here and confirm that my kernel of choice is functional. I just need to debug a few more thing if it doesn't work, but actually right now I see no issues here, could it be just that kata is trying to pass some host devices to the VM that I didn't even know existed? Or is a SCSI device being used to mount default service account secret into my pod?

jodh-intel:
I agree with @devimc, we should not be ignoring errors. Obviously, it's fine to use this branch for your own private debug though.

A couple of thoughts:

- Have you compared your guest kernel config with the Kata guest kernel configs here: https://github.com/kata-containers/packaging/tree/master/kernel ?
- Have you enabled [full debug](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#enable-full-debug)?

errordeveloper:
Let's have the debugging discussion in the original issue - #767.

bergwolf:
Ditto, please do not add these code injections just for a special debugging case.
> agent.ingnore_all_device_errors to avoid any device errors causing agent to die

AFAICT, agent should not die in such case. It just logs the error and fail the corresponding grpc request. Did you actually see agent dying in your test? If so, we should fix it as it's certainly a bug.

errordeveloper:
> Did you actually see agent dying in your test? If so, we should fix it as it's certainly a bug.

I don't know actually, it's possible that the agent does exit, but as a result VM gets thrown out, as it's failed cause as far as CRI is concerned...

jodh-intel:
@errordeveloper - You could try running our data collection script that looks for interesting errors, etc as it might give us further information on the problem you are seeing:

- Run the following:
  ```sh
  $ sudo kata-collect-data.sh > /tmp/kata.log
  ```
- Review the contents of `/tmp/kata.log` (to make sure there are no private/sensitive details).
- If you are happy, pasting the contents of `/tmp/kata.log` as a comment here (or on #767).

amshinde:
Closing this PR as this has gone stale. And we really dont want to add code injections as pointed out by @bergwolf .
@errordeveloper Please feel to reopen, once you have addressed the comments.

Issue #775, opened 2020-04-30
vsock: initiate a quick handshake to avoid VSOCK_DEFAULT_CONNECT_TIMEOUT


Pennyzct:
Related issues: https://github.com/kata-containers/runtime/issues/1917

PR #776, opened 2020-04-30
vsock: initiate a quick handshake to avoid VSOCK_DEFAULT_CONNECT_TIMEOUT(WIP)

Pennyzct:
We will suffer default two seconds timeout, when kata-runtime initiates a connection to agent server, and right now vsock transport in guest part isn't ready.
Here, taking advices from Stefan Hajnoczi, we will do a quick handshake to ensure host initiating connection until vsock device is completely ready.
The quick handshake follows below steps:
1. kata-runtime listens on a vsock port
2. `agent.vsock_host_port=PORT` is added to the kernel command-line options
3. kata-agent parses the port number and connects to the host
This commit covers the third part in kata-agent.

Fixes: #775

Signed-off-by: Penny Zheng <penny.zheng@arm.com>

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -27,6 +27,7 @@ var (
 	channelExistMaxTries   = 200
 	channelExistWaitTime   = 50 * time.Millisecond
 	channelCloseTimeout    = 5 * time.Second
+	vsockReadyTimeout      = 2 * time.Second
 	isAFVSockSupportedFunc = isAFVSockSupported
 )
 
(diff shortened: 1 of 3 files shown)


Pennyzct:
Hi guys
another part resides in kata-runtime, plz see https://github.com/kata-containers/runtime/pull/2656

channel.go
```diff
@@ -132,7 +133,38 @@ func (c *vSockChannel) setup() error {
 }
 
 func (c *vSockChannel) wait() error {
-	return nil
+	if vsockHostPort == uint32(0) {
+		return fmt.Errorf("fail to get vsock host port %d", vsockHostPort)
```
stefanha, line 137: Please mention agent.vhost_vsock_port in this error message so it's clear that this parameter is missing.

channel.go
```diff
@@ -132,7 +133,38 @@ func (c *vSockChannel) setup() error {
...
+				return
+			}
+		}
+	}()
+
+	select {
+	case <-conn:
+		return nil
+	case <-time.After(vsockReadyTimeout):
+		cancel <- true
```
stefanha, line 165: Does this cancelation really work?  When time.After(vsockReadyTimeout) expires we send true to the goroutine that is blocked in vsock.Dial().  Nothing interrupts vsock.Dial() so I guess it will continue waiting until the vsock connection itself completes, failsm or times out.

Goroutines probably aren't necessary here: you can just call vsock.Dial().  It will time out after 2 seconds or you can use setsockopt(SO_VM_SOCKETS_CONNECT_TIMEOUT) to set a different timeout.

PR #785, opened 2020-05-11
Makefile: do not use LDFLAGS to avoid environment contamination

ChenQi1989:
The logic of this Makefile is to use "-extldflags '-static'" if
STATIC is defined. On the other hand, LDFLAGS is the standard name.
So we should avoid using LDFLAGS to avoid build failure.

Signed-off-by: Chen Qi <Qi.Chen@windriver.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ else
 endif
 # go build common flags
 ifdef STATIC
-	LDFLAGS := -extldflags '-static'
+	STATIC_LDFLAGS := -extldflags '-static'
 else
 	BUILDFLAGS := -buildmode=pie
 endif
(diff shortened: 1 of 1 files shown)


ChenQi1989:
> Thanks @ChenQi1989 (and welcome! ;)
> 
> Please note that you need to change your commit message slightly to reference an existing GitHub issue before our CI is happy though - see https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format.
> 
> lgtm

THANKS. I've created an issue and updated the commit.

jodh-intel:
/test-ubuntu

Ping @kata-containers/agent

fidencio:
Btw, I've just checked and we don't have similar issues in the other projects.

PR #788, opened 2020-05-15
namespaces: Allow container with agent PID namespace

### LINKED ISSUE #787: Allow containers to join the agent/init PID namespace.
We have had requests from users where they would like to run debug sidecars for gathering debug/audit data for a pod. For this, it is necessary for an audit container to be in the same PID namespace as the agent/init process to be able gather audit events.

(This is somewhat analogous to the "hostpid" functionality provided by
k8s, but in this case this is applied within the guest and on a per
container basis)

amshinde:
This change adds a new field "agentPidns" to the CreateContainer
request. When this is set to true, the container instead of
creating and joining a new PID namespace will skip creating a PID
namespace and be in the same PID namespace as the agent/init process.

I am adding this functionality to support certain sidecar containers
such as debug/audit containers to be able to gather audit data/kernel
debug info.

Fixes #787


lifupan:
Will this bring  the security issues such as the users can access/kill the agent process?

devimc:
@amshinde what's the complete plan (what else is needed?) to support debug/audit containers?

changing this to `dnm`

grpc.go
```diff
@@ -520,6 +520,12 @@ func (a *agentGRPC) updateContainerConfigNamespaces(config *configs.Config, ctr
 		config.Namespaces = append(config.Namespaces, newUTSNs)
 	}
 
+	// If container needs to be in the agent PID namespace, do not create a new
+	// PID namespace.
+	if ctr.agentPidNs {
+		return
```
jodh-intel, line 526: I think we should log this as it has potential security implications.

protocols/grpc/agent.proto
```diff
@@ -78,6 +78,12 @@ message CreateContainerRequest {
 	// The agent would receive an OCI spec with PID namespace cleared
 	// out altogether and not just the pid ns path.
 	bool sandbox_pidns = 7;
+
+	// This field is used to indicate if container should just join the
+	// the pid namespace of the agent. This functionality is added to
+	// allow debug containers/sidecars to have access to the main pid
+	// namespace.
+	bool agent_pidns = 8;
```
jodh-intel, line 86: Idea: given this is a per container setting and given there are potential security implications of allowing anything to run in the agents NS, we might want to consider passing a similar bool to the agent on startup that says whether the agent should honour this setting. By default, it would either ignore it (and log that fact), or error. However, if `agent_pidns=true` is set in `configuration.toml`, it would be honoured. In fact, if this is *only* to be used in debug scenarios, maybe we should also only honour this setting if `enable_debug=true` for the agent too?

grpc.go
amshinde: Makes sense. Will add a log message here.

protocols/grpc/agent.proto
amshinde: Agree, it makes sense to me to add an additional configuration  called `agent_pidns` in the configuration file to have this feature enabled. That way only the only the root user can enable this feature. I am not so sure about adding the requirement for enable_debug flag as well, will give this some thought.

amshinde:
> @amshinde what's the complete plan (what else is needed?) to support debug/audit containers?

@devimc I was hoping I could achieve this by specifying the container name for which agent pid sharing needs to be enabled, as a pod annotation. I see that the container name is passed down to the container in case of crio. But I do not see the same behaviour in case of containerd, containerd does not pass the container name that kubelet passes to it to the OCI runtime. In any case, since the kubelet annotation itself is not really documented API, I dont think we should be relying on it.

The only way to implement this would be to specify this as an environment variable for the container. 
I think this should be ok to do, given that we plan to enable the agent pidns sharing behaviour by explicitly enabling this in the configuration file. 
Wdyt @devimc @jodh-intel ?

bergwolf:
While I agree this is a useful feature for debugging/auditing use case, we have decided to feature freeze for kata 1.x releases. Please re-submit it to https://github.com/kata-containers/kata-containers Thanks!

If it is strongly desired to include this in the 1.x kata releases like v1.12, please backport it from the 2.0-dev branch after it is merged there.

bergwolf:
As we have discussed this in PTG and the conclusion is to allow new features to be in 1.x releases but they must be ported to 2.0-dev branch as well. For this PR, @amshinde will be the one to port it to 2.0-dev branch. Am I understanding it correctly?

README.md
```diff
@@ -98,6 +98,13 @@ The pipe's capacity for stdout/stderr can be modified by specifying the `agent.c
 to the guest kernel command line. For example, `agent.container_pipe_size=2097152` will set the stdout and stderr
 pipes to 2097152 bytes.
 
+## Agent pid namespace 
+
+The agent will allow a container to share the pid namespace of the agent when `agent.agent_pidns` is true.
```
devimc, line 103: mmm this could be risky, any user would be able to set this through annotations, right?

amshinde:
> Will this bring the security issues such as the users can access/kill the agent process?

@lifupan Yes, this is meant only as a debug feature and needs to be explicitly turned on by changing the configuration.toml file. 
Have a look at the corresponding runtime PR: 
https://github.com/kata-containers/runtime/pull/2723

amshinde:
> As we have discussed this in PTG and the conclusion is to allow new features to be in 1.x releases but they must be ported to 2.0-dev branch as well. For this PR, @amshinde will be the one to port it to 2.0-dev branch. Am I understanding it correctly?

Yes, I shall backport this @bergwolf

amshinde:
@devimc @jodh-intel @lifupan @bergwolf This is ready for review, please take a look.

jodh-intel:
We could now force-merge this (ignoring the known failing Zuul WIP check), but might be good to get final comments / approval from @devimc first.

PR #792, opened 2020-06-03
device: Do not allow container access to the nvdimm rootfs

### LINKED ISSUE #791: Explicitly deny any access to the nvdimm root partition
Explicitly deny any access to the nvdimm root partition by adding the nvdimm device to the device cgroup.

amshinde:
With this change, a container is not longer given access to
the underlying nvdimm root partition.
This is done by explicitly adding the nvdimm root partition
to the device cgroup of the container.

Fixes #791

### DIFF
--- a/device.go
+++ b/device.go
@@ -34,6 +34,7 @@ const (
 	driverNvdimmType    = "nvdimm"
 	driverEphemeralType = "ephemeral"
 	driverLocalType     = "local"
+	vmRootfs            = "/"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


device.go
```diff
@@ -50,6 +50,7 @@ var (
 	getPCIDeviceName    = getPCIDeviceNameImpl
 	getDevicePCIAddress = getDevicePCIAddressImpl
 	scanSCSIBus         = scanSCSIBusImpl
+	rootfsNvdimmDisk    = "/dev/pmem0p1"
```
devimc, line 53: I think there is a different way to the rootfs device

amshinde: @devimc PR updated. PTAL

devimc: thanks

devimc:
restarting ubuntu CI

PR #798, opened 2020-06-18
oci: Fix running of OCI hooks

bpradipt:
OCI hooks fails to run since the code was writing the config.json
to the read-only path. This patch fixes it

Fixes: #2763

Signed-off-by: Pradipta Kumar bpradipt@in.ibm.com

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -676,14 +676,14 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		a.sandbox.addGuestHooks(ociSpec)
 
 		// write the OCI spec to a file so that hooks can read it
-		err = writeSpecToFile(ociSpec)
+		err = writeSpecToFile(ociSpec, req.ContainerId)
 		if err != nil {
 			return emptyResp, err
 		}
 
 		// Change cwd because libcontainer assumes the bundle path is the cwd:
 		// https://github.com/opencontainers/runc/blob/v1.0.0-rc5/libcontainer/specconv/spec_linux.go#L157
-		oldcwd, err := changeToBundlePath(ociSpec)
+		oldcwd, err := changeToBundlePath(ociSpec, req.ContainerId)
 		if err != nil {
 			return emptyResp, err
 		}
@@ -1305,6 +1305,9 @@ func (a *agentGRPC) RemoveContainer(ctx context.Context, req *pb.RemoveContainer
 		}
 	}
 
+	configJsonDir := filepath.Join("/run/libcontainer/", req.ContainerId)
+	//Best effort.. Ignore any errors in the deletion of the directory
+	_ = os.RemoveAll(configJsonDir)
 	delete(a.sandbox.containers, ctr.id)
 
 	return emptyResp, nil
--- a/oci.go
+++ b/oci.go
@@ -22,13 +22,18 @@ import (
 const (
 	ociConfigFile     string      = "config.json"
 	ociConfigFileMode os.FileMode = 0444
+	ociConfigBasePath string      = "/run/libcontainer"
 )
 
-// writeSpecToFile writes the container's OCI spec to "dirname(spec.Root.Path)/config.json"
-// This effectively makes the parent directory a valid OCI bundle.
-func writeSpecToFile(spec *specs.Spec) error {
-	bundlePath := filepath.Dir(spec.Root.Path)
-	configPath := filepath.Join(bundlePath, ociConfigFile)
+// writeSpecToFile writes the container's OCI spec to "/run/libcontainer/<container-id>/config.json"
+// Note that the OCI bundle (rootfs) is at a different path
+func writeSpecToFile(spec *specs.Spec, containerId string) error {
+	configJsonDir := filepath.Join(ociConfigBasePath, containerId)
+	err := os.MkdirAll(configJsonDir, 0700)
+	if err != nil {
+		return err
+	}
+	configPath := filepath.Join(configJsonDir, ociConfigFile)
 	f, err := os.OpenFile(configPath, os.O_WRONLY|os.O_CREATE, ociConfigFileMode)
 	if err != nil {
 		return err
@@ -40,7 +45,7 @@ func writeSpecToFile(spec *specs.Spec) error {
 
 // changeToBundlePath changes the cwd to the OCI bundle path defined as
 // dirname(spec.Root.Path) and returns the old cwd.
-func changeToBundlePath(spec *specs.Spec) (string, error) {
+func changeToBundlePath(spec *specs.Spec, containerId string) (string, error) {
 	cwd, err := os.Getwd()
 	if err != nil {
 		return cwd, err
@@ -51,9 +56,10 @@ func changeToBundlePath(spec *specs.Spec) (string, error) {
 	}
 
 	bundlePath := filepath.Dir(spec.Root.Path)
-	configPath := filepath.Join(bundlePath, ociConfigFile)
+	configPath := filepath.Join(ociConfigBasePath, containerId, ociConfigFile)
 
-	// Verify that config.json is present at the root of the bundle path.
+	// config.json is at "/run/libcontainer/<container-id>/"
+	// Actual bundle (rootfs) is at dirname(spec.Root.Path)
 	if _, err := os.Stat(configPath); err != nil {
 		return cwd, errors.New("invalid OCI bundle")
 	}
--- a/oci_test.go
+++ b/oci_test.go
@@ -17,6 +17,8 @@ import (
 )
 
 func TestChangeToBundlePath(t *testing.T) {
+	skipUnlessRoot(t)
+	containerId := "1"
 	assert := assert.New(t)
 
 	originalCwd, err := os.Getwd()
@@ -37,15 +39,15 @@ func TestChangeToBundlePath(t *testing.T) {
 		Readonly: false,
 	}
 
-	_, err = changeToBundlePath(spec)
+	_, err = changeToBundlePath(spec, containerId)
 	assert.Error(err)
 
 	// Write the spec file to create a valid OCI bundle
 	spec.Root.Path = rootfsPath
-	err = writeSpecToFile(spec)
+	err = writeSpecToFile(spec, containerId)
 	assert.NoError(err)
 
-	cwd, err := changeToBundlePath(spec)
+	cwd, err := changeToBundlePath(spec, containerId)
 	assert.NoError(err)
 	assert.Equal(cwd, originalCwd)
 
@@ -55,6 +57,8 @@ func TestChangeToBundlePath(t *testing.T) {
 }
 
 func TestWriteSpecToFile(t *testing.T) {
+	skipUnlessRoot(t)
+	containerId := "1"
 	assert := assert.New(t)
 
 	bundlePath, err := ioutil.TempDir("", "bundle")
@@ -75,10 +79,17 @@ func TestWriteSpecToFile(t *testing.T) {
 			Readonly: false,
 		},
 	}
-	err = writeSpecToFile(spec)
+	err = writeSpecToFile(spec, containerId)
 	assert.NoError(err)
 
 	file, err := os.Open(path.Join(bundlePath, ociConfigFile))
+	assert.Error(err)
+	defer file.Close()
+
+	_, err = file.Stat()
+	assert.Error(err)
+
+	file, err = os.Open(path.Join("/run/libcontainer/", containerId, ociConfigFile))
 	assert.NoError(err)
 	defer file.Close()
 


oci.go
```diff
@@ -20,14 +20,15 @@ import (
...
 // OCI config file
 const (
-	ociConfigFile     string      = "config.json"
-	ociConfigFileMode os.FileMode = 0444
+	ociConfigFile       string      = "config.json"
+	ociConfigFileMode   os.FileMode = 0444
+	ociConfigBundlePath string      = "/run/kata-containers/sandbox"
 )
 
 // writeSpecToFile writes the container's OCI spec to "dirname(spec.Root.Path)/config.json"
```
devimc, line 28: Update this comment please

oci.go
```diff
@@ -20,14 +20,15 @@ import (
 
 // OCI config file
 const (
-	ociConfigFile     string      = "config.json"
-	ociConfigFileMode os.FileMode = 0444
+	ociConfigFile       string      = "config.json"
+	ociConfigFileMode   os.FileMode = 0444
+	ociConfigBundlePath string      = "/run/kata-containers/sandbox"
```
devimc, line 25: mmm not sure if this is a good idea, config.json is container specific, so in the case of a POD with 2 or more containers, do they share the same config.json ?

bpradipt: @devimc, yeah good catch. They won't share the config.json. I think putting it under sandbox id will be needed.

oci.go
bpradipt: sure

oci.go
bpradipt: @devimc updated the patch.. please take a look and let me know if you see any issues

bpradipt:
@devimc @bergwolf can you please check this and advise

oci.go
```diff
@@ -22,13 +22,19 @@ import (
...
 )
 
-// writeSpecToFile writes the container's OCI spec to "dirname(spec.Root.Path)/config.json"
-// This effectively makes the parent directory a valid OCI bundle.
+// writeSpecToFile writes the container's OCI spec to "/run/libcontainer/<container-id>/config.json"
+// Note that the OCI bundle (rootfs) is at a different path
 func writeSpecToFile(spec *specs.Spec) error {
-	bundlePath := filepath.Dir(spec.Root.Path)
-	configPath := filepath.Join(bundlePath, ociConfigFile)
+	_, containerID := filepath.Split(filepath.Dir(spec.Root.Path))
```
devimc, line 31: please modify the signature of this function to receive the spec and container ID

bpradipt: Thanks @devimc for your suggestion. I was in two minds whether to change the signature or not. It's the better approach

oci.go
```diff
@@ -22,13 +22,18 @@ import (
...
 	ociConfigFileMode os.FileMode = 0444
+	ociConfigBasePath string      = "/run/libcontainer"
 )
 
-// writeSpecToFile writes the container's OCI spec to "dirname(spec.Root.Path)/config.json"
-// This effectively makes the parent directory a valid OCI bundle.
-func writeSpecToFile(spec *specs.Spec) error {
-	bundlePath := filepath.Dir(spec.Root.Path)
-	configPath := filepath.Join(bundlePath, ociConfigFile)
+// writeSpecToFile writes the container's OCI spec to "/run/libcontainer/<container-id>/config.json"
```
devimc, line 28: jfyi - this directory already exist in the guest

```
# ls -lR /run/libcontainer/
/run/libcontainer/:
total 0
drwx------    3 root     root            60 Jun 22 17:40 078a6955095206cb6cf326c82a1dc6d7f5329edab206774f71a668d1b95ea172

/run/libcontainer/078a6955095206cb6cf326c82a1dc6d7f5329edab206774f71a668d1b95ea172:
total 0
drwx--x--x    2 root     root            60 Jun 22 17:40 078a6955095206cb6cf326c82a1dc6d7f5329edab206774f71a668d1b95ea172

/run/libcontainer/078a6955095206cb6cf326c82a1dc6d7f5329edab206774f71a668d1b95ea172/078a6955095206cb6cf326c82a1dc6d7f5329edab206774f71a668d1b95ea172:
total 12
-rw-r--r--    1 root     root         10033 Jun 22 17:40 state.json
```

bpradipt: Yes @devimc. The root `/run/libcontainer` exists, however the container sub-dir doesn't exist when writeSpecToFile gets called.

amshinde: @bpradipt Do we already have directory /run/libcontainer/<pod-id>/<container-id> ?
If not maybe just use that ?

bpradipt: @amshinde when writeSpecToFile is called, we only have `/run/{kata-containers, sandbox-ns}`

oci.go
```diff
@@ -22,13 +22,18 @@ import (
...
+// Note that the OCI bundle (rootfs) is at a different path
+func writeSpecToFile(spec *specs.Spec, containerId string) error {
+	configJsonDir := filepath.Join(ociConfigBasePath, containerId)
```
lifupan, line 31: Once the container terminated, was the file be removed?

bpradipt: Hmm I'm not sure. However given that there could be `poststop` hooks which will run after the container has stopped, the config json might not be removed by the runtime.  @bergwolf would you have any idea ?

lifupan: I think we can do the cleanup in function: RemoveContainer  https://github.com/kata-containers/agent/blob/master/grpc.go#L1252

bpradipt: I see that the `/run/` folder is a tmpfs mount. The config.json should get removed automatically along with other files that are there.

lifupan: Yes, that's only when the pod is stopped.  But if the pod is still live and users create/stop many containers, then there would be much unused spec files left there.

bpradipt: I understand the scenario now. Thanks @lifupan. So in this case should we also handle cleanup of state.json that is kept under `/run/libcontainer/<id>`? And if yes, does it make sense to handle cleanup in a separate patch or club with this one ?
Also the impact of removing config.json on container `poststop` hook needs to be checked. Let me know your thoughts @lifupan

bergwolf: Could you check if `libcontainer` removes `/run/libocntainer/<id>` entirely upon container removal? If so, there is nothing to be done in the agent code.

amshinde: @bpradipt Lets make sure we clean up correctly in this patch.

lifupan: Hi @bpradipt  @bergwolf  The dir  /run/libocntainer/<id>  was created by this patch in agent, I don't think libcontainer would do cleanup of this dir.  BTW, the libcontainer would run he poststophooks in container's Destroy function at https://github.com/kata-containers/agent/blob/master/grpc.go#L1264 or https://github.com/kata-containers/agent/blob/master/grpc.go#L1279, so you can just do the cleanup behind that destroy.

bpradipt:
I have made the changes to remove the directory post container stop. However I noticed one oddity which I had missed earlier. 

Semantically the hooks expect `config.json` to be in the `bundlePath`.  However the `bundlePath` location is now read-only and we have to write the config.json to a different path.  While the actual hooks code can definitely use the new path, but it means that the hooks that were working with standard `runc` containers will not work as-is with Kata containers. 
I feel we'll need to think through this aspect since it affects end user experience.

@bergwolf @lifupan @fidencio @devimc let me know what you think and hope I'm not missing something fundamental here.

grpc.go
```diff
@@ -1273,6 +1273,9 @@ func (a *agentGRPC) RemoveContainer(ctx context.Context, req *pb.RemoveContainer
 				}
 			}
 		}
+		configJsonDir := filepath.Join("/run/libcontainer/", req.ContainerId)
```
lifupan, line 1276: Hi  @bpradipt 
Here you just do the cleanup in case of the timeout == 0,  how about 
doing the cleanup at the  bottom of this function or put it a delay function?

bpradipt: Hi @lifupan yeah sure. I'll make the changes and test it out.  Can you tell me what's the `timeout` is for? I didn't hit the else path during my local testing.

amshinde:
> I have made the changes to remove the directory post container stop. However I noticed one oddity which I had missed earlier.
> 
> Semantically the hooks expect `config.json` to be in the `bundlePath`. However the `bundlePath` location is now read-only and we have to write the config.json to a different path. While the actual hooks code can definitely use the new path, but it means that the hooks that were working with standard `runc` containers will not work as-is with Kata containers.
> I feel we'll need to think through this aspect since it affects end user experience.

@bpradipt Thanks for tackling this. This functionality was broken for a while since we moved to a read-only rootfs. We do not have CI today to catch this. 
Your above concern looks valid to me. As I recall, OCI compliant hooks would  expect `bundlePath` as the working directory during execution and the bundlePath is expected to contain both the container rootfs and config.json.
A OCI compliant hook may need to be rewritten, since the above expectation is not fulfilled with this PR. 
What if we write the `config.json` file to the shared `bundlePath` from the host side, with the `kata-runtime` doing so.
@bergwolf @lifupan @devimc  Do you see any issues with this approach?

amshinde:
@bpradipt Would also like to see an integration test for this being added, this could just be a dummy oci compliant hooks binary, that reads the config.json to get the path to the rootfs and perform a simple mount in the rootfs.

lifupan:
> > I have made the changes to remove the directory post container stop. However I noticed one oddity which I had missed earlier.
> > Semantically the hooks expect `config.json` to be in the `bundlePath`. However the `bundlePath` location is now read-only and we have to write the config.json to a different path. While the actual hooks code can definitely use the new path, but it means that the hooks that were working with standard `runc` containers will not work as-is with Kata containers.
> > I feel we'll need to think through this aspect since it affects end user experience.
> 
> @bpradipt Thanks for tackling this. This functionality was broken for a while since we moved to a read-only rootfs. We do not have CI today to catch this.
> Your above concern looks valid to me. As I recall, OCI compliant hooks would expect `bundlePath` as the working directory during execution and the bundlePath is expected to contain both the container rootfs and config.json.
> A OCI compliant hook may need to be rewritten, since the above expectation is not fulfilled with this PR.
> What if we write the `config.json` file to the shared `bundlePath` from the host side, with the `kata-runtime` doing so.
> @bergwolf @lifupan @devimc Do you see any issues with this approach?
> 

@amshinde  There would be some changes to spec in the guest such as cpuset reset, if the hooks didn't care about this changes, I think writing from  the host side is reasonable.

bergwolf:
@amshinde @bpradipt @lifupan the normal OCI container hooks are setup by containerd and kata runs them on the host side. Are you suggesting that kata should set it up and run in the guest instead?

bpradipt:
> @amshinde @bpradipt @lifupan the normal OCI container hooks are setup by containerd and kata runs them on the host side. Are you suggesting that kata should set it up and run in the guest instead?

@bergwolf this is for OCI hooks running inside the Kata VM (https://github.com/kata-containers/runtime/blob/master/cli/config/configuration-qemu.toml.in#L288).

amshinde:
> @amshinde @bpradipt @lifupan the normal OCI container hooks are setup by containerd and kata runs them on the host side. Are you suggesting that kata should set it up and run in the guest instead?

As @bpradipt This PR is for running the OCI hooks inside the guest. We run hooks passed in the OCI spec on the host, but we added this functionality to support use cases which require certain hooks to be run inside the guest, for eg in case of Nvidia GPU plugin. The hooks need to be part of the guest rootfs. We added this functionality as part of this PR : https://github.com/kata-containers/agent/pull/393

But since we made the guest rootfs as read-only, the above functionality was broken. 
If you look at this comment, https://github.com/kata-containers/agent/pull/346#issuecomment-420012768, the config.json needs to be written in the bundle directory which was being done by the agent. This PR attempts to fix the read-only error caused by the agent attempting to write to the bundle directory in the guest by instead having the agent writing the config.json to a directory under /run. With this though the config.json no longer exists in the bundle directory which may break OCI hooks that depend on this.
My suggestion is to instead have the runtime write the config.json to the shared bundle directory before it is passed as read-only to the guest.

dgibson:
I have a whole batch of thoughts on this.

1) I think always writing the config.json to the shared bundle directory would be really useful for debugging, regardless of its applicability to guest hooks

2) It's not clear to me from the OCI spec whether hooks are allowed to *alter* the config.json.  If so, that's obviously still a problem for guest hooks if it's in a read-only path.

3) Treating guest hooks as kinda-OCI hooks seems bogus to me anyway.  Unlike real OCI hooks they're controlled by the runtime's configuration, rather than the container's configuration.  They also run in a different environment from that specified for OCI hooks.  i.e. they're really not the same as OCI hooks.

Having guest side hooks is certainly useful for cases, but I think we should treat (and name) them as a kata specific thing, not pretend they have any real connection to OCI hooks.  (The fact that they're implemented via OCI hooks executed on the "inner" container is an implementation detail that shouldn't be exposed in naming and configuration).

4) Arguably the presence of guest hooks is a detail of the guest OS build, rather than something that should even appear in the runtime configuration.

5) Regardless of how we manage the guest hooks, having some mechanism where we can alter the "inner" config.json from within the guest would be useful.  We'll want something like this for the VFIO / SR-IOV stuff we're working on (to translate host side vfio devices into guest side vfio devices), and I can think of a number of other cases where we want to modify the inner config.json from the outer config.json due to things being renamed by the virtualization layer.

dgibson:
Of course all those thoughts are for the medium to long term.  Short term is another question, which I also don't have any clear idea on.

bpradipt:
> > @amshinde @bpradipt @lifupan the normal OCI container hooks are setup by containerd and kata runs them on the host side. Are you suggesting that kata should set it up and run in the guest instead?
> 
> As @bpradipt This PR is for running the OCI hooks inside the guest. We run hooks passed in the OCI spec on the host, but we added this functionality to support use cases which require certain hooks to be run inside the guest, for eg in case of Nvidia GPU plugin. The hooks need to be part of the guest rootfs. We added this functionality as part of this PR : #393
> 
> But since we made the guest rootfs as read-only, the above functionality was broken.
> If you look at this comment, [#346 (comment)](https://github.com/kata-containers/agent/pull/346#issuecomment-420012768), the config.json needs to be written in the bundle directory which was being done by the agent. This PR attempts to fix the read-only error caused by the agent attempting to write to the bundle directory in the guest by instead having the agent writing the config.json to a directory under /run. With this though the config.json no longer exists in the bundle directory which may break OCI hooks that depend on this.
> My suggestion is to instead have the runtime write the config.json to the shared bundle directory before it is passed as read-only to the guest.

The agent code scans the guest hook directory to look for hooks and then add those to the config.json for execution by libcontainer.  So writing the config.json by the runtime while possible will not solve the purpose since the runtime is not scanning the guest hook directory to detect the presence of hooks. 

Can we take a short term approach to write the config.json to a read-write path (like in this PR) and document it for guest hooks to make required changes.  At least we'll have the guest side hook working. 
In the long term we should think about a proper solution in the context of Kata-2.0 considering the existing issues and also the comments from @dgibson ? 

@bergwolf @amshinde @fidencio @devimc @jodh-intel @lifupan  what are your thoughts ?

dgibson:
I tend to agree with @bpradipt.  Applying his fix makes things better in the short term, and we can address this more thoroughly in Kata 2.   Putting the config.json in a different (read-write) location isn't exactly "standard" - but really very little about how the guest hooks operate is "standard", even though they have the trappings of OCI hooks.

So, let's get it working, since it has real usefulness in certain cases.

bergwolf:
@bpradipt @dgibson Yes, I agree with you. We can fix it like in this PR as a short term fix and look for a long term fix in kata 2.0.

And IMO for a long term fix to work, we should
1. create container bundle path in a tmpfs location
2. bind mount all rootfs and volumes to directories inside the bundle
3. write `config.json` in the bundle

This not only solves the readonly rootfs problem, but also allows us to move away from the currently misused sandbox shared path, which should not be used as container rootfs destinations when shared fs is not used.

jodh-intel:
@bpradipt - I think the DCO check is failing because although you have the magic SOB line, you haven't specified your name, only an email address (which could be anything).

jodh-intel:
@bpradipt - and apologies if you'd missed that out due to looking at:

https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#general-format

That seems to have regressed as it's "lost" the contributors name. I've raised https://github.com/kata-containers/community/issues/163 so we can resolve that...

bpradipt:
> @bpradipt - I think the DCO check is failing because although you have the magic SOB line, you haven't specified your name, only an email address (which could be anything).

Thanks @jodh-intel . I will re-push.

bpradipt:
@jodh-intel no luck :-( ..  Tried both the formats `name <email>` and `name email`

bpradipt:
Finally, the author and SOB details should match

dgibson:
> @bpradipt @dgibson Yes, I agree with you. We can fix it like in this PR as a short term fix and look for a long term fix in kata 2.0.
> 
> And IMO for a long term fix to work, we should
> 
>     1. create container bundle path in a tmpfs location
> 
>     2. bind mount all rootfs and volumes to directories inside the bundle
> 
>     3. write `config.json` in the bundle
> 
> 
> This not only solves the readonly rootfs problem, but also allows us to move away from the currently misused sandbox shared path, which should not be used as container rootfs destinations when shared fs is not used.

@bergwolf, I've filed an issue at https://github.com/kata-containers/kata-containers/issues/485 to track this longer term work.

PR #807, opened 2020-07-17
device: Ease device access for rootfs device to allow node creation

### LINKED ISSUE #806: Fix docker in docker use case
Reference: https://github.com/kata-containers/runtime/issues/2789

amshinde:
For docker in docker scenario, the nested container created
has entry "b *:* m" in the list of devices it is allowed to access
under /sys/fs/cgroup/devices/docker/{ctrid}/devices.list.

This entry was causing issues while starting a nested container
as we were denying "m" access to the rootfs block devices.
With this change we add back "m" access, the container would be
allowed to create a device node for the rootfs device but will
 not have read-write access to the created device node.
This fixes the docker in docker use case while still making sure
the container is not allowed read/write access to the rootfs.
Note, this could also be fixed by simply skipping {"Type : "b"}
while creating the device cgroup with libcontainer.
But this seems to be undocumented behaviour at this point,
hence refrained from taking this approach.

Fixes #806

### DIFF
--- a/device.go
+++ b/device.go
@@ -496,7 +496,7 @@ func updateDeviceCgroupForGuestRootfs(spec *pb.Spec) {
 		Major:  devMajor,
 		Minor:  devMinor,
 		Type:   "b",
-		Access: "rwm",
+		Access: "rw",
 	}
 
 	spec.Linux.Resources.Devices = append(spec.Linux.Resources.Devices, nvdimmCg)


amshinde:
Note: This is how the devices.list looks like for a container:
```

root@fedora /sys/fs/cgroup/devices/docker/kata_979940c96747f87c3b76c7a817e26a43849097686684e80e93777643e60aceb6
--
  | └─ (lap-fedora) # cat devices.list
  | c 1:5 rwm
  | c 1:3 rwm
  | c 1:9 rwm
  | c 1:8 rwm
  | c 5:0 rwm
  | c 5:1 rwm
  | c 10:232 rwm
  | c 10:238 rwm
  | c 10:196 rwm
  | c *:* m
  | b *:* m
  | c 1:7 rwm
  | c 136:* rwm
  | c 5:2 rwm
  | c 10:200 rwm

```

jodh-intel:
Does this need porting to 2.0?

PR #811, opened 2020-07-25
Backport:device: Ease device access for rootfs device to allow node creation

### LINKED ISSUE #806: Fix docker in docker use case

amshinde:
For docker in docker scenario, the nested container created
has entry "b *:* m" in the list of devices it is allowed to access
under /sys/fs/cgroup/devices/docker/{ctrid}/devices.list.

This entry was causing issues while starting a nested container
as we were denying "m" access to the rootfs block devices.
With this change we add back "m" access, the container would be
allowed to create a device node for the rootfs device but will
 not have read-write access to the created device node.
This fixes the docker in docker use case while still making sure
the container is not allowed read/write access to the rootfs.
Note, this could also be fixed by simply skipping {"Type : "b"}
while creating the device cgroup with libcontainer.
But this seems to be undocumented behaviour at this point,
hence refrained from taking this approach.

Fixes #806

(cherry picked from commit 5dc7ae4bebe5413e957bb52387a13fcaa0c6a6e5)

### DIFF
--- a/device.go
+++ b/device.go
@@ -493,7 +493,7 @@ func updateDeviceCgroupForGuestRootfs(spec *pb.Spec) {
 		Major:  devMajor,
 		Minor:  devMinor,
 		Type:   "b",
-		Access: "rwm",
+		Access: "rw",
 	}
 
 	spec.Linux.Resources.Devices = append(spec.Linux.Resources.Devices, nvdimmCg)


jodh-intel:
CI failing with the MSR error:

```
+ docker run -d --name backward_test --runtime=kata-runtime busybox tail -f /dev/null
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
61c5ed1cbdf8: Pulling fs layer
61c5ed1cbdf8: Verifying Checksum
61c5ed1cbdf8: Download complete
61c5ed1cbdf8: Pull complete
Digest: sha256:4f47c01fa91355af2865ac10fef5bf6ec9c7f42ad2321377c21e844427972977
Status: Downloaded newer image for busybox:latest
685c1cdcdb438fab86f8f31a9c8ad965a454adc2a5c85e811bba29fff14730fa
docker: Error response from daemon: OCI runtime create failed: failed to launch qemu: exit status 1, error messages from qemu log: qemu-system-x86_64: error: failed to set MSR 0x48b to 0x1582e00000000
qemu-system-x86_64: /tmp/tmp.QkZxr6AtEW/src/github.com/qemu/qemu/target/i386/kvm.c:2695: kvm_buf_set_msrs: Assertion `ret == cpu->kvm_msr_buf->nmsrs' failed.: unknown.
```

amshinde:
Ok, I am going to rerun and see if I see the failure again. 
@devimc Any clue on this one? Could it be related to the msr s we are setting for running in CI?

jodh-intel:
Same CI failure :sob:

amshinde:
As @devimc pointed, the 1.11 branch seems to be installing 5.0 qemu instead of 4.11. Looks like we are checking out incorrect repo version somewhere. 
@GabyCT is taking a look at this. We should resolve this before we cut the next stable release.
But this PR has passed on master and 1.10 branches and merged there.

PR #817, opened 2020-08-21
oci: Fix running of OCI hooks

bpradipt:
OCI hooks fails to run since the code was writing the config.json
to the read-only path. This patch fixes it

Fixes: #818 

Signed-off-by: Pradipta Kr. Banerjee <pradipta.banerjee@gmail.com>

### DIFF
--- a/grpc.go
+++ b/grpc.go
@@ -676,14 +676,14 @@ func (a *agentGRPC) CreateContainer(ctx context.Context, req *pb.CreateContainer
 		a.sandbox.addGuestHooks(ociSpec)
 
 		// write the OCI spec to a file so that hooks can read it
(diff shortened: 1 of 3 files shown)


jodh-intel:
Thanks @bpradipt. But that "Fixes" issue doesn't look right?

bpradipt:
> Thanks @bpradipt. But that "Fixes" issue doesn't look right?

@jodh-intel you are right. Looks like for the original PR also the issue reference was incorrect. I created a new issue in agent repo and linked this PR to the same. Thanks.

jodh-intel (approved):
Thanks @bpradipt.

fidencio (approved):
LGTM!

Issue #823, opened 2020-08-28
device: Accept the PCIAddress in the BDF format for block devices 


likebreath:
The kata-agent now only accepts the PCIAddress from kata-runtime in the "bridgeAddr/deviceAddr" format, which is tied to the device hotplug interface in qemu. We should extend extend the kata-agent to accept input PCIAddress in the standard BDF format, which should be more generic to support different hypervisors, e.g. cloud-hypervisor.

dgibson:
I don't think this is a good idea.  To be useful as part of the runtime<->agent protocol, any address we use has to be meaningful to both host and guest.  A (guest) BDF address is meaningful to the guest kernel, but in general the host can only guess at it.  The BB part depends on the guest enumeration of the bus.  Even if we count on the guest firmware to enumerate it in a particular way, the bus numbers don't necessarily have to even remain static for the lifetime of the guest.

The domain (DDDD part) has its own set of issues.  On PC-like machines it will generally be consistent, because I think it's used as a parameter in one of the controller registers (although nearly all PC-like machines only use domain 0000 anyway).  However PCI domains *by definition* have no PCI level connection to each other - they're independent host bridges - so there can be no PCI standard way of numbering them.  On other platforms (e.g. IBM POWER) the bridges are identified by other means and the domain numbers are just allocated dynamically by the guest kernel (and on POWER it's common to have many domains).

Although the "bridgeAddr/slotAddr" format may have originated with qemu, it has real meaning in PCI space which both the host and guest can interpret, regardless of hypervisor - if CLH can generate (a guess at) guest BDF addresses, it can certainly generate the bridge and slot addresses (the "D" part of the BDF will be the device slot).

I *would* however suggest extending the bridgeAddr/slotAddr format to a more general "PCI path" by:

- Allowing any number of entries, to allow for both devices plugged directly into the root bus, and for devices under several layers of bridge
- Allowing each component to be SS.F rather than just SS to allow for multifunction devices.
- Adding a (necessarily machine type specific) way of specifying which domain/host bridge to start the search from.

likebreath:
@dgibson Thanks a lot the detailed comments and explanation. It is really a good discussion on defining/improving the agent-runtime protocol of locating (block) devices. Let's bring this topic to the broader users in our community. Please feel free to add more people here.

WDYT? @kata-containers/architecture-committee @amshinde @egernst @fidencio @sboeuf @rbradford @kata-containers/agent

sboeuf:
[...]

PR #824, opened 2020-08-28
device: Accept the PCIAddress in the BDF format for block devices

likebreath:
The kata-agent now only accepts the PCIAddress from kata-runtime in
the "bridgeAddr/deviceAddr" format, which is tied to the device
hotplug interface in qemu. This patch extends the kata-agent to accept
input PCIAddress in the standard BDF format, which should be more
generic to support different hypervisors, e.g. cloud-hypervisor.

Fixes: #823

Depends-on: github.com/kata-containers/runtime#2909

Signed-off-by: Bo Chen <chen.bo@intel.com>

### DIFF
--- a/device.go
+++ b/device.go
@@ -219,15 +219,12 @@ func virtioBlkCCWDeviceHandler(ctx context.Context, device pb.Device, spec *pb.S
 // Here, bridgeAddr is the address at which the brige is attached on the root bus,
 // while deviceAddr is the address at which the device is attached on the bridge.
 func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec, s *sandbox) error {
-	// When "Id (PCIAddr)" is not set, we allow to use the predicted "VmPath" passed from kata-runtime
-	if device.Id != "" {
-		// Get the device node path based on the PCI device address
-		devPath, err := getPCIDeviceName(s, device.Id)
-		if err != nil {
-			return err
-		}
-		device.VmPath = devPath
+	// Get the device node path based on the PCI device address
+	devPath, err := getPCIDeviceName(s, device.Id)
+	if err != nil {
+		return err
 	}
+	device.VmPath = devPath
 
 	return updateSpecDeviceList(device, spec)
 }
--- a/device.go
+++ b/device.go
@@ -174,9 +174,21 @@ func getDeviceName(s *sandbox, devID string) (string, error) {
 }
 
 func getPCIDeviceNameImpl(s *sandbox, pciID string) (string, error) {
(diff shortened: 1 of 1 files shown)


device.go
```diff
@@ -174,9 +174,17 @@ func getDeviceName(s *sandbox, devID string) (string, error) {
...
 
 func getPCIDeviceNameImpl(s *sandbox, pciID string) (string, error) {
-	pciAddr, err := getDevicePCIAddress(pciID)
-	if err != nil {
-		return "", err
+	var pciAddr string
+	var err error
+
+	// Take the input pciID as pciAddr if its format is standard BDF
+	if len(pciID) == 12 && len(strings.Split(pciID, ":")) == 3 {
```
sboeuf, line 181: A b/d/f string contains only 2 `:`, as it contains 1 `.` to separate the `device` from the `function`.

sboeuf (commented):
One small fix regarding the way you detect a b/d/f, otherwise the PR looks great to me.

device.go
devimc: please don't use magic numbers (12, 3), define new constants and document them

likebreath: That's right. With 2 `:`, the b/d/f string is divided into 3 parts, which is what the `if` statement is looking for.

device.go
```diff
@@ -174,9 +174,20 @@ func getDeviceName(s *sandbox, devID string) (string, error) {
 }
 
 func getPCIDeviceNameImpl(s *sandbox, pciID string) (string, error) {
-	pciAddr, err := getDevicePCIAddress(pciID)
-	if err != nil {
-		return "", err
+	var pciAddr string
+	var err error
+
+	const sizeBDFStr = 12
```
bergwolf, line 180: It worths mentioning BDF format `0000:00:00.0` here to help understand the constants.

bergwolf (commented):
Just a small comment. Otherwise lgtm.

jodh-intel (approved):
Thanks @likebreath.

A couple of points:

- Any way to add unit tests for this change?
- Does this need to be ported to 2.0? If so, could you "do the honours"? :smile:

dgibson (commented):
The code itself looks fine, but I think it's taking an ill-considered approach to the actual problem.  More detailed comments on issue #823.

jodh-intel:
@likebreath, @dgibson - please can you summarise that status of this PR? Is it still required?

likebreath:
@jodh-intel Thanks for the reminder. 

As @dgibson pointed out, the proposed approach here (accepting new format in B/D/F) is not the right way to extend runtime/agent protocol. Let's follow up on the discussions in https://github.com/kata-containers/agent/issues/823. Closing the PR now.

Issue #829, opened 2020-08-31
Could not create destination mount point: /etc/resolv.conf


keloyang:
docker can' t start a kata container， error is the fowllong,
```
[root@centos1 ~]# docker run --rm -ti --runtime kata-runtime  busybox ash
docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not create destination mount point: /etc/resolv.conf: open /etc/resolv.conf: read-only file system: unknown.
```
https://github.com/kata-containers/osbuilder/pull/366 have a check for  /etc/resolv.conf and create if it is not exist.
but sometimes, I only  update kata-agent, not the entire rootfs, and don't run osbuilder/rootfs-builder/rootfs.sh again. so   /etc/resolv.conf is not exist in rootfs, at last ,  it causes the container to fail to run.

PR #830, opened 2020-08-31
network: Fix Could not create destination mount point: /etc/resolv.conf

keloyang:
agent does not check if /etc/resolv.conf exists, if it does not exist,
setDns will fail and causes the container to fail to run. sometimes,we
don't need dns in guest vm, so it's better to skip and return nil if
/etc/resolv.conf does not exists.

Fixes: #829

### DIFF
--- a/network.go
+++ b/network.go
@@ -693,6 +693,11 @@ func setupDNS(dns []string) (err error) {
 		return nil
 	}
 
+	if _, err := os.Stat(guestDNSFile); err != nil && os.IsNotExist(err) {
+		agentLog.Errorf("%s is not exist in guest, dns service may be unavailable", guestDNSFile)
+		return nil
+	}
+
 	if err := os.MkdirAll(filepath.Dir(kataGuestSandboxDNSFile), 0700); err != nil {
 		return err
 	}


bergwolf (approved):
Thanks!

jodh-intel (approved):
Thanks @keloyang!

lgtm.

Do we need a similar change in the 2.0 agent @lifupan, @liubin?

lifupan:
> Thanks @keloyang!
> 
> lgtm.
> 
> Do we need a similar change in the 2.0 agent @lifupan, @liubin?

Hmm, it seemed rust-agent hadn't support the guest DNS, I had opened an issue to track this: https://github.com/kata-containers/kata-containers/issues/689

network.go
```diff
@@ -693,6 +693,11 @@ func setupDNS(dns []string) (err error) {
 		return nil
 	}
 
+	if _, err := os.Stat(guestDNSFile); err != nil && os.IsNotExist(err) {
```
amshinde, line 696: Guest DNS is setup only if the grpc request contains dns bits asking for guest dns to be setup. I feel in this case. since an explicit request has been made to setup dns, we should be creating the /etc/resolve.conf file so that `kataGuestSandboxDNSFile` can be bind-mounted to it. While `/etc/resolve.conf` file  exists by default on most distros, this file is absent on Clear Linux distribution as it is stateless. I have a feeling the original change to set up dns was not tested a Clear Linux based root file system.

cc @bergwolf @lifupan

amshinde (changes requested):

network.go
lifupan: IMHO, most of users wouldn't care about the guest DNS setting, and we wouldn't want this setting error disturbing those users who didn't care about it.  For those users who did need the guest DNS setting, they should make sure the "/etc/resolve.conf" file exists in their custom guest image.

amshinde: Fair enough. This is something we could handle while generating the clearlinux rootfs.

amshinde (approved):

Issue #837, opened 2020-09-14
stop kata containers failed if the container share sandbox container's pid namespace


keloyang:
we can reproduce it like this,
1. create a sandbox and a container.
2. stop the container in another console.

untrusted-podsandbox-config.json and container-config.yaml ,
```
[root@centos1 kata-containerd]# cat examples/untrusted-podsandbox-config.json 
{
  "metadata": {
    "name": "busybox-sandbox",
    "namespace": "default",
    "attempt": 1,
    "uid": "hdishd83djaidwnduwk28bcsb"
  },
  "annotations": {
    "io.kubernetes.cri.untrusted-workload": "true"
  },
  "log_directory": "/tmp",
  "linux": {
  }
}
[root@centos1 kata-containerd]# cat examples/container-config.yaml 
metadata:
  name: busybox
image:
  image: dockerhub.jd.com/jrysk/stress-1
command:
- sleep
- 16881688
log_path: busybox.0.log
mounts:
- container_path: /home
  host_path: /root
linux:
  security_context:
    namespace_options:
      pid: 2 
```

```
[root@centos1 kata-containerd]# crictl runp untrusted-podsandbox-config.json
ed9357f92197eb6e745a73af6217820e9987d3548ea58491144a3c215a615ebc
[root@centos1 kata-containerd]# crictl create ed9357f92197eb6e745a73af6217820e9987d3548ea58491144a3c215a615ebc container-config.yaml untrusted-podsandbox-config.json      
4d2a8f214d722311459d299fe1aea3508b432fa02651383bb0ad6feace8da584
crictl start 4d2a8f214d722311459d299fe1aea3508b432fa02651383bb0ad6feace8da584
4d2a8f214d722311459d299fe1aea3508b432fa02651383bb0ad6feace8da584
[root@centos1 kata-containerd]# crictl exec -ti  4d2a8f214d722311459d299fe1aea3508b432fa02651383bb0ad6feace8da584 sh
```
**don't exit sh and in another console, run crictl stop.**
```
[root@centos1 kata-containerd]# crictl stop 4d2a8f214d722311459d299fe1aea3508b432fa02651383bb0ad6feace8da584
[...]

PR #838, opened 2020-09-14
libcontainer: kill all processes in container' cgroup

keloyang:
reproduce:
cid=$(crictl runp examples/untrusted-podsandbox-config.json)
crictl create  $cid examples/container-config.yaml examples/untrusted-podsandbox-config.json
crictl start $cid
crictl exec -ti  $cid sh

**Don't exit sh and in another console, run crictl stop.**
crictl stop $cid

Fixes: #837.

the root cause is that container exit , but it don't kill the processes in its cgroup, so the cgroup path is busy, and the cgroup path
cannot be deleted.

runc has fix the issue, details for opencontainers/runc#2085

**go dep and go modules are not compatible**

To import v2 or a version after v2 of a dependency package, go module need to add /v2(or a higher version) to the path of import path. take runc for example,
```
require (
        github.com/checkpoint-restore/go-criu/v4 v4.1.0
		...
```
this means runc need github.com/checkpoint-restore/go-criu, and the version of go-criu is v4.1.0, runc import go-criu like this,
```
criu "github.com/checkpoint-restore/go-criu/v4/rpc"
```
if use dep to update vendor fo agent, agent need runc，and runc need go-criu, so dep will try to get github.com/checkpoint-restore/go-criu/v4 
but it is not exist. dep will fail. So we can't use dep to update vendor anymore.

so I change runc code directly in the vendor dir.

### DIFF
--- a/vendor/github.com/opencontainers/runc/libcontainer/state_linux.go
+++ b/vendor/github.com/opencontainers/runc/libcontainer/state_linux.go
@@ -38,7 +38,8 @@ type containerState interface {
 }
 
 func destroy(c *linuxContainer) error {
-	if !c.config.Namespaces.Contains(configs.NEWPID) {
+	if !c.config.Namespaces.Contains(configs.NEWPID) ||
+		c.config.Namespaces.PathOf(configs.NEWPID) != "" {
 		if err := signalAllProcesses(c.cgroupManager, unix.SIGKILL); err != nil {
 			logrus.Warn(err)
 		}


fidencio:
@keloyang, the right way to approach this will be updating the vendored code here: https://github.com/kata-containers/agent/blob/master/Gopkg.toml and then running `dep ensure -update`.

This will generate the changes we need and we can, then, have that added to a commit then opened as a PR.

Please, would you mind to do that and force-push that into this very same branch?

fidencio:
> @keloyang, the right way to approach this will be updating the vendored code here: https://github.com/kata-containers/agent/blob/master/Gopkg.toml and then running `dep ensure -update`.
> 
> This will generate the changes we need and we can, then, have that added to a commit then opened as a PR.
> 
> Please, would you mind to do that and force-push that into this very same branch?

Ouch, now I see the whole explanation on your PR body. :-/
@kata-agent, opinions?

fidencio:
@bergwolf, do you think this one is needed on the 2.0-dev branch?

fidencio:
@keloyang, in any case, @egernst re-worked the runtime part in order to, at least for now, **not** apply cgroups within the VM.
So, for now, this issue is fixed / workedaround.

vendor/github.com/opencontainers/runc/libcontainer/state_linux.go
```diff
@@ -38,7 +38,8 @@ type containerState interface {
 }
 
 func destroy(c *linuxContainer) error {
-	if !c.config.Namespaces.Contains(configs.NEWPID) {
+	if !c.config.Namespaces.Contains(configs.NEWPID) ||
+		c.config.Namespaces.PathOf(configs.NEWPID) != "" {
```
bergwolf, line 42: Please do not modify vendor directory directly. The change will be lost next time we update go vendor. If you believe something is broken, please fix it in the original repository (`runc` in this case) and update kata's vendor.conf instead.

bergwolf (changes requested):

bergwolf:
Ok, I see the reason behind the direct vendor update. However, it is still not acceptable to just modify vendor code directly. Next time we run `dep ensure`, all the changes will be lost.

If `dep` is so broken as shown here, we should consider migrating to `go dep` instead.

fidencio:
@bergwolf, you did the whole migration for 2.x repo. How do-able would be to backport / rewrite those patches to the 1.x branch?

jodh-intel (changes requested):
Hi @keloyang - please tal at how to re-vendor a PR:

- https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#re-vendor-prs
- https://github.com/kata-containers/community/blob/master/VENDORING.md

Here's a "concrete" example from the `runtime` repo (where we do a lot of re-vendoring for packages like `govmm`):

- Here's the change to the `govmm` package itself: https://github.com/intel/govmm/pull/128
- And here's how that change is "re-vendored" into the runtime: https://github.com/kata-containers/runtime/pull/2801

keloyang:
@bergwolf @jodh-intel  @fidencio I have a try and use "dep ensure -v" to update vendor before I raising this pr, but it always failed, 
the reason is go dep and go modules are not compatible, to import v2 or a version after v2 of a dependency package, go module need to add /v2(or a higher version) to the import path. take runc for example,
```
require (
        github.com/checkpoint-restore/go-criu/v4 v4.1.0
	...
```
this means runc need github.com/checkpoint-restore/go-criu, and the version of go-criu is v4.1.0, runc import go-criu like this,
```
criu "github.com/checkpoint-restore/go-criu/v4/rpc"
```
if use dep to update vendor fo agent,  dep will try to get github.com/checkpoint-restore/go-criu/v4, but it is not exist, only github.com/checkpoint-restore/go-criu exist, so dep will fail.  

I think we can't use dep to update vendor anymore and use go mod is a choice.

egernst:
> @keloyang, in any case, @egernst re-worked the runtime part in order to, at least for now, **not** apply cgroups within the VM.
> So, for now, this issue is fixed / workedaround.

Actually -- this is still a real issue. My change was specific setting cpuset.cpus/mems in the guest. We do (and should!) set the cgroups still.

What are next steps here for 1.x? Can we confirm this is not an issue in 2.x agent?

fidencio:

PR #845, opened 2020-09-15
Allow VFIO devices to be passed through as VFIO devices

dgibson:
This patches series implements a fix for https://github.com/kata-containers/agent/issues/833.

It's based on pull requests for other issues, specifically https://github.com/kata-containers/agent/pull/782 and https://github.com/kata-containers/agent/pull/836, so those should be merged first.

This series on its own won't do anything without corresponding host-side fixes for https://github.com/kata-containers/runtime/issues/2938 which I'm still polishing for a PR.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -756,8 +756,7 @@ func (s *sandbox) listenToUdevEvents() {
 		fieldLogger.Infof("Received add uevent")
 
 		// Check if device hotplug event results in a device node being created.
-		if uEv.DevName != "" &&
-			(strings.HasPrefix(uEv.DevPath, rootBusPath) || strings.HasPrefix(uEv.DevPath, acpiDevPath)) {
+		if strings.HasPrefix(uEv.DevPath, rootBusPath) || strings.HasPrefix(uEv.DevPath, acpiDevPath) {
 			// Lock is needed to safely read and modify the sysToDevMap and deviceWatchers.
 			// This makes sure that watchers do not access the map while it is being updated.
 			s.Lock()
--- a/agent.go
+++ b/agent.go
@@ -47,6 +47,8 @@ const (
 	bashPath         = "/bin/bash"
 	shPath           = "/bin/sh"
 	debugConsolePath = "/dev/console"
+
+	vfioGroupPath = "/devices/virtual/vfio"
 )
 
 var (
@@ -756,7 +758,9 @@ func (s *sandbox) listenToUdevEvents() {
 		fieldLogger.Infof("Received add uevent")
 
 		// Check if device hotplug event results in a device node being created.
-		if strings.HasPrefix(uEv.DevPath, rootBusPath) || strings.HasPrefix(uEv.DevPath, acpiDevPath) {
+		if strings.HasPrefix(uEv.DevPath, rootBusPath) ||
+			strings.HasPrefix(uEv.DevPath, acpiDevPath) ||
+			strings.HasPrefix(uEv.DevPath, vfioGroupPath) {
 			// Lock is needed to safely read and modify the sysToDevMap and deviceWatchers.
 			// This makes sure that watchers do not access the map while it is being updated.
 			s.Lock()
--- a/agent.go
+++ b/agent.go
@@ -181,7 +181,7 @@ var logsVSockPort = uint32(0)
 var debugConsoleVSockPort = uint32(0)
 
 // Timeout waiting for a device to be hotplugged
-var hotplugTimeout = 3 * time.Second
+var hotplugTimeout = 10 * time.Second
 
 // Specify the log level
 var logLevel = defaultLogLevel
--- a/device.go
+++ b/device.go
@@ -34,7 +34,12 @@ const (
 	driverNvdimmType    = "nvdimm"
 	driverEphemeralType = "ephemeral"
 	driverLocalType     = "local"
-	vmRootfs            = "/"
+
+	//VFIO devices to be bound to the VM's native drivers (this is
+	//Kata specific behaviour, that requires careful configuration
+	//to get something usable inside the container)
+	driverVfioVmType = "vfio-vm"
+	vmRootfs         = "/"
 )
 
 const (
@@ -97,6 +102,7 @@ var deviceHandlerList = map[string]deviceHandler{
 	driverBlkCCWType:  virtioBlkCCWDeviceHandler,
 	driverSCSIType:    virtioSCSIDeviceHandler,
 	driverNvdimmType:  nvdimmDeviceHandler,
+	driverVfioVmType:  vfioDeviceHandler,
 }
 
 func rescanPciBus() error {
@@ -257,6 +263,46 @@ func virtioSCSIDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spe
 
 func nvdimmDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec, s *sandbox, devIdx devIndex) error {
 	return updateSpecDeviceList(device, spec, devIdx)
+}
+
+// device.Options should have one entry for each PCI device in the VFIO group
+// Each option should have the form "DDDD:BB:DD.F=NN/MM"
+//     DDDD:BB:DD.F is the device's PCI address in the host
+//     NN is the PCI slot of the device's bridge, MM is the PCI slot of the device
+func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *sandbox, devIdx devIndex) error {
+	fieldLogger := agentLog.WithField("host-dev", device.ContainerPath)
+
+	for _, opt := range device.Options {
+		tokens := strings.Split(opt, "=")
+		if len(tokens) != 2 {
+			return fmt.Errorf("Malformed VFIO option %q", opt)
+		}
+
+		hostBdf := tokens[0]
+		guestPCIPath := PciPath{tokens[1]}
+
+		sysfsRelPath, err := pciPathToSysfs(guestPCIPath)
+		if err != nil {
+			return err
+		}
+
+		// We don't actually care what the /dev node is called
+		// (and there may not be any), but this will wait for
+		// a uevent that indicates the device is ready
+		_, err = getDeviceName(s, sysfsRelPath)
+		if err != nil {
+			return err
+		}
+
+		tokens = strings.Split(sysfsRelPath, "/")
+		guestBdf := tokens[len(tokens)-1]
+
+		fieldLogger.WithField("host-bdf", hostBdf).WithField("guest-bdf", guestBdf).Debug("VFIO: PCI device ready")
+	}
+
+	fieldLogger.Debug("VFIO: Group complete")
+
+	return nil
 }
 
 // updateSpecDeviceList takes a device description provided by the caller,
--- a/device.go
+++ b/device.go
@@ -210,7 +210,7 @@ func virtioMmioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Sp
 		return fmt.Errorf("Invalid path for virtioMmioBlkDevice")
 	}
 
-	return updateSpecDeviceList(device, spec, devIdx)
+	return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath)
 }
 
 func virtioBlkCCWDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *sandbox, devIdx devIndex) error {
@@ -225,7 +225,7 @@ func virtioBlkCCWDeviceHandler(ctx context.Context, device pb.Device, spec *pb.S
 	}
 
 	device.VmPath = devPath
-	return updateSpecDeviceList(device, spec, devIdx)
+	return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath)
 }
 
 // device.Id should be a PCI path (see type PciPath)
@@ -239,7 +239,7 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec, 
 		device.VmPath = devPath
 	}
 
-	return updateSpecDeviceList(device, spec, devIdx)
+	return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath)
 }
 
 // device.Id should be the SCSI address of the disk in the format "scsiID:lunID"
@@ -251,11 +251,11 @@ func virtioSCSIDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spe
 	}
 	device.VmPath = devPath
 
-	return updateSpecDeviceList(device, spec, devIdx)
+	return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath)
 }
 
 func nvdimmDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec, s *sandbox, devIdx devIndex) error {
-	return updateSpecDeviceList(device, spec, devIdx)
+	return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath)
 }
 
 // Take the guest device with the given DDDD:BB:SS.F PCI address,
@@ -378,18 +378,18 @@ func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *
 	return nil
 }
 
-// updateSpecDeviceList takes a device description provided by the caller,
-// trying to find it on the guest. Once this device has been identified, the
-// "real" information that can be read from inside the VM is used to update
-// the same device in the list of devices provided through the OCI spec.
-// This is needed to update information about minor/major numbers that cannot
-// be predicted from the caller.
-func updateSpecDeviceList(device pb.Device, spec *pb.Spec, devIdx devIndex) error {
-	// If no ContainerPath is provided, we won't be able to match and
-	// update the device in the OCI spec device list. This is an error.
-	if device.ContainerPath == "" {
+// updateSpecDevice updates a device list in the OCI spec to make it
+// include details appropriate for the VM, instead of the host.  It is
+// given the host path to the device (to locate the device in the
+// original OCI spec) and the VM path which it uses to determine the
+// VM major/minor numbers.
+func updateSpecDevice(spec *pb.Spec, devIdx devIndex, hostPath, vmPath string) error {
+	// If no hostPath is provided, we won't be able to match and
+	// update the device in the OCI spec device list. This is an
+	// error.
+	if hostPath == "" {
 		return grpcStatus.Errorf(codes.Internal,
-			"ContainerPath cannot be empty")
+			"Host path cannot be empty")
 	}
 
 	if spec.Linux == nil || len(spec.Linux.Devices) == 0 {
@@ -398,7 +398,7 @@ func updateSpecDeviceList(device pb.Device, spec *pb.Spec, devIdx devIndex) erro
 	}
 
 	stat := syscall.Stat_t{}
-	if err := syscall.Stat(device.VmPath, &stat); err != nil {
+	if err := syscall.Stat(vmPath, &stat); err != nil {
 		return err
 	}
 
@@ -408,21 +408,22 @@ func updateSpecDeviceList(device pb.Device, spec *pb.Spec, devIdx devIndex) erro
 	minor := int64(unix.Minor(dev))
 
 	agentLog.WithFields(logrus.Fields{
-		"device-path":  device.VmPath,
+		"device-path":  vmPath,
 		"device-major": major,
 		"device-minor": minor,
 	}).Info("handling block device")
 
 	// Update the spec
-	idxData, ok := devIdx[device.ContainerPath]
+	idxData, ok := devIdx[hostPath]
 	if !ok {
 		return grpcStatus.Errorf(codes.Internal,
 			"Should have found a matching device %s in the spec",
-			device.ContainerPath)
+			hostPath)
 	}
 
 	agentLog.WithFields(logrus.Fields{
-		"device-path":        device.VmPath,
+		"host-path":          hostPath,
+		"device-path":        vmPath,
 		"host-device-major":  spec.Linux.Devices[idxData.idx].Major,
 		"host-device-minor":  spec.Linux.Devices[idxData.idx].Minor,
 		"guest-device-major": major,
--- a/device.go
+++ b/device.go
@@ -142,6 +142,7 @@ func pciPathToSysfsImpl(pciPath PciPath) (string, error) {
 }
 
 func getDeviceName(s *sandbox, devID string) (string, error) {
+	var found bool
 	var devName string
 	var notifyChan chan string
 
@@ -151,7 +152,7 @@ func getDeviceName(s *sandbox, devID string) (string, error) {
 	s.Lock()
 	for key, value := range s.sysToDevMap {
 		if strings.Contains(key, devID) {
-			devName = value
+			devName, found = value, true
 			fieldLogger.Infof("Device: %s found in device map", devID)
 			break
 		}
@@ -162,13 +163,13 @@ func getDeviceName(s *sandbox, devID string) (string, error) {
 	// The key of the watchers map is the device we are interested in.
 	// Note this is done inside the lock, not to miss any events from the
 	// global udev listener.
-	if devName == "" {
+	if !found {
 		notifyChan = make(chan string, 1)
 		s.deviceWatchers[devID] = notifyChan
 	}
 	s.Unlock()
 
-	if devName == "" {
+	if !found {
 		fieldLogger.Infof("Waiting on channel for device: %s notification", devID)
 		select {
 		case devName = <-notifyChan:
--- a/device.go
+++ b/device.go
@@ -314,6 +314,7 @@ func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *
 	fieldLogger := agentLog.WithField("host-dev", device.ContainerPath)
 
 	var rebindToVfio bool
+	var group string
 
 	if device.Type != driverVfioVmType {
 		rebindToVfio = true
@@ -351,12 +352,28 @@ func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *
 			if err != nil {
 				return err
 			}
+
+			groupPath := filepath.Join(sysPciDevices, guestBdf, "iommu_group")
+			groupLink, err := os.Readlink(groupPath)
+			if err != nil {
+				return fmt.Errorf("Failed to read %s: %s", groupPath, err)
+			}
+
+			if group == "" {
+				group = filepath.Base(groupLink)
+			} else if group != filepath.Base(groupLink) {
+				// If PCI devices associated with the same
+				// VFIO device (and therefore group) in the
+				// host don't end up in the same group in the
+				// guest, something has gone horribly wrong
+				return fmt.Errorf("%s is not in guest IOMMU group %s", guestBdf, group)
+			}
 		}
 
 		fieldLogger.WithField("host-bdf", hostBdf).WithField("guest-bdf", guestBdf).Debug("VFIO: Device complete")
 	}
 
-	fieldLogger.Debug("VFIO: Group complete")
+	fieldLogger.WithField("guest-group", group).Debug("VFIO: Group complete")
 
 	return nil
 }
--- a/device.go
+++ b/device.go
@@ -373,7 +373,17 @@ func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *
 		fieldLogger.WithField("host-bdf", hostBdf).WithField("guest-bdf", guestBdf).Debug("VFIO: Device complete")
 	}
 
-	fieldLogger.WithField("guest-group", group).Debug("VFIO: Group complete")
+	fieldLogger.WithField("guest-dev", device.VmPath).Debug("VFIO: Group complete")
+
+	if rebindToVfio {
+		vmpath, err := getDeviceName(s, filepath.Join("vfio", group))
+		if err != nil {
+			return fmt.Errorf("Couldn't find device for guest VFIO group %s: %s",
+				group, err)
+		}
+		device.VmPath = vmpath
+		return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath, device.VmPath)
+	}
 
 	return nil
 }
--- a/device.go
+++ b/device.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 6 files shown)


fidencio:
Adding a comment here as a reminder to go through this in the next few days.

@devimc, regardless of how many other reviewers will review this one, I'd appreciate your input before merging this series.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+	fieldLogger.Debugf("VFIO: Found PCI device after %v", time.Since(start))
+
+	overridePath := filepath.Join(sysPath, "driver_override")
+	err := ioutil.WriteFile(overridePath, []byte("vfio-pci"), 0200)
+	if err != nil {
+		return fmt.Errorf("Failed to write %s: %s", overridePath, err)
+	}
+
+	driverPath := filepath.Join(sysPath, "driver")
+	driver, err := os.Readlink(driverPath)
```
c3d, line 287: This is reading the symbolic link, so I'd expect you would get something like '../../../bus/pci/drivers/xhci_hcd' at this stage (took that from a USB)

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+	if os.IsNotExist(err) {
+		// This just means the device isn't bound to any driver
+		driver = ""
+	} else if err != nil {
+		return fmt.Errorf("Failed to readlink %s: %s", driverPath, err)
+	}
+
+	fieldLogger.Debugf("VFIO: Device previously bound to '%s'", driver)
+
+	if driver != "vfio-pci" && driver != "" {
```
c3d, line 297: … but here, you are treating it as if the link only contained `vfio-pci`. Should it _end with_ `vfio-pci`?

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+		fieldLogger.Debugf("VFIO: Unbinding from %s", driver)
```
c3d, line 298: I find it slightly surprising to `unbind` from a function called `Rebind`. Even if I believe that I understand what you are doing, it may be useful to explain that you are just unbinding the driver so that you can rebind below.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+		unbindPath := filepath.Join(sysPath, bdf, "driver", "unbind")
+		err = ioutil.WriteFile(unbindPath, []byte(bdf), 0200)
+		if err != nil {
+			return fmt.Errorf("Failed to write %s: %s:", unbindPath, err)
+		}
+	}
+
+	err = ioutil.WriteFile(pciProbeFile, []byte(bdf), 0200)
```
c3d, line 306: Shouldn't you do a sanity check after that, or is the PCI probe instantaneous?

(I would have expected another loop here waiting for the device to show up)

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+	return nil
+}
+
+func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *sandbox, devIdx devIndex) error {
+	fieldLogger := agentLog.WithField("host-dev", device.ContainerPath)
+
+	var group string
+
+	for _, opt := range device.Options {
+		l := strings.Split(opt, "=")
```
c3d, line 321: You'll probably get complaints for the name `l`… Some other code in this file uses "tokens".

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+		if len(l) != 2 {
+			return fmt.Errorf("Malformed VFIO option \"%s\"", opt)
```
c3d, line 323: Add how to fix it, e.g. `expected hostBDF=guestBridge/guestDeviceID`.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+	err = ioutil.WriteFile(pciProbeFile, []byte(bdf), 0200)
+	if err != nil {
+		return fmt.Errorf("Failed to write %s: %s", pciProbeFile, err)
+	}
+
+	fieldLogger.Info("VFIO: Rebind complete")
+	return nil
+}
+
+func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *sandbox, devIdx devIndex) error {
```
c3d, line 315: Please document this function a little better, in particular what it expects on input. I'm especially interested in the split of responsibilities between whoever generates the spec, the runtime and the agent, in order who knows what about the guest PCI devices, PCI bridges, IOMMU groups, etc.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+
+	var group string
+
+	for _, opt := range device.Options {
+		l := strings.Split(opt, "=")
+		if len(l) != 2 {
+			return fmt.Errorf("Malformed VFIO option \"%s\"", opt)
+		}
+
+		hostBdf := l[0]
```
c3d, line 326: I would probably explain the format of the BDF you expect here, and maybe validate it a little (even though it is validated indirectly by reading `/sys` later).

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+		guestPath := l[1]
```
c3d, line 327: This is not a file path, right? Maybe `guestPCIPath`?

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+
+		groupPath := filepath.Join(sysBusPrefix, guestBdf, "iommu_group")
+		groupLink, err := os.Readlink(groupPath)
+		if err != nil {
+			return fmt.Errorf("Failed to read %s: %s", groupPath, err)
+		}
+
+		if group == "" {
+			group = filepath.Base(groupLink)
+		} else if group != filepath.Base(groupLink) {
```
c3d, line 352: Here, unlike for the `driver` case above, you are using the `Base`, which seems correct.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+			group, err)
+	}
+	device.VmPath = vmpath
+
+	fieldLogger.WithField("guest-dev", device.VmPath).Debug("VFIO: Group complete")
+
+	return updateSpecDevice(spec, devIdx, device.ContainerPath, device.VmPath, device.VmPath)
+}
+
+func VfioTranslateEnv(s *sandbox, env []string) []string {
```
c3d, line 374: Ah, this is where the magic happens. Now I finally get it. Smart.

I would add a comment showing the kind of inputs that you expects. Basically, something like "the container receives information about the PCI devices through environment variables that look like `PCIDEVICE_ 0000:00:14.0=...` etc, and explain the replacement you are doing.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+	newEnv := make([]string, 0, len(env))
+
+	for _, e := range env {
+		if strings.HasPrefix(e, "PCIDEVICE_") {
```
c3d, line 378: I would make `PCIDEVICE_` configurable, although it currently seems pretty hard-coded in the rest of the stack. That function should probably take that as input. Ask around to see if this logic would not apply to other env variables.

device.go
c3d: I believe that function `VfioTranslateEnv` has no unit test.

c3d (changes requested):
Looks good to me, with possibly one bug (ReadLink of the `driver` entry not taking the base path of the result). So waiting for your input on this.

agent.go
```diff
@@ -47,6 +47,8 @@ const (
 	bashPath         = "/bin/bash"
 	shPath           = "/bin/sh"
 	debugConsolePath = "/dev/console"
+
+	vfioGroupPath = "/devices/virtual/vfio"
```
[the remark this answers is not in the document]
dgibson: Yeah, that's a problem for another day.

device.go
```diff
@@ -34,16 +35,20 @@ const (
...
-)
 
-const (
-	pciBusMode = 0220
+	// Time to wait for hotplug of VFIO devices.  SHPC and PCI-E
+	// hotplug protocols include a 5s delay, so this can take a
+	// while
+	vfioTimeout = 10 * time.Second
+
+	vfioHotplugPollInterval = 20 * time.Millisecond
```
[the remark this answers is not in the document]
dgibson: Just seemed like a reasonable compromise between not polling too often, but not having too much latency after the device arrives.

device.go
dgibson: Good point, I've added a line to chop this down to the base name.

device.go
dgibson: > I find it slightly surprising to `unbind` from a function called `Rebind`

Um.. why?  It's a re-bind precisely because we're unbinding from the old thing then binding to the new thing.

device.go
dgibson: At this point we don't care that the probe is complete, just that it's underway.

The later patch "Create device nodes for VFIO devices" adds a call to getDeviceName() which effectively loops waiting for the udev event showing that `/dev/vfio/NN` has appeared, which is what we actually need to wait for here.

device.go
dgibson: Oh, for goodness sake.  It's a local, short is good.

device.go
dgibson: I don't see that that's helpful - if this error appears, it indicates a bug in the runtime, not something user serviceable.

device.go
dgibson: Um... the same as every other *Handler function...

I don't think it's reasonable to expect me to fix every underdocumented piece of the surrounding code before adding the new stuff.

device.go
dgibson: Right.

device.go
dgibson: Hmm.  Configurable where, though?

device.go
dgibson: Hrm, BDF is a pretty standard concept in PCI, and it's used in other places in the code.  I really can't think of what to explain that's useful without being verbose overkill.

I don't think validation makes sense here, because we don't actually care about the internal structure of this value.  All we care is that it's the thing we need to replace in the `PCIDEVICE_` environment variables.

device.go
dgibson: Fair call, adjusted.

device.go
dgibson: But, fine, whatever, I've renamed it to `tokens`.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+		if err == nil {
+			break
+		}
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("Error accessing %s: %s", sysPath, err)
+		}
+		if elapsed := time.Since(start); elapsed > vfioTimeout {
+			return fmt.Errorf("Timeout waiting for %s (%v)", sysPath, elapsed)
+		}
+		time.Sleep(vfioHotplugPollInterval)
```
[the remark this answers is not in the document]
dgibson: I added a comment with some times - they're from my laptop, so I don't know how much they'll vary on other systems.  It's about 6.2-6.3s with SHPC, which is slower than I expected - I know there's a 5s delay in the protocol, but I'm surprised that the rest of everything takes a whole extra second.  With PCI-E (hotplug_vfio_on_root_bus=true) it's much faster around 100-120ms.  The latter might suggest reducing that delay granularity.

Adding jitter seems like serious overkill to me: it's hard to imagine multiple containers being in lockstep to a precision that matters with an interval of 20ms.  Plus it's only a stat() in the loop, not anything that hits the kernel particularly hard.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
-// the same device in the list of devices provided through the OCI spec.
-// This is needed to update information about minor/major numbers that cannot
-// be predicted from the caller.
-func updateSpecDeviceList(device pb.Device, spec *pb.Spec) error {
-	// If no ContainerPath is provided, we won't be able to match and
-	// update the device in the OCI spec device list. This is an error.
-	if device.ContainerPath == "" {
+func vfioRebindDevice(bdf string) error {
+	fieldLogger := agentLog.WithField("guest-bdf", bdf)
+	fieldLogger.Debug("VFIO: Attempting to rebind")
```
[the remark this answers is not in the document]
dgibson: Ok, I've raised this to an Info message and added the timeout.  There's already an Info message at the end of the function.

agent.go
```diff
@@ -146,6 +148,7 @@ type sandbox struct {
 	storages          map[string]*sandboxStorage
 	stopServer        chan struct{}
 	oomEvents         chan string
+	vfioDeviceMap     map[string]string
```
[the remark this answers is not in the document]
dgibson: I've added a comment.  As I think we discussed on slack, this map is purely host to guest BDF mappings.  I think you're confusing it with the sysToDevMap, which is different.

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
-// trying to find it on the guest. Once this device has been identified, the
-// "real" information that can be read from inside the VM is used to update
-// the same device in the list of devices provided through the OCI spec.
-// This is needed to update information about minor/major numbers that cannot
-// be predicted from the caller.
-func updateSpecDeviceList(device pb.Device, spec *pb.Spec) error {
-	// If no ContainerPath is provided, we won't be able to match and
-	// update the device in the OCI spec device list. This is an error.
-	if device.ContainerPath == "" {
+func vfioRebindDevice(bdf string) error {
```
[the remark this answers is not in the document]
dgibson: Done.

dgibson:
New push fixes some of Christoph's comments (others I still need to look at).

I now have a first draft of the kata-runtime changes which go with this at https://github.com/kata-containers/runtime/pull/2963

device.go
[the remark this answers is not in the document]
c3d: That's what I thought. Had I seen "16.67ms", I would have said "That guy works in the video industry". Random numbers that aren't.

device.go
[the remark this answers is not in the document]
c3d: If there is any host-side wait on that `stat()` for whatever reason, then your thousand containers will align themselves in lockstep quite naturally.

That `stat()` is not running on bare metal, but in a VM that can get suspended or be in I/O wait for any number of reasons. For all we know, your entire storage might be running off a SAN with 40s worth of I/Os backed up. It finally completes that specific I/O that was blocking your VMs, and then your 1000 containers get woken up at the same time, pile their 1000 stat(), everyone goes waiting again, and we start getting complaints about slow container start time.

Yes, it's probably overkill, but it's also so simple to avoid. Certainly not enough for me to block that PR, but clearly a small improvement that can avoid subtle bugs later.

device.go
c3d: That is obvious _to you_. Again, you are not talking to VFIO specialists here. You are talking to Go programmers with a focus on the higher-levels in the stack. Know your audience.

So my feeling here is that if I find it surprising, other readers will, and you can cancel that surprise with a comment like:

```
// If there is a driver already, we must unbind it before rebinding below
```

Nothing to you, quite helpful to readers less experienced than you.

device.go
c3d: That makes sense, and is worth explaining. I suggest a comment like:

```
// At this point, we don't care that the probe is complete, just that it's undersay
// We will later wait in getDeviceName() for an event to show up under /dev/vfio/NN
```

and then the code is perfectly understandable by a non-VFIO specialist.

device.go
c3d: Ah, I see. If that's the case, then I would say so.

```
Malformed VFIO option '%s' received from runtime
```

That way, you are pointing the finger in the right direction. Maybe also add a comment here indicating where the runtime is constructing these, so that someone not intimately familiar with the code can quickly know where too look.

c3d: Also, what prevents a user from specifying a device option directly?

device.go
c3d: No, I don't think it's the same thing at all. Here, you are constructing a host to guest mapping that does not exist for the other cases. This difference is highlighted by the fact that the `sandbox` struct contains two maps dedicated to your specific algorithm, `vfioDeviceMap` and `sysToDevMap`, which are not needed (or only marginally) for other handlers.

As I indicated in my comment, I'm not really interested in what goes in what parameter or whatnot, but really more in the overall flow of data. The way I understand it, we have the following, which is so thoroughly unfamiliar to anybody who did not focus on the problem for one month that I am more likely than not to get it wrong:

- Some external entity like a high-level network management service assigns a VF to this container

- This is passed down in the manifest through an envrionment variable in the form `PCIDEVICE_foo=bar`, where `foo` and `bar` are something I already entirely forgot. But basically the workload know `foo`, and from there can get `bar` which I believe is generally speaking a `/dev/` entry.

- Since they are host BDFs, they are woefully inadequate for use in the guest. So here comes your magic.

- Underneath our VM is qemu. qemu is the one that knows what the host BDF will be mapped to in the guest. So we query qemu to get that information. Who does? I guess it's the runtime sending some qmp requests, though I have not fully reviewed that code yet.

- Well, at that stage, qemu has no idea about `/dev` entries in the guest. All it knows is hardware-level stuff, like PCI layout. So the best we get is a PCI B:D.F. The agent will have to deal with figuring out the `/dev` entry from that.

- The runtime then decorates the OCI spec with additional information about host to guest mappings. That comes in the form of device options that look like `B:D.F=guestBridge/guestDeviceID`. (Not sure about other options…)

- The agent then rebinds the device using the guestDeviceID to look it up under guest `/sys`

- The agent then waits for the device to show up as `/dev/vfio/NN` listening to udev event.

- When that happened, it rewrites the environment, specifically the `PCIDEVICE_foo=bar` variables so that they now correctly point to guest entities.

Is that a good summary? If so, could you convert that into a comment, with appropriate corrections, and with correct function names in the runtime and agent?

device.go
c3d: Thanks 😄

device.go
c3d: BDF is standard in PCI jargon, but not necessarily for this audience.

Generally, ACRONYMs in code are OBNOXIOUS, where ACRONYM is, obviously, pretty standard among l33t for Alphabetical Code for Remembering Odd Names You Make up, and OBNOXIOUS is similarly pretty standard (among French developers with a knack for non sense) for Obfuscated Blurbs Not Ordinarily Crossing/Intersecting Our Usual Standards.

In that specific case, though, I'm really interested in specifying that we expect hostBDF to be in the form `B:D.F` and matching what we find in the `PCIDEVICE_...` environment variable, and `guestPath` to be in the form `GuestBridge/GuestDeviceID` as matching entries in guest `/sys/bus/pici/devices`.

Regarding validation, I agree with you it can't be done easily that early. However, that made me think that we will need (in a later commit, no need to amend this PR) to add a validation that the loop rewriting the environment actually finds matching `PCIDEVICE_` environment variables. If not, then things will fail, and I think we should at least `Warn` of not `Error` out.

device.go
c3d: Just to make it clear: that would have to happen later, not in this PR. Just a note to do it later.

The agent has a `config.go` that lists some agent options. That would be a good starting point. I would make it a list of prefixes for the names of environment variables to rewrite, default to `[ "PCIDEVICE_" ]`.

Also, this is something that you might want to be able to override with annotations. See https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/annotations/annotations.go#L233. I have not checked if agent annotations work well though.

c3d:
> New push fixes some of Christoph's comments (others I still need to look at).
> 
> I now have a first draft of the kata-runtime changes which go with this at [kata-containers/runtime#2963](https://github.com/kata-containers/runtime/pull/2963)

Cool, I will look at it. I will also update the comment I made regarding the runtime part with your link.

device.go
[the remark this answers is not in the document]
dgibson: It's a stat() on a guest kernel virtual file system, so, no, a host-side IO delay wouldn't synchronize it.  I really think adding jitter would just be premature complexity.

But... in any case, I'm looking at replacing this path with something looking for a udev event, which will make this moot.

device.go
dgibson: Eh, sure, fine.

device.go
dgibson: Well, the error format here is consistent with that for all the other device types, which likewise validate they got the information they expected from the runtime.

> Also, what prevents a user from specifying a device option directly?

There's no mechanism to do so.  The device list we're working through here isn't the list of devices from the OCI spec - it's a Kata internal device list created by kata-runtime and passed via grpc to kata-agent.  The Options, despite the name, is, in all cases, generated whole cloth by the runtime, it doesn't come from user input.

device.go
dgibson: > No, I don't think it's the same thing at all. Here, you are constructing a host to guest mapping that does not exist for the other cases. This difference is highlighted by the fact that the `sandbox` struct contains two maps dedicated to your specific algorithm, `vfioDeviceMap` and `sysToDevMap`, which are not needed (or only marginally) for other handlers.

Sorry, I was meaning specifically w.r.t. "what it expects on input".  It's true this is rather more complex than the other handlers.  `sysToDevMap` is absolutely used by the other handlers, to exactly the same degree it is here.

> As I indicated in my comment, I'm not really interested in what goes in what parameter or whatnot, but really more in the overall flow of data. The way I understand it, we have the following, which is so thoroughly unfamiliar to anybody who did not focus on the problem for one month that I am more likely than not to get it wrong:
> 
>     * Some external entity like a high-level network management service assigns a VF to this container
> 
>     * This is passed down in the manifest through an envrionment variable in the form `PCIDEVICE_foo=bar`, where `foo` and `bar` are something I already entirely forgot. But basically the workload know `foo`, and from there can get `bar` which I believe is generally speaking a `/dev/` entry.

In part.  The environment variable is set in order to inform the container payload of which device to use, it's not really intended to have meaning to the container runtime itself - but we need to thinking about it because of the address change.  For the benefit of the container runtime, the OCI spec includes both the VFIO control device `/dev/vfio/vfio` and one or more VFIO group devices `/dev/vfio/$NUMBER` in the linux/devices section of the config.json.

>     * Since they are host BDFs, they are woefully inadequate for use in the guest. So here comes your magic.
> 
>     * Underneath our VM is qemu. qemu is the one that knows what the host BDF will be mapped to in the guest. So we query qemu to get that information. Who does? I guess it's the runtime sending some qmp requests, though I have not fully reviewed that code yet. (Update: See proposed changes here: [kata-containers/runtime#2963](https://github.com/kata-containers/runtime/pull/2963))

Actually qemu doesn't know the guest BDF either, because it depends on the guest OS's (or firmware's) enumeration of the PCI bus.  PCI domains have their own cans of worms, though we can sidestep that for now since we're only considering PC-like machines which only use a single domain / host bridge.

What qemu *does* know is the PCI slot & function of the device, and each PCI bridge used to get to it.  We can use those to construct a "PCI path" which is meaningful to both host and guest.  The "bridgeSlot/deviceSlot" format is basically this - just abbreviated because we happen to know that we always place the devices under exactly one bridge (I do hope to generalize this at some point, but that's Kata2 work).

So, kata-runtime queries the slot for the device and its bridge from qemu, and what it sends to the agent is a mapping from host BDF to guest "PCI path".

>     * Well, at that stage, qemu has no idea about `/dev` entries in the guest. All it knows is hardware-level stuff, like PCI layout. So the best we get is a PCI B:D.F. The agent will have to deal with figuring out the `/dev` entry from that.

As above, not even a PCI B:D.F, only a PCI path.

>     * The runtime then decorates the OCI spec with additional information about host to guest mappings. That comes in the form of device options that look like `B:D.F=guestBridge/guestDeviceID`. (Not sure about other options…)

It's not really decorating the OCI spec, this is additional information passed in parallel with the OCI spec when we make the CreateContainer grpc call to the agent.

>     * The agent then rebinds the device using the guestDeviceID to look it up under guest `/sys`

First the agent has to resolve the PCI paths to guest side BDFs  Then it can rebind the device via `/sys`.

>     * The agent then waits for the device to show up as `/dev/vfio/NN` listening to udev event.

Yes.
>     * When that happened, it rewrites the environment, specifically the `PCIDEVICE_foo=bar` variables so that they now correctly point to guest entities.

Yes.

> Is that a good summary? If so, could you convert that into a comment, with appropriate corrections, and with correct function names in the runtime and agent?

Ok in outline, some errors in details.  I'll see what I can do about putting it into comment form.

device.go
dgibson: > BDF is standard in PCI jargon, but not necessarily for this audience.

Hm, maybe.  I mean, I'd prefer to call it "PCI address" which is what it is.  Unfortunately the Kata code badly abuses "pci address" to mean several different things in different places.  "BDF" was the most succinct way I could think of to make it clear that we're talking a normal DDDD:BB:SS.F style PCI address.

> Generally, ACRONYMs in code are OBNOXIOUS, where ACRONYM is, obviously, pretty standard among l33t for Alphabetical Code for Remembering Odd Names You Make up, and OBNOXIOUS is similarly pretty standard (among French developers with a knack for non sense) for Obfuscated Blurbs Not Ordinarily Crossing/Intersecting Our Usual Standards.

As a general principle I absolutely disagree.  Appropriate use of jargon - including acryonyms - is essential to making code readable without verbosity (which itself diminishes readability).  Now, what jargon is appropriate for the context is a non-trivial question, I'll grant you.

> In that specific case, though, I'm really interested in specifying that we expect hostBDF to be in the form `B:D.F` and matching what we find in the `PCIDEVICE_...` environment variable, and `guestPath` to be in the form `GuestBridge/GuestDeviceID` as matching entries in guest `/sys/bus/pici/devices`.
> 
> Regarding validation, I agree with you it can't be done easily that early. However, that made me think that we will need (in a later commit, no need to amend this PR) to add a validation that the loop rewriting the environment actually finds matching `PCIDEVICE_` environment variables. If not, then things will fail, and I think we should at least `Warn` of not `Error` out.

Warn.. maybe.  We absolutely shouldn't fail on this.  It is entirely possible to have a working setup without any PCIDEVICE variables, and in fact it's quite likely if someone is creating a container manually (e.g. with podman) rather than with k8s.  For such ad-hoc cases you can either manually set the right address in the payload, or have scripts inside the payload to locate the device via heuristics like "find the only VFIO-accessible PCI device of the type(s) I care about" (my test containers for development do exactly this).

agent.go
```diff
@@ -757,13 +763,15 @@ func (s *sandbox) listenToUdevEvents() {
 
 		// Check if device hotplug event results in a device node being created.
 		if uEv.DevName != "" &&
-			(strings.HasPrefix(uEv.DevPath, rootBusPath) || strings.HasPrefix(uEv.DevPath, acpiDevPath)) {
-			// Lock is needed to safey read and modify the pciDeviceMap and deviceWatchers.
+			(strings.HasPrefix(uEv.DevPath, rootBusPath) ||
+				strings.HasPrefix(uEv.DevPath, acpiDevPath) ||
+				strings.HasPrefix(uEv.DevPath, vfioGroupPath)) {
+			// Lock is needed to safey read and modify the sysToDevMap and deviceWatchers.
```
bpradipt, line 769: nit: existing typo `safey` which can be corrected in this series

device.go
```diff
@@ -229,37 +230,182 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
-// trying to find it on the guest. Once this device has been identified, the
-// "real" information that can be read from inside the VM is used to update
-// the same device in the list of devices provided through the OCI spec.
-// This is needed to update information about minor/major numbers that cannot
-// be predicted from the caller.
-func updateSpecDeviceList(device pb.Device, spec *pb.Spec) error {
-	// If no ContainerPath is provided, we won't be able to match and
-	// update the device in the OCI spec device list. This is an error.
-	if device.ContainerPath == "" {
+// Take the guest device with the given DDDD:BB:SS.F PCI address,
```
bpradipt, line 252: Should the comment refer to "DDDD:BB:DD.F" ? (Domain:Bus:Dev.Func )

device.go
bpradipt: The environment variable is actually specific to the Kubernetes device plugin being used. For example the Intel SRIOV plugin inserts `PCIDEVICE_` (https://github.com/intel/sriov-network-device-plugin). Same with OpenShift SRIOV device plugin (https://github.com/openshift/sriov-network-device-plugin).
I don't think there is any standardization. 
Can this environment variable prefix be a runtime config in `configuration.toml` ?

device.go
```diff
@@ -229,37 +230,182 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+// include details appropriate for the VM, instead of the host.  It is
+// given the host path to the device (to locate the device in the
+// original OCI spec), the VM path which it uses to determine the VM
+// major/minor number, and the final path with which to present the
+// device in the (inner) container
+func updateSpecDevice(spec *pb.Spec, devIdx devIndex, hostPath, vmPath, finalPath string) error {
+	// If no hostPath is provided, we won't be able to match and
+	// update the device in the OCI spec device list. This is an
+	// error.
+	if hostPath == "" {
```
bpradipt, line 406: Should there be a check for empty `vmPath` as well?

device.go
```diff
@@ -229,37 +231,174 @@ func virtioBlkDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec,
...
+	// the waiting logic in getDeviceName here, because the udev
+	// devices will only appear as we expect *after* we've
+	// rebound.
+	start := time.Now()
+	for {
+		_, err := os.Stat(sysPath)
+		if err == nil {
+			break
+		}
+		if !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
dgibson: I've now implemented a uevent based wait here, though it will show up on issue #781, since I'm moving some of this series there.

device.go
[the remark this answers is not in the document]
dgibson: uevent based wait now implemented

dgibson:
New update is a substantial rework, rebased on #850

device.go
c3d: I guess `S` stands for slot? `D` is already taken by "Domain", so having it twice would be ambiguous.

device.go
```diff
@@ -314,6 +314,7 @@ func vfioDeviceHandler(ctx context.Context, device pb.Device, spec *pb.Spec, s *
 	fieldLogger := agentLog.WithField("host-dev", device.ContainerPath)
 
 	var rebindToVfio bool
+	var group string
```
c3d, line 317: Apparently, github has no good way to comment on the commit message.

>other, they most appear as the same IOMMU group in the guest as well.

Typo: s/most/must/.

device.go
c3d: If it's empty, it will fail on the `syscall.Stat()` below with a somewhat sensible error message, so I would not create a special case for that.

device.go
```diff
@@ -398,7 +398,7 @@ func updateSpecDeviceList(device pb.Device, spec *pb.Spec, devIdx devIndex) erro
 	}
 
 	stat := syscall.Stat_t{}
-	if err := syscall.Stat(device.VmPath, &stat); err != nil {
+	if err := syscall.Stat(vmPath, &stat); err != nil {
 		return err
```
c3d, line 413: I wonder if it would be useful to wrap this error in some context? Something like

```
err = grpcStatus.Errorf(codes.Internal, "Invalid device VM path %s: %s", vmPath, err)
```

c3d (approved):
I did not see any serious functional issue, so approving.

A few minor suggestions, like a typo in a commit message, comments, etc.

fidencio:

PR #849, opened 2020-09-22
Consistent device address matching between `getDeviceName()` and `listenToUdevEvents()`

dgibson:
This is a fix for issue #848.  While we're there, simplify the code a bit and rename a poorly named data structure.

PR #850, opened 2020-09-23
Wait explicitly for VFIO devices to complete hotplug, instead of relying on PCI rescan

dgibson:
Currently the only thing that "ensures" that VFIO devices are actually probed in the VM before the container executes is a forced PCI rescan.  That has a bunch of problems as described in issue #781 .

This PR, in conjuction with a runtime PR to come shortly avoids the rescan by instead having the agent explicitly wait for the expected VFIO devices to become present.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -761,8 +761,7 @@ func (s *sandbox) listenToUdevEvents() {
 		fieldLogger.Infof("Received add uevent")
 
 		// Check if device hotplug event results in a device node being created.
-		if uEv.DevName != "" &&
-			(strings.HasPrefix(uEv.DevPath, rootBusPath) || strings.HasPrefix(uEv.DevPath, acpiDevPath)) {
+		if strings.HasPrefix(uEv.DevPath, rootBusPath) || strings.HasPrefix(uEv.DevPath, acpiDevPath) {
 			// Lock is needed to safely read and modify the sysToDevMap and deviceWatchers.
 			// This makes sure that watchers do not access the map while it is being updated.
 			s.Lock()
--- a/agent.go
+++ b/agent.go
@@ -181,7 +181,7 @@ var logsVSockPort = uint32(0)
 var debugConsoleVSockPort = uint32(0)
 
 // Timeout waiting for a device to be hotplugged
-var hotplugTimeout = 3 * time.Second
+var hotplugTimeout = 10 * time.Second
 
 // Specify the log level
 var logLevel = defaultLogLevel
--- a/device.go
+++ b/device.go
@@ -34,7 +34,12 @@ const (
 	driverNvdimmType    = "nvdimm"
 	driverEphemeralType = "ephemeral"
 	driverLocalType     = "local"
-	vmRootfs            = "/"
+
+	//VFIO devices to be bound to the VM's native drivers (this is
+	//Kata specific behaviour, that requires careful configuration
+	//to get something usable inside the container)
+	driverVfioVmType = "vfio-vm"
+	vmRootfs         = "/"
 )
 
 const (
@@ -97,6 +102,7 @@ var deviceHandlerList = map[string]deviceHandler{
 	driverBlkCCWType:  virtioBlkCCWDeviceHandler,
 	driverSCSIType:    virtioSCSIDeviceHandler,
 	driverNvdimmType:  nvdimmDeviceHandler,
+	driverVfioVmType:  vfioDeviceHandler,
 }
 
 func rescanPciBus() error {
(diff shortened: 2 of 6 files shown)


dgibson:
I'm making the claim that no backport is needed on the basis that while this is a bug, it's apparently more or less worked up to this point.

dgibson:
It looks like my depends-on tag broke the CI for reasona I don't understand yet.  Removing it.

bpradipt:
@dgibson shall we wait for https://github.com/kata-containers/agent/pull/849 to merge and then you can update this PR ?

dgibson:
> @dgibson shall we wait for #849 to merge and then you can update this PR ?

Done :)

agent.go
```diff
@@ -181,7 +181,7 @@ var logsVSockPort = uint32(0)
 var debugConsoleVSockPort = uint32(0)
 
 // Timeout waiting for a device to be hotplugged
-var hotplugTimeout = 3 * time.Second
+var hotplugTimeout = 10 * time.Second
```
devimc, line 184: some container engines like docker and podman kill the container if its status has not changed to `started` after 10s - so the kata container will be killed before timing out - did you consider this?

devimc (approved):
thanks @dgibson - lgtm - I have one question

agent.go
dgibson: Well, in that case the higher level engine could time out before us, but that's kind of unavoidable.  In practice the total delay with SHPC is usualy ~6s, so we'd probably complete in time.  The 10s here is just on the basis that the timeout should generally be substantially longer than the actual expected time, to avoid false positives.

bpradipt (approved):
/lgtm
Thanks @dgibson

jodh-intel (commented):
Thanks @dgibson.

Could you try adding a few unit tests though please?

Although the `virtioBlkCCWDeviceHandler` device handler sneaked in without any tests, I'd rather each handler has *something* in `device_test.go`. Related to this, not that `getDevicePCIAddress` is a variable, so its implementation can be changed to provoke interesting test scenarios. Also, we could consider doing the same for `getDeviceName()` to exercise more of `vfioDeviceHandler()` error paths potentially.

dgibson:
@jodh-intel new push has some unit tests, as requested.

jodh-intel (approved):
Thanks @dgibson. We do also like negative testing (`assert.Error(err)` for as many scenarios as possible), so feel free to consider adding additional tests for that.

lgtm

dgibson:
> Thanks @dgibson. We do also like negative testing (`assert.Error(err)` for as many scenarios as possible), so feel free to consider adding additional tests for that.

I'm a bit disinclined to do heaps here at this time, since it will all have to be ported to Rust pretty soon anyway - and I have notions for longer term cleanups that may obsolete a bunch of it.  I also think quite a lot of the Kata unit tests are so tightly coupled to the code as not to actually be all that useful (they end up testing how it's done, rather than what it's doing).

devimc:
@dgibson  VFIO CI is not happy - can you add: `Depends-on: github.com/kata-containers/runtime#2981` in your commit message?

dgibson:
Temporarily setting DNM, while I try to track down the CI failure.

dgibson:
Urgh.  After all that looks like the failure I was chasing is already present on the main branch.

devimc:
thanks @dgibson but VFIO CI is still failing :(

```
01:05:25 + sudo kata-runtime --kata-config /tmp/tmp.dwr2HumN0S/configuration.toml run --detach -b /tmp/tmp.dwr2HumN0S/bundle --pid-file=/tmp/tmp.dwr2HumN0S/pid vfiotest
01:05:34 rpc error: code = Unknown desc = PCI Identifier for device should be of format [bridgeAddr/deviceAddr], got 
01:05:34 ++ handle_error 86
```

dgibson:
> thanks @dgibson but VFIO CI is still failing :(

Yeah, I basically wasted a day chasing down the wrong failure :( (the one in jenkins-ci-ubunut-18-04-initrd).  Now to figure out how to reproduce and debug the one in jenkins-ubuntu-18-04-vfio.

device.go
```diff
@@ -256,6 +246,46 @@ func nvdimmDeviceHandler(_ context.Context, device pb.Device, spec *pb.Spec, s *
...
+	fieldLogger := agentLog.WithField("host-dev", device.ContainerPath)
+
+	for _, opt := range device.Options {
+		tokens := strings.Split(opt, "=")
+		if len(tokens) != 2 {
+			return fmt.Errorf("Malformed VFIO option \"%s\"", opt)
+		}
+
+		hostBdf := tokens[0]
+		guestPCIPath := tokens[1]
```
jodh-intel, line 263: Given the vfio CI failure @devimc spotted in https://github.com/kata-containers/agent/pull/850#issuecomment-702141192, please could you check both these are `!= ""`.

dgibson: Um.. no.. or at least not right away.  I need to figure out why/how we're getting an empty string here, and probably fix the other side of this.

likebreath (approved):
LGTM. Catching up on the changes from runtime.

dgibson:
Rebased on #855, which we'll need to fix the CI failure on clh.

dgibson:
Since the CI issues are proving so hard to debug, I've decided not to pursue this in Kata1, and leave it to Kata2.

Issue #859, opened 2020-10-26
fail to hotplug device if host memory size large enough


jongwu:
In the current implementation, the maxMem in kata guest is the host memory size. in general, host memory size is not so large so nothing special happen. But if the host memory size large enough, the maxMem will be increase along it. qemu will reserve a large memory zone for the normal memory then the PCIE address area may be affected, especially the base address. As the device path under /sys is subjected by the PCIE base address, that's say, the prefix of the device path, which is named "rootBusPath" in kata-agent, is not static. It is against the current implementation in kata-agent.

How to reproduce this bug:
make sure memory size in your host is large enough, but I'm not sure the lower limit of it, in arm64 it will be around 250G, or you can set a large maxMem in kata-runtime and recompile.
 hot plug device using docker:
`docker run --rm --runtime kata-runtime --device /dev/loop0 ubuntu stat /dev/loop0`

error will occur:
docker: Error response from daemon: OCI runtime create failed: rpc error: code = DeadlineExceeded desc = Timeout reached after 3s waiting for device 0:0:0:0/block: unknown.
ERRO[0004] error waiting for container: context canceled

I have only tested it on arm64. I guess this bug is cross arch, but I'm not sure. so @jodh-intel @alicefr - can you try this on x86 ,ppc and s390?

dgibson:
Fwiw, x86 should not be affected, the sysfs path for the root bus is pretty much static there.  ppc64 shouldn't be affected by this specific bug - the root bus path shouldn't be affected by max memory.  However, the root bus sysfs path *can* vary, at least in theory, for a bunch of other reasons on ppc, so I suspect your change will be vaulable there.  I don't know about s390.

jongwu:
thanks @dgibson !

alicefr:
@jongwu s390x is a very special case. It is not using PCI devices, but CCW. So, my guess is that this doesn't affect s390x at all.
However, I changed recently job so I don't have access to an s390x system to verify it. Ccing @jschintag

I think, it will be useful if you could paste the agent log.

Out of curiosity, do you have the same behavior with the rust agent?

jongwu:
@alicefr - , on arm64, this kata bug is related with qemu.  the PCIE memory map overlaps with the normal memory ( I don't think it is a bug of qemu) then the base address of PCIE may vary with the normal memory size.  so I think the rust agent also has this bug on arm64, even I have not tested it.

alicefr:
[...]

PR #860, opened 2020-10-26
rootBusPath: create rootBusPath dynamically.

jongwu:
Currently, rootBusPath is set as a constant value. Bus this pcie bus path
is not alway static, eg. the rootBus on arm64 is
"/devices/platform/4010000000.pcie/pci0000:00". The address part of
"4010000000.pcie" may vary with the maxMem in qemu so it should not be a
fixed value.

For exmaple:
HIGH PCIE address reserved in qemu on virt is from 0 to 512G. The lower
part of address may be allocated to normal memory. In general, normal memory
is largely less than 256G, so the base address of HIGH PCIE can be 256G.
But, in case the maxmem is large enough, like around 256G, the base address
of HIGH PCIE must be increase, eg. 300G.
In the current implementation of kata-runtime, the maxmem in qemu is the host
memory size. So if the host memory size is large enough, the prefix of the
pci device path name will be different from rootBusPath set in kata-agent which
can lead to failure on device hotplug.

This patch offer a mechanism to create rootBusPath dynamically but only give
implemention for arm64 and return the default value of rootBusPath for other arch.

Fixes: #859

@jodh-intel @devimc

### DIFF
--- a/device_arm64.go
+++ b/device_arm64.go
@@ -6,9 +6,14 @@
 
 package main
 
-const (
-	rootBusPath = "/devices/platform/4010000000.pcie/pci0000:00"
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+)
 
+const (
 	// From https://www.kernel.org/doc/Documentation/acpi/namespace.txt
 	// The Linux kernel's core ACPI subsystem creates struct acpi_device
 	// objects for ACPI namespace objects representing devices, power resources
@@ -16,3 +21,22 @@ const (
 	// sysfs as directories in the subtree under /sys/devices/LNXSYSTM:00
 	acpiDevPath = "/devices/LNXSYSTM"
 )
+
+func createRootBusPath() (string, error) {
+	startRootBusPath := "/devices/platform"
+	endRootBusPath := "/pci0000:00"
+	sysStartRootBusPath := filepath.Join(sysfsDir, startRootBusPath)
+	files, err := ioutil.ReadDir(sysStartRootBusPath)
+	if err != nil {
+		return "", fmt.Errorf("Error reading %s: %s", sysStartRootBusPath, err)
+	}
+
+	// find out the directory end with ".pcie"
+	for _, file := range files {
+		if strings.HasSuffix(file.Name(), ".pcie") && file.IsDir() {
+			return filepath.Join(startRootBusPath, file.Name(), endRootBusPath), nil
+		}
+	}
+
+	return "", fmt.Errorf("no pcie bus found under %s", sysStartRootBusPath)
+}
--- a/agent.go
+++ b/agent.go
@@ -709,6 +709,11 @@ func (s *sandbox) waitForStopServer() {
 
 func (s *sandbox) listenToUdevEvents() {
 	fieldLogger := agentLog.WithField("subsystem", "udevlistener")
+	rootBusPath, err := createRootBusPath()
+	if err != nil {
+		fieldLogger.Warnf("Error creating root bus path")
+		return
+	}
 
 	uEvHandler, err := uevent.NewHandler()
 	if err != nil {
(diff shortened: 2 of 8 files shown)


alicefr:
cc @jschintag

dgibson (approved):
I think this is a good idea.  ppc64 doesn't need this just yet, but at some point it will need the ability to handle multiple PCI roots, and this gets us a little closer to doing that.

devimc (approved):
thanks @jongwu

device_arm64.go
```diff
@@ -6,13 +6,37 @@
...
+func createRootBusPath() string {
+	startRootBusPath := "/devices/platform"
+	endRootBusPath := "/pci0000:00"
+	sysStartRootBusPath := filepath.Join(sysfsDir, startRootBusPath)
+	files, err := ioutil.ReadDir(sysStartRootBusPath)
+	if err != nil {
+		fmt.Errorf("Error reading %s: %s", sysStartRootBusPath, err)
+	}
+
+	// find out the director end with ".pcie"
```
jodh-intel, line 34: Is this a typo ("directory" rather than "director")?

device_arm64.go
```diff
@@ -6,13 +6,37 @@
...
+	for _, file := range files {
+		if strings.HasSuffix(file.Name(), ".pcie") {
```
jodh-intel, line 36: If the comment above has a typo, shouldn't this also test that `file.IsDir() == true`?

device_arm64.go
```diff
@@ -6,13 +6,37 @@
...
+			return filepath.Join(startRootBusPath, file.Name(), endRootBusPath)
+		}
+	}
+
+	return ""
```
jodh-intel, line 41: What does returning `""` mean? Is that a valid scenario or should this function actually return `(string, error)`?

device_arm64.go
```diff
@@ -6,13 +6,37 @@
...
 	acpiDevPath = "/devices/LNXSYSTM"
 )
+
+func createRootBusPath() string {
+	startRootBusPath := "/devices/platform"
+	endRootBusPath := "/pci0000:00"
+	sysStartRootBusPath := filepath.Join(sysfsDir, startRootBusPath)
+	files, err := ioutil.ReadDir(sysStartRootBusPath)
+	if err != nil {
+		fmt.Errorf("Error reading %s: %s", sysStartRootBusPath, err)
```
jodh-intel, line 31: This suggests the function should be returning an `error` to me...?

jodh-intel (changes requested):
Hi @jongwu - thanks for raising. I've added some comments so blocking PR for now...

jodh-intel (approved):
Thanks @jongwu.

lgtm

jodh-intel:
@jongwu - Travis is failing - looks like you need to update the tests for the latest changes.

jongwu:
thanks @jodh-intel , I have fixed it. now could you please retrigger ci again?

jongwu:
hello @jodh-intel -, I can't see ARM CI, could please trigger it?

jodh-intel:
/test-arm

@jongwu - I've added you to the runtime + agent repos so you should be able to trigger builds yourself from now on. Also, are you planning to port this to the new 2.0 rust agent?

amshinde (approved):

amshinde:
Dont see an ARM CI.
@chavafg Is it not enabled for the agent?

chavafg:
hmm, yeah, it seems that arm CI has not been enabled for this repo. @jongwu could you help adding it? let me know if you need help.

jongwu:
Thanks @jodh-intel -, I have composed the rust version of this patch, see [rust](https://github.com/kata-containers/kata-containers/pull/1060). It's not easy for me to write rust code.

jodh-intel:
Thanks @jongwu! No problem - we can help if necessary :smile:

jongwu:
thanks @jodh-intel .

jongwu:
Hello @jodh-intel @chavafg, all tests pass. isn't time to merge this?

Issue #861, opened 2020-10-29
make proto fail using docker 18.06 on arm64


jongwu:
make proto fail on arm64 see http://jenkins.katacontainers.io/job/kata-containers-agent-ARM-PR/16/consoleFull

Step 5/20 : RUN yum makecache && yum install -y 	git automake libtool glibc-headers gcc-c++ make
11:51:30  ---> Running in 4868b4b0255c
11:55:32 CentOS-8 - AppStream                            0.0  B/s |   0  B     04:00    
11:55:32 Errors during downloading metadata for repository 'AppStream':
11:55:32   - Curl error (6): Couldn't resolve host name for http://mirrorlist.centos.org/?release=8&arch=aarch64&repo=AppStream&infra=container [Could not resolve host: mirrorlist.centos.org]
11:55:32 Error: Failed to download metadata for repo 'AppStream': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for http://mirrorlist.centos.org/?release=8&arch=aarch64&repo=AppStream&infra=container [Could not resolve host: mirrorlist.centos.org]
11:55:32 The command '/bin/sh -c yum makecache && yum install -y 	git automake libtool glibc-headers gcc-c++ make' returned a non-zero code: 1
11:55:33 Makefile:120: recipe for target 'build-image' failed
11:55:33 make: *** [build-image] Error 1
11:55:33 Build step 'Execute shell' marked build as failure

it seems docker issue, as this issue will be disappear if using docker 19.03

PR #862, opened 2020-10-29
ci: install docker 19.03 for arm64 to let build image go

jongwu:
"make proto" will fail on arm64 using docker 18.06. This bug will be
gone if using docker 19.03. so upgrade docker before "make proto"
for arm64.

Fixes: #861

@jodh-intel @devimc

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -13,6 +13,12 @@ pushd "${tests_repo_dir}"
 .ci/run.sh
 testcidir=$(dirname "$0")
 
+# docker version may need to be upgraded to let make proto go on arm64
+arch=$(go env GOARCH)
+if [ "$arch" == "arm64" ]; then
+	"../agent/.ci/run_arm64.sh"
+fi
+
 echo "Starting docker service before making proto"
 sudo systemctl start docker
 
--- a/.ci/run_arm64.sh
+++ b/.ci/run_arm64.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Arm Ltd Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source "${tests_repo_dir}/.ci/lib.sh"
+
+arch=$(go env GOARCH)
+install_docker_ubuntu() {
+        pkg_name="docker-ce"
+        repo_url="https://download.docker.com/linux/ubuntu"
+        curl -L "${repo_url}/gpg" | sudo apt-key add -
+        sudo -E add-apt-repository "deb [arch=${arch}] ${repo_url} $(lsb_release -cs) stable"
+        sudo -E apt-get update
+        docker_version_full=$(apt-cache madison $pkg_name | grep "$docker_version" | awk '{print $3}' | head -1)
+        sudo -E apt-get -y install "${pkg_name}=${docker_version_full}"
+}
+
+# make proto will fail on arm64 when use docker 18.06 and using docker 19.03
+# can avoid this failure. For now this change is only for ubuntu as we know little
+# about the other cases.
+main() {
+	ID=$(cat /etc/os-release | grep "^ID=" | cut -f 2 -d "=")
+        if [ "$ID" != "ubuntu" ]; then
+		echo "docker upgrade is only done for ubuntu"
+                exit 0
+        fi
+        current_docker_version=$(docker version | awk '/Engine/{getline; print $2 }')
+        current_docker_version=${current_docker_version%.*}
+        docker_version=$(get_version "externals.docker.architecture.aarch64.agent.version")
+        docker_version=${docker_version/v}
+        docker_version=${docker_version/-*}
+        if [[ `echo "$current_docker_version < $docker_version" | bc` -eq 1 ]]; then
+                command -v docker >/dev/null 2>&1 && "./cmd/container-manager/manage_ctr_mgr.sh" docker remove
+                echo "reinstall docker $docker_version for arm64"
+                install_docker_ubuntu $docker_version
+        fi
+}
+
+main


.ci/run.sh
```diff
@@ -9,10 +9,27 @@ set -e
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
+install_docker() {
+	docker_version=${docker_version:-19.03}
```
devimc, line 13: is this version of docker required only in this repo? why don't change in the tests repo?

.ci/run.sh
```diff
@@ -9,10 +9,27 @@ set -e
...
+}
+
 pushd "${tests_repo_dir}"
 .ci/run.sh
 testcidir=$(dirname "$0")
 
+if [ $arch == "arm64" ]; then
+	command -v docker >/dev/null 2>&1 && "${testcidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker remove
+	echo "reinstall docker 19.03 for arm64"
+	install_docker 19.03
```
devimc, line 30: this `install_docker` is distro specific, I think you should rename it to `install_docker_ubuntu` or something else that contains the `ubuntu` word

.ci/run.sh
```diff
@@ -9,10 +9,27 @@ set -e
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
+install_docker() {
+	docker_version=${docker_version:-19.03}
+	pkg_name="docker-ce"
+	repo_url="https://download.docker.com/linux/ubuntu"
```
devimc, line 15: before installing docker, could you please check that you are running on ubuntu?

devimc (commented):
thanks @jongwu - I have some questions

c3d (approved):
Looks good to me, but I only checked there was no impact on x86.

.ci/run.sh
jongwu: yeah, I find this is necessary only in agent, also change in tests repo won't work the version is available for 18.06 for now.

.ci/run.sh
jongwu: ok

.ci/run.sh
jongwu: ok. thanks

.ci/run.sh
jodh-intel: This might work today, but what about when Docker 19.03 is no longer available? It would be much safer to check the installed version of docker and if it is "< 19.03", upgrade it.

But we store these sorts of version details in the versions database:

- https://github.com/kata-containers/runtime/blob/master/versions.yaml#L211

And if it is architecture-specific, add an `aarch64` block like here:

- https://github.com/kata-containers/runtime/blob/master/versions.yaml#L101..L106

.ci/run.sh
```diff
@@ -9,10 +9,28 @@ set -e
...
+}
+
 pushd "${tests_repo_dir}"
 .ci/run.sh
 testcidir=$(dirname "$0")
 
+if [ "$arch" == "arm64" ]; then
+	command -v docker >/dev/null 2>&1 && "${testcidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker remove
+	echo "reinstall docker 19.03 for arm64"
+	install_docker_ubuntu 19.03
```
jodh-intel, line 31: Please don't hard code version numbers in strings - use variables to make the scripts easier to maintain.

jodh-intel (changes requested):

.ci/run.sh
jongwu: @jodh-intel - any comments?

.ci/run.sh
```diff
@@ -9,10 +9,39 @@ set -e
 cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
+install_docker_ubuntu() {
+	pkg_name="docker-ce"
+	repo_url="https://download.docker.com/linux/ubuntu"
+	curl -fsSL "${repo_url}/gpg" | sudo apt-key add -
```
jodh-intel, line 15: These options look inconsistent to me:

```
 -f, --fail          Fail silently (no output at all) on HTTP errors
 -S, --show-error    Show error even when -s is used
 -s, --silent        Silent mode
```

Why are you suppressing errors anyway as that will make debugging a remote CI very hard?

jodh-intel (changes requested):
Thanks @jongwu. Buit imho these changes look confusing: they are architecture, distro *and* container manager specific. Further, there are no comments in the code explaining what is happening. Plus, they make the script overly complex for the non-ARM case.

It would be much cleaner I think if you created something like `.ci/run-aarch64.sh` and called it a function in it conditionally.

.ci/run.sh
jongwu: thanks @jodh-intel -, I fold the change into a new file, also tweak this curl command.

.ci/run.sh
```diff
@@ -13,6 +13,12 @@ pushd "${tests_repo_dir}"
 .ci/run.sh
 testcidir=$(dirname "$0")
 
+# docker version may need to be upgraded to let make proto go on arm64
+arch=$(go env GOARCH)
+if [ "$arch" == "arm64" ]; then
+	"../agent/.ci/run_arm64.sh"
```
jodh-intel, line 19: This path looks odd - shouldn't it be something like the following?:

```bash
cidir=$(dirname "$0")
"${cidir}/run_arm64.sh"
```

jodh-intel (approved):
Thanks @jongwu. That's much clearer imho.

lgtm

.ci/run.sh
jongwu: but it won't work as the current dir is .../tests. run_arm64.sh is not found in the path of ".ci/run_arm64.sh"

jodh-intel: Ack! I was only looking at that line in isolation ;)

jongwu:
hello @jodh-intel @devimc -, should this be merged?

devimc (approved):
thanks @jongwu

PR #872, opened 2020-12-22
mount: Fix hugepages support

### LINKED ISSUE #871: hugepage support in Kata
This is related to the runtime issue https://github.com/kata-containers/runtime/issues/2353

bpradipt:
Fixes: #871

Signed-off-by: Pradipta Banerjee <pradipta.banerjee@gmail.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -10,6 +10,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -26,12 +27,14 @@ import (
 )
 
 const (
-	type9pFs       = "9p"
-	typeVirtioFS   = "virtiofs"
-	typeRootfs     = "rootfs"
-	typeTmpFs      = "tmpfs"
-	procMountStats = "/proc/self/mountstats"
-	mountPerm      = os.FileMode(0755)
+	type9pFs             = "9p"
+	typeVirtioFS         = "virtiofs"
+	typeRootfs           = "rootfs"
+	typeTmpFs            = "tmpfs"
+	typeHugeTlbfs        = "hugetlbfs"
+	procMountStats       = "/proc/self/mountstats"
+	mountPerm            = os.FileMode(0755)
+	sysfsHugepagesPrefix = "/sys/kernel/mm/hugepages"
 )
 
 var flagList = map[string]int{
@@ -111,6 +114,15 @@ func mount(source, destination, fsType string, flags int, options string) error 
 		absSource = source
 	case typeTmpFs:
 		absSource = source
+	case typeHugeTlbfs:
+		absSource = source
+		//Allocate hugepages before mount
+		///sys/kernel/mm/hugepages/hugepages-1048576kB/nr_hugepages
+		///sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
+		//options eg "pagesize=2048,size=107374182"
+		if err = allocateHugePages(options); err != nil {
+			return grpcStatus.Errorf(codes.Internal, "Could not mount hugepages: %v", err)
+		}
 	default:
 		absSource, err = filepath.EvalSymlinks(source)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


mount.go
```diff
@@ -132,6 +143,29 @@ func mount(source, destination, fsType string, flags int, options string) error
...
+	//sysfs entry is always of the form hugepages-${size}kB
+	//Ref: https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt
+	pageSizeKbStr := strconv.FormatInt((pageSize/1024), 10) + "kB"
+	sysfsHugepagesPath := sysfsHugepagesPrefix + "hugepages-" + pageSizeKbStr + "/" + "nr_hugepages"
+
+	sizeStr := strings.TrimPrefix(opt[1], "size=")
+	size, _ := strconv.ParseInt(sizeStr, 10, 64)
+	numpages := strconv.FormatInt((size / pageSize), 10)
+	if err := ioutil.WriteFile(sysfsHugepagesPath, []byte(numpages), 0644); err != nil {
+		agentLog.WithField("sysfsHugepagesPath", sysfsHugepagesPath).WithError(err).Errorf("Could not allocate hugepages")
```
egernst, line 163: We should be returning err here, right?

mount.go
```diff
@@ -111,6 +115,13 @@ func mount(source, destination, fsType string, flags int, options string) error
 		absSource = source
 	case typeTmpFs:
 		absSource = source
+	case typeHugeTlbfs:
+		absSource = source
+		//Allocate hugepages before mount
+		///sys/kernel/mm/hugepages/hugepages-1048576kB/nr_hugepages
+		///sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages
+		//options eg "pagesize=2MB,size=107374182"
+		allocateHugePages(options)
```
egernst, line 124: I imagine this could fail (and doesn't it return an error?) We should be looking to return an error here.

egernst (commented):
This part seems more straight forward... though I see one issue (need to return error).  The hard part is making sure "these" huge pages are backed by "actual" huge pages from the host.

mount.go
```diff
@@ -132,6 +144,35 @@ func mount(source, destination, fsType string, flags int, options string) error
...
+	//options eg "pagesize=2048,size=107374182"
+	opt := strings.Split(options, ",")
+	pageSizeStr := strings.TrimPrefix(opt[0], "pagesize=")
+	pageSize, err := strconv.ParseInt(pageSizeStr, 10, 64)
+	if err != nil {
+		return err
+	}
+	//sysfs entry is always of the form hugepages-${size}kB
+	//Ref: https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt
+	sysfsHugepagesPath := fmt.Sprintf("%s/hugepages-%skB/nr_hugepages", sysfsHugepagesPrefix, strconv.FormatInt((pageSize/1024), 10))
```
jodh-intel, line 161: Can you use `path.Join()` / `filepath.Join()` here?

mount.go
```diff
@@ -132,6 +144,35 @@ func mount(source, destination, fsType string, flags int, options string) error
...
 }
 
+// Allocate hugepages by writing to sysfs
+func allocateHugePages(options string) error {
+
+	agentLog.WithField("HugePagesOption", options).Info("HugePages option string")
+
+	//options eg "pagesize=2048,size=107374182"
+	opt := strings.Split(options, ",")
+	pageSizeStr := strings.TrimPrefix(opt[0], "pagesize=")
```
jodh-intel, line 154: I can see this aligns with https://github.com/kata-containers/runtime/pull/3109 in terms of the order of the options. But this would break if the runtime switched the option order.

To protect against this, I'd rather we loop over `opt` and look for the `pagesize=` and `size=` prefixes instead.

mount.go
```diff
@@ -132,6 +144,35 @@ func mount(source, destination, fsType string, flags int, options string) error
...
+	//Ref: https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt
+	sysfsHugepagesPath := fmt.Sprintf("%s/hugepages-%skB/nr_hugepages", sysfsHugepagesPrefix, strconv.FormatInt((pageSize/1024), 10))
+
+	sizeStr := strings.TrimPrefix(opt[1], "size=")
+	size, err := strconv.ParseInt(sizeStr, 10, 64)
+	if err != nil {
+		return err
+	}
+	numpages := strconv.FormatInt((size / pageSize), 10)
+	if err := ioutil.WriteFile(sysfsHugepagesPath, []byte(numpages), 0644); err != nil {
```
jodh-intel, line 169: I guess the perms shouldn't need to be used, but "just in case" couldn't this be `0600` (or even `0200`)?

mount.go
bpradipt: yeah sure

mount.go
bpradipt: I used the default permissions that are present for the entries. However `0200` can also be used. The permissions here are irrelevant since the entries already exists.

mount.go
bpradipt: Sure I'll change it. I was hoping to avoid looping since we control the options string passed by the runtime.

bpradipt:
The vfio test failure is unrelated and looks like this is the cause  `/usr/bin/cloud-hypervisor does not exist` .

@dgibson @jodh-intel can you please take a look at this PR.  I have incorporated all the changes. Post your review I can start working on forward porting the hugepages changes to 2.x. Thanks

egernst:
I have some questions still, from https://github.com/kata-containers/runtime/issues/2353#issuecomment-757059874

I’m okay with adding support, but I think it should be behind a feature gate, since I *think* this may cause issues in Kube (our memory overhead will consume a lot of the pages).

dgibson (approved):
LGTM

jodh-intel (approved):
Thanks @bpradipt.

lgtm

bpradipt:
> I have some questions still, from [kata-containers/runtime#2353 (comment)](https://github.com/kata-containers/runtime/issues/2353#issuecomment-757059874)
> 
> I’m okay with adding support, but I think it should be behind a feature gate, since I _think_ this may cause issues in Kube (our memory overhead will consume a lot of the pages).

Sorry @egernst, I completely missed it.  I have posted few questions and let's discuss to arrive at a consensus.

devimc:
hold on please, the CI didn't run

devimc:
I'm investigating the issue

```
INFO: No forcing label found
Short circuit fast path skipping the rest of the CI.
```

devimc:
@bpradipt - I see you want to merge this in master

```
bpradipt wants to merge 1 commit into kata-containers:master
```

is that correct?

bpradipt:
> @bpradipt - I see you want to merge this in master
> 
> ```
> bpradipt wants to merge 1 commit into kata-containers:master
> ```
> 
> is that correct?

This is for kata-containers/agent master (1.x)

devimc:
CI is buggy, no changes were detected :disappointed: 

```
+ info 'Checking for any changed files that will prevent CI fastpath return'
+ echo -e 'INFO: Checking for any changed files that will prevent CI fastpath return'
INFO: Checking for any changed files that will prevent CI fastpath return
++ can_we_skip
++ '[' '' = true ']'
+++ get_pr_changed_file_details_full
+++ local filters=
+++ filters+=A
+++ filters+=C
+++ filters+=M
+++ filters+=R
+++ filters+=UX
+++ git diff-tree -r --name-status --diff-filter=ACMRUX origin/master HEAD
++ filenames=
+++ echo ''
+++ awk '{print $NF}'
++ filenames=
++ '[' -z '' ']'
++ local_info 'No files found'
++ msg='No files found'
++ info 'No files found'
++ echo -e 'INFO: No files found'
INFO: No files found
```

fidencio:
On a side note, I decided to take a look at the **jenkins-ci-ubuntu-18-04** and it's building the rust agent. Why? Oo

PR #880, opened 2021-01-14
agent: Fix resourece leak

keloyang:
Every ExecProcess will create a struct for process and container, and kata-agent need call WaitProcess to remove them.
Normally WaitProcess will be called by kata-shim(**exitcode, err := shim.wait()**)，but kata-shim can't do that in some cases, 
it will make resource leak, e.g. mem and /dev/pts/N. So I think the cleanup for ExecProcess can't depend on kata-shim calling WaitProcess,it should depends on whether the process exits.

take the following case, we will see the /det/pts/N leak,
1. create a pod with kata-container, omited.
2. terminal 1, check pts, and exec to kata-container
```
[root@centos0 deployment]# kubectl get pod
NAME                     READY     STATUS    RESTARTS   AGE
pts-6794f84bd9-jzp5b   1/1       Running   0          1h
# check pts 
[root@centos0 deployment]#  kubectl exec -ti pts-6794f84bd9-jzp5b ls /dev/pts
0  ptmx
# exec to kata container and run sleep 30, before sleep 30 exited, kill kata-shim(with  -terminal) at terminal 2
[root@centos0 deployment]#  kubectl exec -ti pts-6794f84bd9-jzp5b sleep 30
# check pts again
[root@centos0 deployment]#  kubectl exec -ti pts-6794f84bd9-jzp5b ls /dev/pts
0  1  ptmx
...
# repeat kubectl exec -ti pts-6794f84bd9-jzp5b sleep 30 and kill kata-shim
[root@centos0 deployment]#  kubectl exec -ti pts-6794f84bd9-jzp5b ls /dev/pts
0  1  2  3  4  5  6  7  9  8  ptmx
```
2. terminal 2, kill kata-shim which is created by **kubectl exec -ti pts-6794f84bd9-jzp5b sleep 30**
```
[root@centos1 ~]# ps -eaf|grep kata-shim
root      572188  572134  0 18:56 ?        00:00:00 /kata/bin/kata-shim -agent unix:///run/vc/sbs/56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595/proxy.sock -container 56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595 -exec-id 56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595
root      572333  572314  0 18:56 ?        00:00:00 /kata/bin/kata-shim -agent unix:///run/vc/sbs/56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595/proxy.sock -container 6e8a25168697148e7065a9253c2caa5626548f6ed02c877ddc9dc93caaa6b76f -exec-id 6e8a25168697148e7065a9253c2caa5626548f6ed02c877ddc9dc93caaa6b76f
root      823320  572314  0 20:32 pts/3    00:00:00 /kata/bin/kata-shim -agent unix:///run/vc/sbs/56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595/proxy.sock -container 6e8a25168697148e7065a9253c2caa5626548f6ed02c877ddc9dc93caaa6b76f -exec-id 014f96c3-67a6-422c-9709-33f717cc624f -terminal
[root@centos1 ~]# kill -9 823320
```

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -19,6 +19,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strings"
 	"sync"
 	"syscall"
@@ -47,6 +48,8 @@ const (
 	bashPath         = "/bin/bash"
 	shPath           = "/bin/sh"
 	debugConsolePath = "/dev/console"
+
+	maxExitCodeNum = 2048
 )
 
 var (
@@ -113,6 +116,7 @@ type container struct {
 	useSandboxPidNs bool
 	agentPidNs      bool
 	ctx             context.Context
+	processExitCode exitProcess
 }
 
 type sandboxStorage struct {
@@ -146,6 +150,19 @@ type sandbox struct {
 	storages          map[string]*sandboxStorage
 	stopServer        chan struct{}
 	oomEvents         chan string
+}
+
+type exitProcessStatus struct {
+	exitCode    chan int
+	id          string
+	sequenceNum int
+}
+
+type exitProcess struct {
+	sync.Mutex
+
+	exitCodes map[string]*exitProcessStatus
+	sequence  int
 }
 
 var agentFields = logrus.Fields{
@@ -873,6 +890,71 @@ func (s *sandbox) setupSignalHandler() error {
 	errCh := make(chan error, 1)
 	go s.signalHandlerLoop(sigCh, errCh)
 	return <-errCh
+}
+
+type exitStatusList []*exitProcessStatus
+
+func (p exitStatusList) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+func (p exitStatusList) Len() int           { return len(p) }
+func (p exitStatusList) Less(i, j int) bool { return p[i].sequenceNum-p[j].sequenceNum < 0 }
+
+func (e *exitProcess) add(id string, exitCode chan int) {
+	e.Lock()
+	defer e.Unlock()
+
+	if e.exitCodes == nil {
+		e.exitCodes = map[string]*exitProcessStatus{}
+	}
+
+	// delete (first in and first out) half of the elements if the length arrive maxExitCodeNum
+	if len(e.exitCodes) >= maxExitCodeNum {
+		index := 0
+		statusList := make(exitStatusList, maxExitCodeNum)
+		for _, status := range e.exitCodes {
+			statusList[index] = status
+			index++
+		}
+		sort.Sort(statusList)
+
+		for index, status := range statusList {
+			if index >= maxExitCodeNum/2 {
+				break
+			}
+			delete(e.exitCodes, status.id)
+		}
+	}
+
+	e.exitCodes[id] = &exitProcessStatus{
+		exitCode:    exitCode,
+		sequenceNum: e.sequence,
+		id:          id,
+	}
+	e.sequence++
+}
+
+func (e *exitProcess) wait(id string) (int, error) {
+	e.Lock()
+	status, find := e.exitCodes[id]
+	e.Unlock()
+
+	if find {
+		exitCode := <-status.exitCode
+
+		e.delete(id)
+
+		//refill with the exitcode which can be read out
+		//by another exitProcess.wait
+		status.exitCode <- exitCode
+		return exitCode, nil
+	}
+
+	return -1, fmt.Errorf("can't find process %s", id)
+}
+
+func (e *exitProcess) delete(id string) {
+	e.Lock()
+	defer e.Unlock()
+	delete(e.exitCodes, id)
 }
 
 // getMemory returns a string containing the total amount of memory reported
--- a/grpc.go
+++ b/grpc.go
@@ -474,8 +474,26 @@ func (a *agentGRPC) postExecProcess(ctr *container, proc *process) error {
 	}
 
 	ctr.setProcess(proc)
-
+	a.addExitCodeProcess(proc, ctr)
 	return nil
+}
+
+func (a *agentGRPC) addExitCodeProcess(proc *process, ctr *container) {
+	ctr.processExitCode.add(proc.id, proc.exitCodeCh)
+
+	go func() {
+		libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
+		exitCode, _ := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
+
+		//refill the exitCodeCh with the exitcode which can be read out
+		//by another WaitProcess(). Since this channel isn't be closed,
+		//here the refill will always success and it will be free by GC
+		//once the process exits.
+		proc.exitCodeCh <- exitCode
+
+		proc.closePostExitFDs()
+		ctr.deleteProcess(proc.id)
+	}()
 }
 
 // This function updates the container namespaces configuration based on the
@@ -1001,27 +1019,15 @@ func isSignalHandled(pid int, signum syscall.Signal) bool {
 }
 
 func (a *agentGRPC) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest) (*pb.WaitProcessResponse, error) {
-	proc, ctr, err := a.sandbox.getProcess(req.ContainerId, req.ExecId)
+	ctr, err := a.getContainer(req.ContainerId)
 	if err != nil {
 		return &pb.WaitProcessResponse{}, err
 	}
 
-	defer proc.Do(func() {
-		proc.closePostExitFDs()
-		ctr.deleteProcess(proc.id)
-	})
-
-	// Using helper function wait() to deal with the subreaper.
-	libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
-	exitCode, err := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
+	exitCode, err := ctr.processExitCode.wait(req.ExecId)
 	if err != nil {
 		return &pb.WaitProcessResponse{}, err
 	}
-	//refill the exitCodeCh with the exitcode which can be read out
-	//by another WaitProcess(). Since this channel isn't be closed,
-	//here the refill will always success and it will be free by GC
-	//once the process exits.
-	proc.exitCodeCh <- exitCode
 
 	return &pb.WaitProcessResponse{
 		Status: int32(exitCode),
--- a/agent_test.go
+++ b/agent_test.go
@@ -14,6 +14,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strings"
 	"syscall"
 	"testing"
@@ -988,4 +989,24 @@ func TestRunOOMEventMonitor(t *testing.T) {
 	assert.Equal(cid, oomEvent)
 
 	close(eventChan)
+}
+
+func TestExitProcess(t *testing.T) {
+	assert := assert.New(t)
+	containerExitProcess := exitProcess{}
+	for i := 0; i <= maxExitCodeNum; i++ {
+		id := fmt.Sprintf("id-%d", i)
+		containerExitProcess.add(id, nil)
+	}
+
+	listLen := len(containerExitProcess.exitCodes)
+	statusList := make(exitStatusList, listLen)
+	index := 0
+	for _, status := range containerExitProcess.exitCodes {
+		statusList[index] = status
+		index++
+	}
+	sort.Sort(statusList)
+	assert.Equal(statusList[0].sequenceNum, maxExitCodeNum/2)
+	assert.Equal(statusList[listLen-1].sequenceNum, maxExitCodeNum)
 }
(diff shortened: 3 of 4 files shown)


keloyang:
ping @jon @liwei @jbryce @gnawux PTAK, thanks.

grpc.go
```diff
@@ -1000,29 +1001,31 @@ func isSignalHandled(pid int, signum syscall.Signal) bool {
...
+	exitCode, _ := a.sandbox.subreaper.wait(proc.exitCodeCh, libContProcess)
+
 	//refill the exitCodeCh with the exitcode which can be read out
 	//by another WaitProcess(). Since this channel isn't be closed,
 	//here the refill will always success and it will be free by GC
 	//once the process exits.
 	proc.exitCodeCh <- exitCode
 
+	proc.closePostExitFDs()
+	ctr.deleteProcess(proc.id)
```
lifupan, line 1015: You cannot delete the process here, since this fn would be called before the shim calling WaitProcess, and thus the shim wouldn't get the process's exit code.

keloyang: @lifupan thanks for you review. I think you are right ,but I  think shim can improve the call for WaitProcess, shim should not call WaitProcess after shim.proxyStdio, it's better to have them both in parallel(use go shim.proxyStdio https://github.com/kata-containers/shim/blob/master/main.go#L354, or other code in kata-containerd-shim-v2), this will make shim can get exitcode.  If  not parallel , Can we add some delay for closePostExitFDs?

lifupan: @keloyang 

Actually we shouldn't call the WaitProcess parallelly with shim.proxyStdio, since this would cause the lost of io data, what's why we called the WaitProcess after shim.proxyStdio done. 

For the issue of resource leak, I think it's not easy to avoid, since we depend on the shim to get the process's exit code, and if the shim didn't call the WaitProcess, then the essential resource should be kept for WaitProcess calling.

keloyang: @lifupan thanks for you review
>I think it's not easy to avoid

I think we can still do some thing to reduce the impact of the problem.  for example, when the process exit, kata-agent delay some time and then call deleteProcess,  so shim.Wait can get exitcode in the delay time. 

I have a usage scenario，if the exec process has no io data  in an hour, kata-proxy will close the connection from kata-shim.  this make shim.wait failed and lead kata-agent resource leak, at the end, kubectl can't exec to  the kata container.

I think the stability of the program should depend on itself as much as possible, normal operation should not cause program problems, just like our operating system, we constantly create a process and then kill it, this should not cause problems in our operating system.

lifupan: > @lifupan thanks for you review
> 
> > I think it's not easy to avoid
> 
> I think we can still do some thing to reduce the impact of the problem. for example, when the process exit, kata-agent delay some time and then call deleteProcess, so shim.Wait can get exitcode in the delay time.
> 
> I have a usage scenario，if the exec process has no io data in an hour, kata-proxy will close the connection from kata-shim. this make shim.wait failed and lead kata-agent resource leak, at the end, kubectl can't exec to the kata container.

Why the kata-proxy/kata-shim closed after no io data in an hour?  Did you killed it by your system? 

> 
> I think the stability of the program should depend on itself as much as possible, normal operation should not cause program problems, just like our operating system, we constantly create a process and then kill it, this should not cause problems in our operating system.

Actually, if you didn't reap/wait the child process, there would be zombie process left too.   It's the same with the agent issue. If you just create the process, and didn't do the reap, there's the case to left the resources in the system too.

keloyang: >Actually, if you didn't reap/wait the child process, there would be zombie process left too. It's the same with the agent issue. If you just create the process, and didn't do the reap, there's the case to left the resources in the system too.

@lifupan thanks， but I don't agree with you，create a process and then kill it , it is  a normal operation, even if don't reap it, the system init will reap the exited process. It is diffrent from kata-agent，if WaitProcess is not called，no other cat help kata-agent to do the closePostExitFDs. The leak resources  are not zombie  process, but fd.

I have add delay to call closePostExitFDs, PTAL, thanks.

lifupan: > > Actually, if you didn't reap/wait the child process, there would be zombie process left too. It's the same with the agent issue. If you just create the process, and didn't do the reap, there's the case to left the resources in the system too.
> 
> @lifupan thanks， but I don't agree with you，create a process and then kill it , it is a normal operation, even if don't reap it, the system init will reap the exited process. It is diffrent from kata-agent，if WaitProcess is not called，no other cat help kata-agent to do the closePostExitFDs. The leak resources are not zombie process, but fd.
> 

The init process only do the reap when the child process had no parent, if the parent process didn't exit and didn't do the wait/reap to the exited child process, then there would be zombie process left in the system too and the kernel/init wouldn't cleanup all of the process's resource for you. 

> I have added delay to call closePostExitFDs, PTAL, thanks.

The delay call cannot make sure the later wait call could get the process's exit code. I think you can add some resource cleanup here, but the some essential resources should be kept until the wait called to do the last resource clean.

keloyang: @lifupan thanks very much for your reply.

>The init process only do the reap when the child process had no parent

I understand the issue and when the exited process 's father become init, the init will reap the exited process, but I think it is not  the focus of our discussion. -_-
>add some resource cleanup here, but the some essential resources should be kept 

It is improvement， and I have changed th pr, please take a look.

agent.go
```diff
@@ -875,6 +892,71 @@ func (s *sandbox) setupSignalHandler() error {
...
+func (e *exitProcess) add(id string, exitCode chan int) {
+	e.Lock()
+	defer e.Unlock()
+
+	if e.exitCodes == nil {
+		e.exitCodes = map[string]*exitProcessStatus{}
+	}
+
+	// delete (first in and first out) half of the elements if the length arrive maxExitCodeNum
+	if len(e.exitCodes) >= maxExitCodeNum {
```
lifupan, line 910: this means it couldn't support more than maxExitCodeNum execs run and exit at the same time?  
This isn't what we want.

lifupan:
Hi @keloyang  

I think you'd better describe your use case by detail and let use know why you killed the shim and stop the container process reaping,  then figure out how to solve your issue.

liubin:
Agree with @lifupan .

I'm wondering whati are `some cases`, if these cases are critical errors, we should not let the containers continually run.

yyyeerbo:
This deliberate-man-made resource leak is the same as don't wait process in userspace but blame the kernel for resource leak..

If the shim is dead, maybe we'd better kill the container/pod associated with the dead shim, instead of letting the pod continue running

keloyang:
@lifupan @liubin @yyyeerbo Thanks. 
> This deliberate-man-made resource leak is the same as don't wait process in userspace but blame the kernel for resource leak.

If  " don't wait process in userspace", it is program bug, the author need fix. If process want reap child, but due to some exception(e.g. it was killed),  os init maybe reap the exited process'child.

if kata-shim exit before shim.wait or the connection to kata-proxy was closed before shim.wait, these all make WaitProcesss can't be called and lead kata-agent resourece leak, and It's not recoverable.

If the process of kata-shim is created by exec(e.g. kubectl exec -ti ...),  kata container will not be affected by this shim no matter the shim is abnormal or dead. In the followning ， we can kill -9 823320 and the kata container will still run well, but the kata-agent will leak a fd abort /dev/pts/ptmx, The discussion for this pr is about exec. 
```
[root@centos1 ~]# ps -eaf|grep kata-shim
root      572188  572134  0 18:56 ?        00:00:00 /kata/bin/kata-shim -agent unix:///run/vc/sbs/56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595/proxy.sock -container 56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595 -exec-id 56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595
root      572333  572314  0 18:56 ?        00:00:00 /kata/bin/kata-shim -agent unix:///run/vc/sbs/56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595/proxy.sock -container 6e8a25168697148e7065a9253c2caa5626548f6ed02c877ddc9dc93caaa6b76f -exec-id 6e8a25168697148e7065a9253c2caa5626548f6ed02c877ddc9dc93caaa6b76f
root      823320  572314  0 20:32 pts/3    00:00:00 /kata/bin/kata-shim -agent unix:///run/vc/sbs/56b5e8192430090ba129a7a4823d646a46bac5684055a3e557e8c87efaf37595/proxy.sock -container 6e8a25168697148e7065a9253c2caa5626548f6ed02c877ddc9dc93caaa6b76f -exec-id 014f96c3-67a6-422c-9709-33f717cc624f -terminal
[root@centos1 ~]# kill -9 823320
```
I may not have explained the case clearly above, ues the following case, kata-proxy can limit the concurrent number for the command of "kubectl exec -ti ... ", if concurrent number reached 3, kata-proxy will close the connection from kata-shim. This is an exception test,run the script(yamux.sh), we can't exec to the container some hours later and will see the container can't alloc pts. for exception test, it can make the connection of "kubectl exec" is closed, but it should not lead we can't exec to the container again.
```
# Suppose there is only one pod'name contain nginx
[root@centos0 scripts]# cat yamux.sh
#!/bin/bash

for i in {1..100000};
do
        echo "=========== ${i} ==========="
        kubectl exec -ti `kubectl get pod|grep nginx|awk '{print $1}'` sleep 10
        sleep 1
done
```

hack kata-proxy
```
From 3d097fc0fa3158c6ed3d159eff249d6319aa36a2 Mon Sep 17 00:00:00 2001
From: Shukui Yang <keloyangsk@gmail.com>
Date: Mon, 11 Jan 2021 16:44:54 +0800
Subject: [PATCH] Add proxy connection limit

---
 proxy.go | 52 ++++++++++++++++++++++++++++++++++++++++++++++++----
 1 file changed, 48 insertions(+), 4 deletions(-)

diff --git a/proxy.go b/proxy.go
index ab062a5..f2c87d5 100644
--- a/proxy.go
+++ b/proxy.go
@@ -89,7 +89,7 @@ func heartBeat(session *yamux.Session) {
        }
 }
 
-func serve(servConn io.ReadWriteCloser, proto, addr string, results chan error) (net.Listener, *yamux.Session, error) {
+func serve(servConn io.ReadWriteCloser, proto, addr string, results chan error, sy *shimYamux) (net.Listener, *yamux.Session, error) {
        sessionConfig := yamux.DefaultConfig()
        // Disable keepAlive since we don't know how much time a container can be paused
        sessionConfig.EnableKeepAlive = false
@@ -137,20 +137,23 @@ func serve(servConn io.ReadWriteCloser, proto, addr string, results chan error)
                                return
                        }
 
+                       sy.Add(1)
+
                        // add 2 for the two copy goroutines
                        wg.Add(2)
-                       go proxyConn(conn, stream, wg)
+                       go proxyConn(conn, stream, wg, sy)
                }
        }()
 
        return l, session, nil
 }
 
-func proxyConn(conn1 net.Conn, conn2 net.Conn, wg *sync.WaitGroup) {
+func proxyConn(conn1 net.Conn, conn2 net.Conn, wg *sync.WaitGroup, sy *shimYamux) {
        once := &sync.Once{}
        cleanup := func() {
                conn1.Close()
                conn2.Close()
+               sy.Done()
        }
        copyStream := func(dst io.Writer, src io.Reader) {
                _, err := io.Copy(dst, src)
@@ -164,6 +167,12 @@ func proxyConn(conn1 net.Conn, conn2 net.Conn, wg *sync.WaitGroup) {
 
        go copyStream(conn1, conn2)
        go copyStream(conn2, conn1)
+
+       if !sy.Validate() {
+               time.AfterFunc(time.second*5, func(){
+                       once.Do(cleanup)
+               })
+       }
 }
 
 func unixAddr(uri string) (string, error) {
@@ -450,8 +459,11 @@ func realMain() {
                }
        }()
 
+       sy := &shimYamux{
+               yamuxMax: 5,
+       }
        results := make(chan error)
-       l, s, err := serve(servConn, "unix", listenAddr, results)
+       l, s, err := serve(servConn, "unix", listenAddr, results, sy)
        if err != nil {
                logger().WithError(err).Fatal("failed to serve")
                os.Exit(1)
@@ -479,6 +491,38 @@ func realMain() {
        logger().Debug("shutting down")
 }
 
+type shimYamux struct {
+       yamuxNum int
+       yamuxMax int
+
+       sync.Mutex
+}
+
+func (s* shimYamux) Add (num int) {
+       s.Lock()
+       defer s.Unlock()
+
+       s.yamuxNum += num
+}
+
+func (s* shimYamux) Done () {
+       s.Lock()
+       defer s.Unlock()
+
+       s.yamuxNum--
+}
+
+func (s* shimYamux) Validate () bool {
+       s.Lock()
+       defer s.Unlock()
+
+       if s.yamuxNum > s.yamuxMax {
+               return false
+       }
+       return true
+}
+
+
 func main() {
        defer handlePanic()
        realMain()
-- 
2.19.0
```

fidencio:

PR #882, opened 2021-01-20
arm/pci: add create pci bus path support for acpi

jongwu:
PCI bus path in qemu/arm64 is different between acpi enabled and
disabled. It's better to check if there is a pci bus path for acpi which
is a static value before try to offer a varying pci bus path for no acpi.

Fixes: #881

@jodh-intel @devimc

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -711,7 +711,7 @@ func (s *sandbox) listenToUdevEvents() {
 	fieldLogger := agentLog.WithField("subsystem", "udevlistener")
 	rootBusPath, err := createRootBusPath()
 	if err != nil {
-		fieldLogger.Warnf("Error creating root bus path")
+		fieldLogger.WithError(err).Error("Fail to create root bus path")
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -711,7 +711,7 @@ func (s *sandbox) listenToUdevEvents() {
 	fieldLogger := agentLog.WithField("subsystem", "udevlistener")
 	rootBusPath, err := createRootBusPath()
 	if err != nil {
-		fieldLogger.Warnf("Error creating root bus path")
+		fieldLogger.Warnf("Error creating root bus path: %s", err)
```
devimc, line 714: can you use `.WithError()` ?

devimc (approved):
thanks @jongwu  - I have a question

agent.go
jongwu: OK, thanks @devimc

jongwu:
hello @lifupan @jodh-intel , any comments?

jongwu:
hello @jodh-intel @devimc , any comments?

devimc:
@jongwu lgtm - thanks

jodh-intel (approved):
Thanks @jongwu.

PR #883, opened 2021-02-09
WIP:  mount: don't error of virtiofs share is already mounted

egernst:
In the event that the virtiofs device is already mounted at the
requested destination, don't error out. We'll check before attempting to
mount to see if the destination is already a mount point. If so, skip
doing the mount in the agent.

This facilitates mounting the sharedfs automatically in the guest before
the agent service starts.

Signed-off-by: Eric Ernst <eric.g.ernst@gmail.com>

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -70,6 +70,38 @@ func createDestinationDir(dest string) error {
 	return os.MkdirAll(targetPath, mountPerm)
 }
 
+func isDeviceMountPoint(path string) bool {
+	// If we are at the root, then our check will fail, since we don't have a parent dir
+	if path == "/" {
+		return false
+	}
+
+	// If the path doesn't exist, it certainly isn't a mount
+	if _, err := os.Stat(path); err != nil {
+		return false
+	}
+
+	// We can determine that the path is a mount point if the device
+	// associated with it is different than its parents
+	childStat := syscall.Stat_t{}
+	parentStat := syscall.Stat_t{}
+	parent_path := filepath.Dir(path)
+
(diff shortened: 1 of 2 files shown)


jodh-intel (approved):
Thanks @egernst.

Are you planning to add this feature to the 2.x agent (could you raise an issue for tracking either way)?

lifupan:
Hi @egernst 

It's LGTM for this patch, but I want to know in which case the virtiofs fs would have been mounted outside agent?

egernst:
@lifupan  - yeah, good question. This is an enhancement to facilitate rootfs that is different than what is currently provided in default osbuilder/agent.  It is a bit of downstream enabling, but I don't think its too invasive, and may be helpful for others.

In particular, we have a use case where we'd like the kataShared virtiofs mounted inside the VM even before sandbox create request is received by the agent.  An example of where this would be helpful is if we had certificates that are unique to the host, or other secret that changes too frequently to be directly embedded into the guest rootfs (see https://github.com/kata-containers/kata-containers/pull/1463).  In the event that our rootfs has a systemd automount for the kataShared tag (or equivalent) we'd like to make sure the agent can skip the mount. 

Unfortunately we can't just mount regardless and handle the error from the mount sys call, since the error messages are just about useless. Hence us checking if path exists, and if so, if it is a mount point.

lifupan:
@egernst  Thanks for your explanation.

lifupan (approved):

egernst:
At this point I'm not looking to utilize this on 1.12, and was planning on enabling this use case on top of 2.0. I'll close for now!

